PR #6, opened 2015-09-17
Truncate AMI description to 255 characters

bburtin:
Truncate the AMI description, so that it doesn't exceed 255 characters.
Convert _get_encrypted_image_name() to a new _append_suffix() function
that's called when generating both the AMI name and suffix.

Remove brittle tests that make assumptions about the format of the AMI
name suffix.

Change-Id: I64b63236bc6f6cd30c8ceac8000386bbf78e0039

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -95,10 +95,9 @@ TAG_ENCRYPTOR_AMI = 'BrktEncryptorAMI'
 TAG_DESCRIPTION = 'Description'
 
 NAME_ENCRYPTED_IMAGE = '%(original_image_name)s %(encrypted_suffix)s'
-NAME_ENCRYPTED_IMAGE_SUFFIX = '(encrypted %(nonce)s)'
-DESCRIPTION_ENCRYPTED_IMAGE = (
-    '%(original_image_description)s - based on %(image_id)s, '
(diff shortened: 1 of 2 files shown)


brkt_cli/__init__.py
```diff
@@ -252,14 +251,21 @@ def _get_encrypted_suffix():
...
+def _append_suffix(name, suffix, max_length=None):
+    """ Append the suffix to the given name.  If the appended length exceeds
+    max_length, truncate the name to make room for the suffix.
+
+    :return: The possibly truncated name with the suffix appended
+    """
+    if not suffix:
+        return name
+    if max_length:
+        original_length = max_length - len(suffix)
```
mpage: Nit: I'd probably call this ```truncated_length```

brkt_cli/__init__.py
```diff
@@ -580,12 +586,12 @@ def run(aws_svc, enc_svc_cls, image_id, encryptor_ami):
             raise Exception("Can't find image %s" % encryptor_ami)
 
         # Register the new AMI.
-        name = _get_encrypted_image_name(image.name)
+        name = _append_suffix(
+            image.name, _get_encrypted_suffix(), 128)
```
mpage: Nit: use ```max_length``` keyword arg for consistency with the call below.

PR #7, opened 2015-09-21
Image encryptor fails without sufficient details

bburtin:
When image encryption fails, point the user to the instance console log.
Write the console log to a temporary file, if possible.

Change-Id: I0a573506e96cc323f4d67cf14bc7246776f5d418

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -46,6 +46,7 @@ import datetime
 import requests
 import string
 import sys
+import tempfile
 import time
 import uuid
 import warnings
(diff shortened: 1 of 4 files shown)


brkt_cli/__init__.py
```diff
@@ -206,6 +207,11 @@ def _get_encryption_progress_message(start_time, percent_complete, now=None):
     return msg
 
 
+class EncryptionError(Exception):
+    def __init__(self, message):
```
Kuzmin: you can get away with 

    class EncryptionError(Exception):
        pass

dramdass:
Looks good to me

PR #8, opened 2015-09-22
Use s3 bucket for fetching encryptor AMI

dramdass:
Remove Python 2.7.9 with openssl 1.0+ dependency. Instead
fetch encryptor AMI from s3 bucket. The bucket location depends
on environment and can be overriden by BRACKET_ENVIRONMENT
environment variable. Defaults to stage currently.

Testing done:
- Set BRACKET_ENVIRONMENT to my environment and encrypted
an AMI


brkt_cli/__init__.py
```diff
@@ -120,11 +119,9 @@
...
 EVENTUAL_CONSISTENCY_TIMEOUT = 10
 
-# Right now this is the STAGE endpoint. We need to make this PROD
-# when we have customers running this. This is superceded by the
-# API_URL environment variable if it exists
-API_URL = \
-    "https://stage-api-lb-1316607304.us-west-2.elb.amazonaws.com"
+# Right now this is the STAGE bucket. We need to make this PROD
+BRACKET_ENVIRONMENT = "stage"
+BRACKET_BUCKET_FMT = "http://solo-brkt-%s-net.s3.amazonaws.com/amis.json"
```
bburtin: Should this be something more generic, like ENCRYPTOR_AMIS_URL?  Seems like "bucket" is an implementation detail.  Also, I don't know how much value we're getting out of breaking out "stage" into a separate variable.  Your call on both of these.

brkt_cli/__init__.py
```diff
@@ -276,22 +273,17 @@ def _get_encrypted_image_name(original_name, suffix=None):
...
+    if not bracket_env:
+        raise Exception('No bracket environment found')
+    bucket_url = BRACKET_BUCKET_FMT % (bracket_env)
+    r = requests.get(bucket_url)
     if r.status_code not in (200, 201):
-        raise Exception('Getting encryptor ami gave response: %s', r.text)
-    ami = r.json()['ami_id']
+        raise Exception('Getting %s gave response: %s', bucket_url, r.text)
+    ami = r.json()[region]
     if not ami:
```
bburtin: I don't think this will happen if you're using [] instead of .get().

dramdass: Good call. Will use get

brkt_cli/__init__.py
dramdass: I like ENCRYPTOR_AMIS_URL. Eventually SEs and developers will need to use this env variable to override which backend to point to so want to keep than env variable simple so will keep as separate.

dramdass:
I made the changes and rebased my environment and retested. This should be ready for merge now

bburtin:
Ship it.

PR #9, opened 2015-09-23
Delete EBS volumes created by the copier instance

bburtin:
Tag volumes attached to the snapshotter and encryptor instances.  After
instances have terminated, search for any volumes that were unexpectedly
orphaned by AWS and delete them.

Add a cleanup step for the snapshotter instance, so that it gets
terminated if something goes wrong while snapshotting the root volume.

Make the encryptor session id the same kind of shorter nonce that we use
to avoid name conflicts.

Change-Id: Ic01ebe69f49b7c24bf70f65d034d7f5580edefc8


bburtin:
Found bugs.  Will open a new pull request later.

PR #13, opened 2015-09-25
Retry logic for fetching console output

rtroxell:
Fetching console logs on ecnryption failures rarely, if ever, succeeds
due to the typical lag in console updates from AWS. This patch adds
retry logic that waits up to 5 minutes for console output to show up
in AWS.

Testing
- Ran test.py
- Ran brkt encrypt-ami against an ami with encryption failures, verified console logs are downloaded (took about 2 minutes before the fetch worked)

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -112,6 +112,8 @@ SLEEP_ENABLED = True
 
 EVENTUAL_CONSISTENCY_TIMEOUT = 10
 
+GET_CONSOLE_TIMEOUT = 5 * 60
+
 # Right now this is the STAGE bucket. We need to make this PROD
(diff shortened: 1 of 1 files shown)


bburtin:
Richard and I talked about this pull request, and decided to wait to merge it.  I think safe to assume that the user will know how to retrieve the console log, given an instance ID.  I'd rather terminate immediately, rather than sit around for 5 minutes, hoping that the console log becomes available.  We can resurrect this pull request if this becomes an issue in the field.

rtroxell:
Thanks for quashing my github dreams Boris.

bburtin:
I'm happy to unquash by assigning some CLI bugs to you!

PR #15, opened 2015-09-30
reference_templates dir and first reference vpc

bobveznat:
With this commit we gain a troposphere implementation of a secure VPC
containing dual subnets in two availability zones. Troposphere was run
against the implementation and the generated json file is checked in for
convenience. This json file may be used directly by cloudformation.
Further documentation of the VPC architecture is in
vpc-dual-az-with-nat.py


reference_templates/vpc-dual-az-with-nat.json
```diff
@@ -0,0 +1,1111 @@
...
+                "NATAMI": "ami-ada746e9"
+            },
+            "us-west-2": {
+                "NATAMI": "ami-75ae8245"
+            }
+        }
+    },
+    "Outputs": {
+        "PrivateSubnet0": {
+            "Description": "SubnetId of a private subnet for 1",
```
synfinatic: id miss-match.  0 != 1

reference_templates/vpc-dual-az-with-nat.json
```diff
@@ -0,0 +1,1111 @@
...
+            "Properties": {
+                "FromPort": "22",
+                "GroupId": {
+                    "Ref": "InternetClientSG"
+                },
+                "IpProtocol": "tcp",
+                "SourceSecurityGroupId": {
+                    "Ref": "NATSecurityGroup"
+                },
+                "ToPort": "22"
```
synfinatic: while alphabetical order is generally a good thing, putting FromPort and ToPort at opposite ends of the block makes it hard for humans to parse IMHO.

reference_templates/vpc-dual-az-with-nat.py
```diff
@@ -0,0 +1,454 @@
+"""A VPC with a pair of public and private subnets.
```
synfinatic: #!/usr/bin/env python and chmod 755

reference_templates/vpc-dual-az-with-nat.py
```diff
@@ -0,0 +1,454 @@
...
+
+# This next block of code deals with the configuration of Network ACLs. The
+# default configuration here allows access to the Internet using tcp/80 (HTTP),
+# tcp/443 (HTTPS) and udp/123 (NTP) access to the Internet.
+# These ACLs also permit SSH from the trusted cidr to the public subnet and
+# permit SSH from the public subnet to the private subnet.
+# Nothing else is explicitly permitted and thus nothing else works.
+#
+# When making changes here don't forget to also make changes in the
+# corresponding security group(s)
```
synfinatic: it's worth making a comment somewhere that NACL's aren't stateful because it's 2015 and that's just expected behavior and people are likely to look at these rules and go o_O

PR #16, opened 2015-10-02
reference_templates dir and first reference vpc

bobveznat:
With this commit we gain a troposphere implementation of a secure VPC
containing dual subnets in two availability zones. Troposphere was run
against the implementation and the generated json file is checked in for
convenience. This json file may be used directly by cloudformation.
Further documentation of the VPC architecture is in
vpc-dual-az-with-nat.py


bobveznat:
This is the same as pull request 15, but post squashing.

PR #25, opened 2015-10-13
Allow specifying the encrypted AMI name as a command option

bburtin:
Add an --encrypted-ami-name option.  When the name is specified, pass it
through to the register_image() call.  Break out the maximum name length
into a constant.

Change-Id: I3cf20075137552ecdeb3edc12ab5d599386e7795


bburtin:
Found out that Python doesn't like len(None).  Closing this pull request to open another with the bug fixed.

PR #26, opened 2015-10-14
Another cfn template: vpc-dual-az-with-nat-and-vpn

bobveznat:
Similar to the previously added cloud formation template, now with site
to site VPN!

### DIFF
--- a/reference_templates/vpc-dual-az-with-nat-and-vpn.py
+++ b/reference_templates/vpc-dual-az-with-nat-and-vpn.py
@@ -1,0 +1,556 @@
+#!/usr/bin/env python
+"""A VPC with a pair of public and private subnets and an ipsec vpn.
+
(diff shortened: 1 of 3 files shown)


reference_templates/vpc-dual-az-with-nat-and-vpn.py
```diff
@@ -0,0 +1,556 @@
...
+    # A rule to allow the use of udp/123 (NTP)
+   SecurityGroupRule(
+        IpProtocol="udp",
+        FromPort="123",
+        ToPort="123",
+        SourceSecurityGroupId=Ref(internet_client_sg),
+    ),
+]
+
+nat_in_rules = internet_client_rules
```
synfinatic: why two names for the same thing which is only used once?

synfinatic:
other then that one nit, it LGTM

bburtin:
Closing old pull request.

PR #28, opened 2015-10-16
Add Instance type option to the brkt-cli

krishnanrs:
Any suggestions on how to add the instance type validation?
Also, should the entered instance type be used for the volume encryptor instance as well?

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -881,6 +881,7 @@ def main():
             return 1
             
    instance_type = values.instance_type
+   # TODO: Validate the entered instance type
    if not instance_type:
       instance_type = 'm3.medium'
 
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -421,11 +421,11 @@ def _run_encryptor_instance(aws_svc, encryptor_image_id, snapshot, root_size,
     return instance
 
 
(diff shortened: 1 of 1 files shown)


brkt_cli/__init__.py
```diff
@@ -806,6 +807,13 @@ def main():
         dest='key_name',
         required=True
     )
+   encrypt_ami.add_argument(
+        '--instance-type',
```
bburtin: Should we call this --snapshotter-instance-type, so that people don't confuse it with the encryptor instance type?  If you agree, the help description should also be more specific.

bburtin:
This looks good.  My only concern is whether we think that this option really is useful to customers.  If it's for internal testing, we shouldn't expose it.  Fewer knobs to turn is generally a good thing.

brkt_cli/__init__.py
krishnanrs: If we decide to use the instance type only for the snapshotter, then I would agree with you. However that was one of my question(s). Typically using larger encryptor instance can also speed up data copy/encryption.

bburtin: Then we should use this option for both instances?

bburtin:
Krishnan and I agreed to close this pull request because there's no short-term need.

PR #34, opened 2015-10-20
Fix handling of ephemeral drive mappings.

xcllnt:


brkt_cli/__init__.py
```diff
@@ -633,14 +632,15 @@ def run(aws_svc, enc_svc_cls, image_id, encryptor_ami,
...
-            t.ephemeral_name = 'ephemeral%d' % (i,)
-            i += 1
-            new_bdm[drive] = t
+        # Propagate any ephemeral drive mappings to the soloized image
+        guest_image = aws_svc.get_all_images([image_id])[0]
+        guest_bdm = guest_image.block_device_mapping
+        for key in guest_bdm.keys():
+            guest_vol = guest_bdm[key]
+            if guest_vol.ephemeral_name:
+                log.info('Propagating block device mapping for %s at %s' %
```
bburtin: End users will see this.  Will they understand what it means?

PR #36, opened 2015-10-20
Update bracket environment to prod

dramdass:
This will point customers to the prod Bracket environment by default


bburtin:
Ship it!

PR #37, opened 2015-10-21
Don't require key pair name

bburtin:
Make --key a hidden option.  It's no longer required to launch the
encryptor node, and is only used for development.

Change-Id: I834c9c1d4a8be0547820acb4177514f9902faa30


mpage:
LGTM :+1:

PR #41, opened 2015-10-22
Break out encrypt ami into own module

berndtbrkt:
Allows for library usage.  Also, provides a better organization
for future functionality.


bburtin:
Looks good.  My only reaction is that we shouldn't mix command line parsing with application logic.  I would either leave the command line parsing code in __init__.py or move it into a separate module.  Also, a couple of nits:

1) Can you rename any formerly-private functions in encrypt_ami that are called from __init__, so that they aren't prefixed with "_"?
2) If it's convenient, see if there's a way to factor out the VERSION variable.  Currently there are two version strings: one in __init__.py and one in setup.py.  I tried to combine them, but ran into circular dependency issues.  This could be a good opportunity to clean this up.

bburtin:
LGTM.

PR #45, opened 2015-10-26
Make the --key argument optional

bburtin:
The refactoring work in commit 72eaa85ebe15d3953f697476a592d9b0d94d5662
reverted the recent change to the --key argument and made it a required
argument.  Make --key a hidden, optional argument again.


bburtin:
That's a good point.  I'll do that now.

bburtin:
Closing and reopening a new pull request that makes both arguments hidden.

PR #48, opened 2015-10-26
Handle empty AWS API responses

bburtin:
The AWS API can occasionally return an empty result set instead of an
API error.  Detect this case and raise an EC2ResponseError, to avoid a
"list index out of range" exception.

### DIFF
--- a/brkt_cli/service.py
+++ b/brkt_cli/service.py
@@ -174,6 +174,19 @@ def retry_boto(error_code_regexp, max_retries=5, initial_sleep_seconds=0.25):
     return retry_decorator
 
 
+def _get_first_element(list, error_status):
+    """ Return the first element in the list.  If the list is empty, raise
+    an EC2ResponseError with the given error status.  This is a workaround
+    for the case where the AWS API erroneously returns an empty list instead
+    of an error.
+    """
+    if list:
+        return list[0]
+    else:
+        raise EC2ResponseError(
+            error_status, 'AWS API returned an empty response')
+
+
 class AWSService(BaseAWSService):
 
     def __init__(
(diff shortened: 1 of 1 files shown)


brkt_cli/service.py
```diff
@@ -362,7 +390,14 @@ def delete_security_group(self, sg_id):
             raise Exception('Unknown error while deleting security group')
 
     def get_key_pair(self, keyname):
```
berndtbrkt: why isn't there a retry on this?

brkt_cli/service.py
```diff
@@ -231,7 +231,14 @@ def run_instance(self,
...
     def get_instance(self, instance_id):
-        return self.conn.get_only_instances([instance_id])[0]
+        instances = self.conn.get_only_instances([instance_id])
+        if not instances:
+            # Handle the AWS API incorrectly returning an empty response.
+            raise EC2ResponseError(
+                'InvalidInstanceID.NotFound',
+                'AWS API returned an empty response'
+            )
+        return instances[0]
```
berndtbrkt: I feel like this could be refactored a bit.  Maybe create a function which attempts to return the first item, or raise an error:

def get_only_ec2_resource(resources, error_msg)

It's probably fine as-is, it just seems like a lot of boilerplate.

brkt_cli/service.py
bburtin: It's incidental.  We don't create key pairs during the encryption process, so there's no need to retry.  If we retried, it would create a 7-second-long delay when the user specifies an invalid key pair name on the command line.  I know the code seems inconsistent.  Let me know if you have suggestions on improving it.

brkt_cli/service.py
bburtin: Good point.  Let me see if there's a way to consolidate the logic.

PR #49, opened 2015-10-27
Add setup.cfg

bburtin:
This file is needed for submitting to PyPI.


bburtin:
Danny verified that pip install works.  Merging.

PR #50, opened 2015-10-27
Update README.md

bburtin:
Remove the GitHub URL from pip install instructions.  It's no longer necessary, now that brkt-cli is on PyPI.


bburtin:
Danny verified that pip install works.  Merging.

PR #54, opened 2015-11-03
Propagate ephemeral drive mappings

bburtin:
Copy ephemeral drive mappings from the source AMI, instead of using
hardcoded device paths.


bburtin:
I had to manually merge Marcel's change because the code moved to encrypt_ami.py.

PR #56, opened 2015-11-09
Updates reference VPC to include egress on 7001/2

robertabbott:
The brkt HSMProxy and API are now accessed via ports
7001 and 7002 respectively. This commit updates the
reference VPC to enable egress on ports 7001 and
7002

### DIFF
--- a/reference_templates/vpc-dual-az-with-nat.py
+++ b/reference_templates/vpc-dual-az-with-nat.py
@@ -217,7 +217,8 @@ public_egress_acl_entries = [
     ('PublicAllowHTTPOut', 110, 6, 80, '0.0.0.0/0'),
     ('PublicAllowTCPResponsesOut', 120, 6, (1024, 65535), '0.0.0.0/0'),
     ('PublicAllowSSHToPrivateInstances', 130, 6, 22, vpc_cidr),
-    ('PublicAllowNTPOut', 140, 17, 123, '0.0.0.0/0')
+    ('PublicAllowNTPOut', 140, 17, 123, '0.0.0.0/0'),
+    ('PublicAllowBrktOut', 150, 6, (7001, 7002), '0.0.0.0/0')
 ]
 for entry in public_egress_acl_entries:
     add_public_egress_acl_entry(*entry)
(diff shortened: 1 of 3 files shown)


reference_templates/vpc-dual-az-with-nat.py
```diff
@@ -217,7 +217,9 @@ def add_private_egress_acl_entry(name, number, proto, port, dst_cidr):
     ('PublicAllowHTTPOut', 110, 6, 80, '0.0.0.0/0'),
     ('PublicAllowTCPResponsesOut', 120, 6, (1024, 65535), '0.0.0.0/0'),
     ('PublicAllowSSHToPrivateInstances', 130, 6, 22, vpc_cidr),
-    ('PublicAllowNTPOut', 140, 17, 123, '0.0.0.0/0')
+    ('PublicAllowNTPOut', 140, 17, 123, '0.0.0.0/0'),
+    ('PublicAllowBrktHSM', 150, 6, 7001, '0.0.0.0/0'),
```
bobveznat: This can be s/7001/(7001, 7002)/ (it takes a tuple for port range) and delete rule 160.

bobveznat:
Thanks for giving this a shot. You need a few more rules. When I think about this I start at the node that's making the request and imagine where the packets go. Assume the instance is in the private subnet.

The packet must leave the private subnet (private egress) on its way to the NAT. It must then enter the public subnet (public ingress). Once within the public subnet it must enter the NAT instance (nat security group). The NAT instance does its NATing, and passes the packet towards the igw (public egress).

Now our packet has reached the API server. We need to handle the response.

First, the response packet enters the public subnet (public ingress), then the instance (the NAT sg, security groups are stateful though so no rules are required here). Then the instance passes the packet back towards the private subnet (public subnet egress, private subnet ingress).

So you need all of those. A lot of the rules are already in there (the allow tcp responses rules handle a lot of your replies). In general look at where port 443 is in there and mimic it for 7001, 7002. You need the same ones.

robertabbott:
Updated it with rules for private subnet egress, public subnet ingress, NAT instance ingress and public subnet egress.

bobveznat:
This looks better. Have you deployed this and launched an instance in it to verify connectivity?

robertabbott:
yes. I launched a metavisor instance and it was able to talk out to Yeti

PR #57, opened 2015-11-11

bburtin:
This is an initial code review, just to validate the approach.  Before
merging, we need to finalize the EULA text and update the integration
tests to pass the username and password to the CLI.

Add --username (required) and --password (optional) arguments to brkt-
cli.  If --password is not specified, prompt the user for the password.
EULA text.  The user must type the word "yes" (case-insensitive) in
order to accept the EULA.

service API.


bburtin:
The change to the version code in setup.py is cruft.  I'll fix it.

setup.py
```diff
@@ -15,7 +15,7 @@
 import re
 from setuptools import setup
 
-version = ''
+version = '0.9.4'
```
bburtin: This is cruft.  I'll remove it in the next pull request.

bburtin:
Will open another pull request with the latest code.

PR #58, opened 2015-11-12

bburtin:
This is an initial code review, just to validate the approach.  Before
merging, we need to finalize the eval agreement URL and update the
integration tests to pass the username and password to the CLI.

Add --username (required) and --password (optional) arguments to brkt-
cli.  If --password is not specified, prompt the user for the password.
EULA text.  The user must type the word "yes" (case-insensitive) in
order to accept the EULA.

service API.


bburtin:
Closing, will reopen with the next version of the code.

PR #60, opened 2015-11-13

bburtin:
Add the following options at the top level of the brkt command:

--username (required)
--password (optional)
--api-root (hidden, development only)
--no-verify-cert (hidden, development only)

brkt-cli now authenticates with the Bracket service before doing
any work.

If --password is not specified, prompt the user for the password.
EULA text.  The user must type the word "YES" (case-insensitive) in
order to accept the EULA.

service API.


berndtbrkt:
Any thought about accepting api keys (as env vars) instead of requiring username/password?  You could make the username/password required if the keys are not present.

bburtin:
Honestly I don't know what combination of environment variables, config files, command line options, and prompting is best.  Let me do some more research.  If someone knows of an established convention, please let me know.

bburtin:
The MySQL docs discourage using environment variables: "This method of specifying your MySQL password must be considered extremely insecure and should not be used. Some versions of ps include an option to display the environment of running processes."

https://dev.mysql.com/doc/refman/5.7/en/password-security-user.html

Can we keep the current scheme and add support for a config file in a subsequent commit?

bburtin:
Closing and reopening to resolve conflict with master.

PR #65, opened 2015-11-18
Update README.md

dancudds:
Changed order of access:secret key to match IAM console and changed name

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ Before running the **brkt** command, make sure that you've set the AWS
 environment variables:
 
 ```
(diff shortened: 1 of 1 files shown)


bburtin:
Thanks Dan.  Just so I'm educated, what was unclear/inconsistent about the previous terminology?

dancudds:
Just to help copy/paste users like me :) 
The order of the access:secret keys was different to the IAM console and the AWS_SECRET_ACCESS_KEY was prompting for <access key> causing me to mix the two :(

PR #68, opened 2015-11-23
Check Python version

bburtin:
Exit with an error status if the Python version is not 2.7.


gdmen:
lgtm

brkt_cli/__init__.py
```diff
@@ -35,6 +35,17 @@
...
+    # Check Python version.
+    version = '%d.%d' % (sys.version_info.major, sys.version_info.minor)
+    if version != '2.7':
+        print(
+            'brkt-cli requires Python 2.7.  Version',
+            version,
+            'is not supported.',
+            file=sys.stderr
+        )
+        exit(1)
```
gdmen: Should this be 'return 1'?

bburtin: Good call.

PR #69, opened 2015-11-23
Check Python version

bburtin:
Exit with an error status if the Python version is not 2.7.


bburtin:
Gary reviewed this code in a previous pull request.  Merging.

PR #70, opened 2015-11-24
Gdm refactor s3

gdmen:
review please, I'll squash and do a single commit pull request after review


brkt_cli/aws_service.py
```diff
@@ -371,6 +378,29 @@ def delete_security_group(self, sg_id):
         if not ok:
             raise Exception('Unknown error while deleting security group')
 
+    def create_instance_profile(self, base_name, policy):
+        base_name += '-' + self.session_id
+        profile_name = base_name + '-profile'
+        role_name = base_name + '-role'
+        policy_name = base_name + '-policy'
+        iam_conn = boto.connect_iam()
```
bburtin: Can you add comments that explain why we need a separate connection here?

brkt_cli/encrypt_ami.py
```diff
@@ -271,11 +276,12 @@ def get_encryptor_ami(region):
         raise BracketError('No bracket environment found')
     bucket_url = ENCRYPTOR_AMIS_URL % (bracket_env)
     log.debug('Getting encryptor AMI list from %s', bucket_url)
-    r = requests.get(bucket_url)
-    if r.status_code not in (200, 201):
+    r = urllib2.urlopen(bucket_url)
```
bburtin: Just out of curiosity, what does urllib2 do that requests doesn't?  I thought requests was the more robust library.

gdmen: urllib2 doesn't require a 3rd party dependency and this is the only place we use requests

brkt_cli/aws_service.py
gdmen: I assumed that IAM and S3 connections are different from a vpc connection. I'll check.

brkt_cli/encrypt_ami.py
bburtin: Only in the short term, unfortunately.  Once we require login for brkt-cli, we'll depend on brkt_sdk, which depends on requests.  But you're welcome to pick apart that dependency too.  :-)

brkt_cli/encrypt_ami_s3.py
```diff
@@ -0,0 +1,180 @@
...
+from boto.ec2.blockdevicemapping import (
+    BlockDeviceMapping,
+    EBSBlockDeviceType,
+)
+from boto.s3.connection import S3Connection
+
+from brkt_cli.util import BracketError
+
+from encrypt_ami import (
+    _append_suffix,
```
bburtin: Seems like we should make these functions public, since they're called from another module?

gdmen: sure

brkt_cli/aws_service.py
gdmen: Coincidentally removed this. I'm having the user input an instance profile name on the command line instead.

gdmen:
Fixed & squashed pull request coming on another branch.

PR #71, opened 2015-11-24
Add a timeout for encryption progress

bburtin:
Raise an EncryptionError if there's not 1% of encryption progress in 10
minutes.


gdmen:
lgtm

PR #74, opened 2015-11-27
Add encrypt_ami_s3 with options for S3 source.

gdmen:
This is currently lib funcionality only and not hooked up to the brkt
command.

New flags in encrypt_ami_s3:
--guest-root ubuntu --bucket encryptor --iam-profile s3_ro

- aws_service instance launch now takes user-data and instance profile
name args.
- remove requests dependency (replaced with urllib2)
- rename some encrypt_ami functions to public (no underscore)


gdmen:
I should figure out how to run the tests before making pull requests. . .

PR #75, opened 2015-11-27
Add encrypt_ami_s3 with options for S3 source.

gdmen:
This is currently lib funcionality only and not hooked up to the brkt
command.

New options in encrypt_ami_s3:
s3_bucket, s3_key, iam_profile

- aws_service instance launch now takes user-data and instance profile
name args.
- remove requests dependency (replaced with urllib2)
- rename some encrypt_ami functions to public (no underscore)


brkt_cli/aws_service.py
```diff
@@ -227,12 +232,14 @@ def run_instance(self,
...
                 block_device_map=block_device_map,
-                security_group_ids=security_group_ids
+                security_group_ids=security_group_ids,
+                user_data=user_data,
+                instance_profile_name=instance_profile_name
             )
             return reservation.instances[0]
-        except EC2ResponseError:
+        except EC2ResponseError as e:
             # Log the failed operation, so that the user has context.
```
bburtin: Get rid of this comment, since it no longer applies?

brkt_cli/encrypt_ami.py
```diff
@@ -208,13 +214,14 @@ def _wait_for_encryption(enc_svc,
...
     last_progress = 0
+    last_state = ''
 
     while err_count < max_errs:
         try:
             status = enc_svc.get_status()
             err_count = 0
         except Exception as e:
-            log.warn("Failed getting encryption status: %s", e)
+            log.debug("Failed getting encryption status: %s", e)
```
bburtin: With this change, the user won't know why the CLI couldn't talk to the encryptor.  Is this what we want?

brkt_cli/encrypt_ami.py
```diff
@@ -229,14 +236,18 @@ def _wait_for_encryption(enc_svc,
...
+            last_state = state
             progress_deadline = Deadline(progress_timeout)
 
         # Log progress once a minute.
         now = time.time()
         if now - last_log_time >= 60:
-            log.info('Encryption is %d%% complete', percent_complete)
+            state_display = 'Step 1 of 2: Download from S3'
+            if state == encryptor_service.ENCRYPT_RUNNING:
+                state_display = 'Step 2 of 2: Encryption'
```
bburtin: Looks like this would print "step 2 of 2" in the non-S3 case?

brkt_cli/encrypt_ami.py
```diff
@@ -248,6 +259,9 @@ def _wait_for_encryption(enc_svc,
             if failure_code == encryptor_service.FAILURE_CODE_UNSUPPORTED_GUEST:
                 raise UnsupportedGuestError(
                     'The specified AMI uses an unsupported operating system')
+            if failure_code == encryptor_service.FAILURE_CODE_AWS_PERMISSIONS:
```
bburtin: Would the user know what to do to fix this?  Maybe we should make the failure code more specific?

brkt_cli/encrypt_ami.py
gdmen: I think yes, this is sometimes spammy even in a working case. In successful tests of mine, these have printed between successful contacts w/ the encryptor instance.

brkt_cli/encrypt_ami.py
```diff
@@ -509,6 +529,206 @@ def _terminate_instance(aws_svc, id, name, terminated_instance_ids):
...
+    aws_svc.create_tags(ami)
+    ami_info['ami'] = ami
+    ami_info['name'] = name
+
+    log.info('Created encrypted AMI %s (%s)', ami, description)
+
+    return ami_info
+
+
+def cleanup_aws(aws_svc, encryptor_instance, terminated_instance_ids, sg_id):
```
bburtin: Oops, I recently rewrote this code.  Can you see if my version works for you?

brkt_cli/encrypt_ami_s3.py
```diff
@@ -0,0 +1,152 @@
+# Copyright 2015 Bracket Computing, Inc. All Rights Reserved.
```
bburtin: Do we want this in the brkt-cli repo?

brkt_cli/encryptor_service.py
```diff
@@ -48,13 +51,14 @@ class EncryptorService(BaseEncryptorService):
...
         try:
             self.get_status()
+            log.debug("Successfully got encryptor status")
             return True
         except Exception as e:
             log.debug("Couldn't get encryptor status: %s", e)
             return False
 
     def get_status(self, timeout_secs=2):
-        url = 'http://%s:%d/encryption_status' % (self.hostname, self.port)
```
bburtin: Backward-incompatible change?

brkt_cli/encrypt_ami.py
bburtin: But now, if the CLI can't talk to the encryptor instance, the user will see "Encryption service unavailable", without any context.

gdmen: When it's working, this will be a blip, when it's broken, then can run it with -v

brkt_cli/encrypt_ami.py
gdmen: I can make the output more specific, saying what permissions are needed. The helptext already says that.

brkt_cli/encrypt_ami.py
gdmen: Yeah, I think I need to do a manual merge

brkt_cli/encrypt_ami_s3.py
gdmen: Every other file has it?

brkt_cli/encryptor_service.py
gdmen: Don't know why we were using encryption status before? The encryptor ami doesn't have any reference to an 'encryption_status' endpoint (prior to my changes, but I'll check again)

brkt_cli/encrypt_ami.py
gdmen: Changed it to 'The specified IAM profile has insufficient permissions'

brkt_cli/encrypt_ami.py
gdmen: done

brkt_cli/aws_service.py
gdmen: done

brkt_cli/encryptor_service.py
gdmen: > git log -Sencryption_status

I don't see the string 'encryption_status' anywhere in the history of the avatar repo

brkt_cli/encrypt_ami.py
gdmen: done

gdmen:
@bburtin Replied to or fixed everything & re-pushed.

brkt_cli/encrypt_ami_s3.py
bburtin: Sorry, I should have been more specific.  I was talking about the entire file.  Is encrypting from S3 something that we want to expose to all customers?  I was under the impression that the S3 encryption CLI is in another repo.  If this is the case, seems like, the supporting code should also be there?

brkt_cli/encrypt_ami.py
bburtin: The reason I'm concerned about this is because customers want to run encryption in a VPC.  This makes it likely that they launch an encryptor that the CLI can't talk to.  When this happens, we need to be clear about what went wrong.  I don't believe that saying "Encryption service unavailable" and expecting the user to rerun with -v (which takes 5-10 minutes) is a good end user experience.

It shouldn't be hard to store the exception that was raised and put the text into the EncryptionError that is raised at the end of the function.  Does that sound reasonable?

gdmen: Sure

gdmen: Changed it back to log.warn and just added a log.warn("Retrying. . .") after

gdmen:
Opening a new pull request with encrypt_ami_s3 removed (moved to cephalopod)

PR #77, opened 2015-12-02
Update encrypted images

rohang06:
This change adds the update-encrypted-ami option to the brkt_cli.
It allows a user to retrieve a metavisor upgrader AMI and use it
to upgrade an already encrypted AMI. This provides a speed
improvement for users who wish to update their encrypted AMIs.

The user must specify the metavisor updater AMI as an argument.
This requirement will be taken away in a near-future commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ optional arguments:
 $ python brkt update-encrypted-ami --help
 usage: brkt update-encrypted-ami [-h] --updater-ami UPDATER_AMI_ID --region
                                  REGION [--encrypted-ami-name NAME]
+                                 [--no-validate-ami]
                                  AMI_ID
 
 positional arguments:
@@ -64,6 +65,7 @@ optional arguments:
   --region REGION       AWS region (e.g. us-west-2)
   --encrypted-ami-name NAME
                         Specify the name of the generated encrypted AMI
+  --no-validate-ami     Don't validate encrypted AMI properties
 ```
 
 ## Configuration
@@ -103,23 +105,23 @@ messages are written to stderr.
 
 ## Updating an encrypted AMI
 
-Run **brkt update-encrypted-ami** to update anw encrypted AMI based on an existing
+Run **brkt update-encrypted-ami** to update an encrypted AMI based on an existing
 encrypted image:
 
 ```
 $ brkt update-encrypted-ami --region us-east-1 --updater-ami ami-32430158 ami-72094e18
-11:09:46 Commencing update for AMI ami-72094e18
-11:09:46 Creating guest volume snapshot
-11:10:18 Launching metavisor updater instance
+13:38:14 Using zone us-east-1a
+13:38:15 Updating ami-72094e18
+13:38:15 Creating guest volume snapshot
 ...
-11:11:53 metavisor updater snapshots ready
+13:39:25 Encrypted root drive created.
 ...
-11:11:53 Terminating encryptor instance i-9ecfe120
+13:39:28 waiting for snapshot ready
+13:39:48 metavisor updater snapshots ready
 ...
-11:11:55 Registered AMI ami-b9d795d3 based on the snapshots.
-11:12:00 Created encrypted AMI ami-b9d795d3 based on ami-72094e18
-Done.
-ami-b9d795d3
+13:39:54 Created encrypted AMI ami-63733e09 based on ami-72094e18
+13:39:54 Done.
+ami-63733e09
 ```
 
 When the process completes, the new AMI id is written to stdout.  All log
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 **brkt-cli** is a command-line interface to the [Bracket Computing](http://www.brkt.com)
 service.  It produces an encrypted version of an Amazon Machine Image, which can then be
-launched in EC2.
+launched in EC2. It can also update an already encrypted version of an Amazon Machine Image,
+which can then be launched in EC2.
 
 ## Requirements
 
@@ -47,6 +48,23 @@ optional arguments:
   --no-validate-ami     Don't validate AMI properties
   --region NAME         AWS region (e.g. us-west-2)
 ```
+```
+$ python brkt update-encrypted-ami --help
+usage: brkt update-encrypted-ami [-h] --updater-ami UPDATER_AMI_ID --region
+                                 REGION [--encrypted-ami-name NAME]
+                                 AMI_ID
+
+positional arguments:
+  AMI_ID                The AMI that will be encrypted
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --updater-ami UPDATER_AMI_ID
+                        The metavisor updater AMI that will be used
+  --region REGION       AWS region (e.g. us-west-2)
+  --encrypted-ami-name NAME
+                        Specify the name of the generated encrypted AMI
+```
 
 ## Configuration
 
@@ -78,6 +96,30 @@ $ brkt encrypt-ami --key my-aws-key --region us-east-1 ami-76e27e1e
 15:57:12 Deleting snapshot copy of original root volume snap-847da3e1
 15:57:12 Done.
 ami-07c2a262
+```
+
+When the process completes, the new AMI id is written to stdout.  All log
+messages are written to stderr.
+
+## Updating an encrypted AMI
+
+Run **brkt update-encrypted-ami** to update anw encrypted AMI based on an existing
+encrypted image:
+
+```
+$ brkt update-encrypted-ami --region us-east-1 --updater-ami ami-32430158 ami-72094e18
+11:09:46 Commencing update for AMI ami-72094e18
+11:09:46 Creating guest volume snapshot
+11:10:18 Launching metavisor updater instance
+...
+11:11:53 metavisor updater snapshots ready
+...
+11:11:53 Terminating encryptor instance i-9ecfe120
+...
+11:11:55 Registered AMI ami-b9d795d3 based on the snapshots.
+11:12:00 Created encrypted AMI ami-b9d795d3 based on ami-72094e18
+Done.
+ami-b9d795d3
 ```
 
 When the process completes, the new AMI id is written to stdout.  All log
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -177,9 +177,6 @@ def command_update_encrypted_ami(values, log):
                file=sys.stderr)
         return 1
     aws_svc.connect(region)
-    zones = [str(z.name) for z in aws_svc.conn.get_all_zones()]
-    zone = zones[0]
-    log.info('Using zone %s', zone)
     encrypted_ami = values.ami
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -47,6 +48,25 @@ optional arguments:
   --no-validate-ami     Don't validate AMI properties
   --region NAME         AWS region (e.g. us-west-2)
 ```
+```
+$ python brkt update-encrypted-ami --help
+usage: brkt update-encrypted-ami [-h] --updater-ami UPDATER_AMI_ID --region
+                                 REGION [--zone ZONE]
```
bburtin: Let's remove the zone argument, unless it's really necessary.

brkt_cli/__init__.py
```diff
@@ -197,6 +157,147 @@ def main():
...
+    if util.validate_region(region) == 1:
+        print ('Invalid region %s' % region,
+               file=sys.stderr)
+        return 1
+    nonce = util.make_nonce()
+    default_tags = encrypt_ami.get_default_tags(nonce, '')
+    aws_svc = aws_service.AWSService(
+        nonce, None, default_tags=default_tags)
+    aws_svc.connect(region)
+    zones = [str(z.name) for z in aws_svc.conn.get_all_zones()]
```
bburtin: Be careful with aws_svc.conn.  It won't work with unit tests.  AWSService is meant to be an abstraction.

README.md
rohang06: Fixed

README.md
```diff
@@ -82,3 +100,27 @@ ami-07c2a262
 
 When the process completes, the new AMI id is written to stdout.  All log
 messages are written to stderr.
+
+## Updating an encrypted AMI
+
+Run **brkt update-encrypted-ami** to update anw encrypted AMI based on an existing
```
dramdass: nit: s/anw/an

dramdass:
Small comment above. Code looks reasonable to me - nice refactors. I'll let Boris give the final approval

brkt_cli/update_encrypted_ami.py
```diff
@@ -0,0 +1,135 @@
...
+    guest_volume_mapping = \
+        encrypted_image.block_device_mapping.get('/dev/sda5')
+    if not guest_volume_mapping:
+        return None, None, None, \
+            'Invalid block device mapping: /dev/sda5 not present'
+    snapshot = aws_service.conn.get_all_snapshots(
+        guest_volume_mapping.snapshot_id)[0]
+    # Create new volume off this snapshot, snapshot the volume and
+    # return the snapshot
+    guest_volume = snapshot.create_volume(zone)
```
dramdass: do we need to do this snapshot -> volume -> snapshot logic here? the customer owns the guest root snapshot so we shouldn't need to do this

dramdass:
One more comment here.

brkt_cli/__init__.py
```diff
@@ -197,6 +157,147 @@ def main():
...
+            return 1
+    region = values.region
+    if util.validate_region(region) == 1:
+        print ('Invalid region %s' % region,
+               file=sys.stderr)
+        return 1
+    nonce = util.make_nonce()
+    default_tags = encrypt_ami.get_default_tags(nonce, '')
+    aws_svc = aws_service.AWSService(
+        nonce, None, default_tags=default_tags)
```
bburtin: I just realized that the encryptor_ami argument to the AWSService constructor is not used.  Can you use this opportunity to remove it?

brkt_cli/__init__.py
```diff
@@ -197,6 +157,147 @@ def main():
...
+    else:
+        zone = zones[0]
+        log.info('Using zone %s', zone)
+    encrypted_ami = values.ami
+    guest_ami_error = aws_svc.validate_guest_ami(encrypted_ami)
+    if guest_ami_error:
+        print(guest_ami_error, file=sys.stderr)
+        return 1
+    updater_ami = values.updater_ami
+    updater_ami_error = aws_svc.validate_encryptor_ami(values.updater_ami)
```
bburtin: Is this the type of validation that we want to do?  Seems like we should be verifying that the encrypted AMI is one that we encrypted.

brkt_cli/__init__.py
```diff
@@ -197,6 +157,136 @@ def main():
...
+    if updater_ami_error:
+        log.error('Update failed: %s', updater_ami_error)
+        return 1
+    # Initial validation done
+    log.info('Commencing update for AMI %s', encrypted_ami)
```
bburtin: Nit: "Commencing update..." sounds wordy.  How about simply

log.info('Updating %', encrypted_ami)

Since the ID starts with "ami-", it's clear from the message that we're talking about an AMI.  Your call.

brkt_cli/update_encrypted_ami.py
```diff
@@ -0,0 +1,135 @@
...
+        encrypt_ami.NAME_METAVISOR_LOG_SNAPSHOT: mv_log_snapshot
+    }
+
+
+def retrieve_guest_volume_snapshot(aws_service,
+                                   encrypted_ami_id,
+                                   zone):
+    # Verify expected device mapping is present, return with info
+    log.info('Creating guest volume snapshot')
+    encrypted_image = aws_service.conn.get_image(encrypted_ami_id)
```
bburtin: This should be aws_service.get_image(), to maintain the service abstraction.

brkt_cli/update_encrypted_ami.py
```diff
@@ -0,0 +1,135 @@
...
+    guest_volume_mapping = \
+        encrypted_image.block_device_mapping.get('/dev/sda5')
+    if not guest_volume_mapping:
+        return None, None, None, \
+            'Invalid block device mapping: /dev/sda5 not present'
+    snapshot = aws_service.conn.get_all_snapshots(
```
bburtin: aws_service.get_snapshot()

brkt_cli/update_encrypted_ami.py
bburtin: I'm thinking the same thing as Dennis.  If we really need this logic, make sure we clean up resources that are created here.  From first glance, it looks like we could leak resources created inside this function if things go south.

brkt_cli/util.py
```diff
@@ -38,3 +35,23 @@ def is_expired(self):
...
         """
         return self.clock.time() >= self.deadline
+
+
+def make_nonce():
+    """Returns a 32bit nonce in hex encoding"""
+    return str(uuid.uuid4()).split('-')[0]
+
+
+def validate_ami_name(name):
```
bburtin: Unused?

brkt_cli/util.py
```diff
@@ -38,3 +35,23 @@ def is_expired(self):
...
+    try:
+        service.validate_image_name(name)
+        return 0
+    except service.ImageNameError as e:
+        return 1
+
+
+def validate_region(region):
```
bburtin: Use AWSService to get the list of regions.  Also, this function should return True/False, not 1/0.  Can this be in __init__.py, since it's the only callsite?

README.md
rohang06: Fixed in next diff

brkt_cli/update_encrypted_ami.py
rohang06: I thought that if a customer had deregistered their AMI, the snapshot would go away. This isn't the case, so I'll remove this logic. Fixed in next diff

brkt_cli/__init__.py
rohang06: Fixed in next diff

brkt_cli/__init__.py
rohang06: I can't see a definable way to verify that this is the one we encrypted. For example:
image.description
u'Based on ami-76e27e1e, encrypted by Bracket Computing'
image.name
u'ubuntu/images/hvm/ubuntu-trusty-14.04-amd64-server-20141125 (encrypted 0c09d5a3)'
image.tags
{u'BrktEncryptorSessionID': u'1dcd61a0', u'BrktEncryptor': u'True', u'BrktEncryptorAMI': u'ami-c62259ac'}
Are all mutable so wouldn't prove anything. Moreover, shouldn't the user be keeping track of their bracket AMIs that they want to update?

brkt_cli/__init__.py
rohang06: fixed in next diff

brkt_cli/update_encrypted_ami.py
rohang06: fixed in next diff

brkt_cli/__init__.py
bburtin: Users can make mistakes, and we need to fail early and clearly when they do.  I'm pretty sure that we tag the encrypted AMI.  Can you check the tags during validation?

brkt_cli/update_encrypted_ami.py
rohang06: fixed in next diff

brkt_cli/util.py
rohang06: fixed in next diff

brkt_cli/util.py
rohang06: fixed in next diff

brkt_cli/update_encrypted_ami.py
```diff
@@ -0,0 +1,120 @@
...
+        encrypt_ami.NAME_METAVISOR_ROOT_SNAPSHOT: mv_root_snapshot,
+        encrypt_ami.NAME_METAVISOR_GRUB_SNAPSHOT: mv_grub_snapshot,
+        encrypt_ami.NAME_METAVISOR_LOG_SNAPSHOT: mv_log_snapshot
+    }
+
+
+def retrieve_guest_volume_snapshot(aws_service,
+                                   encrypted_ami_id):
+    # Verify expected device mapping is present, return with info
+    log.info('Creating guest volume snapshot')
```
bburtin: This message is misleading.  Should be safe to omit it, since we're not modifying anything?

brkt_cli/update_encrypted_ami.py
```diff
@@ -0,0 +1,120 @@
...
+    if not guest_volume_mapping:
+        return None, None,\
+            'Invalid block device mapping: /dev/sda5 not present'
+    # Need to recover the volume iops and size
+    volume_info = {'iops': guest_volume_mapping.iops,
+                   'size': guest_volume_mapping.size,
+                   'type': guest_volume_mapping.volume_type}
+    guest_volume_snapshot = aws_service.get_snapshot(
+        guest_volume_mapping.snapshot_id)
+    return guest_volume_snapshot, volume_info, None
```
bburtin: Get rid of the 3rd value?

brkt_cli/util.py
```diff
@@ -19,11 +21,6 @@ class BracketError(Exception):
     pass
 
 
-def make_nonce():
```
bburtin: Looks like you can revert the changes to this file.

rohang06:
replaced by #78

PR #78, opened 2015-12-03
Update encrypted ami

rohang06:
This change adds the update-encrypted-ami option to the
brkt_cli. It allows a user to retrieve a metavisor
upgrader AMI and use it to upgrade an already encrypted
AMI. This provides a speed improvement for users who
wish to update their encrypted AMIs.

The user must specify the metavisor updater AMI as an
argument. This requirement will be taken away in a
near-future commit.

### DIFF
--- a/brkt_cli/aws_service.py
+++ b/brkt_cli/aws_service.py
@@ -17,9 +17,14 @@ import re
 import boto
 import boto.sts
 import logging
-
 from boto.exception import EC2ResponseError
 import time
+
+from encrypt_ami import (
+    TAG_ENCRYPTOR,
+    TAG_ENCRYPTOR_SESSION_ID,
+    TAG_ENCRYPTOR_AMI
+)
 
 PLATFORM_WINDOWS = 'windows'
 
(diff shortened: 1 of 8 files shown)


rohang06:
Opened new pull request to squash the commits of rrg_mv_upgrade_instances_updated

brkt_cli/aws_service.py
```diff
@@ -324,6 +332,22 @@ def validate_guest_ami(self, ami_id):
                 ami_id, image.hypervisor)
         return None
 
+    def validate_guest_encrypted_ami(self, ami_id):
+        # Is this the right type of AMI?
+        error = self.validate_guest_ami(ami_id)
+        if error:
+            return error
+        # Is this encrypted by Bracket?
+        image = self.conn.get_all_images([ami_id])[0]
```
bburtin: self.get_image(ami_id) or self.conn.get_image(ami_id) would be more concise

brkt_cli/aws_service.py
```diff
@@ -324,6 +332,22 @@ def validate_guest_ami(self, ami_id):
...
+        tags = image.tags
+        expected_tags = (TAG_ENCRYPTOR,
+                         TAG_ENCRYPTOR_SESSION_ID,
+                         TAG_ENCRYPTOR_AMI)
+        missing_tags = set(expected_tags) - set(tags.keys())
```
bburtin: Since this is just a sanity check, it may be less heavy-handed to check for any of the tags, as opposed to all of the tags.  Your call.

rohang06: I think we should check all tags if we are checking for them to begin with. We let the user know which tags are missing, and they can add the no-validate flag if they want to skip this.

brkt_cli/aws_service.py
rohang06: I saw that validate_guest_ami() and validate_encryptor_ami() retrieve the AMI in a similar manner

PR #80, opened 2015-12-07
Specify subnet and security group for image encryption

bburtin:
Add --subnet and --security-group options to the encrypt-ami subcommand.
The subnet is used when running both the snapshotter and encryptor
instances.  The security groups are used when running the encryptor
instance.

Clean up the command line parsing code to check the subparser name
instead of referencing argv[0].  The old code broke when passing -v to
the brkt top-level command.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -98,6 +98,15 @@ def command_encrypt_ami(values, log):
             # Validate the key pair name.
             aws_svc.get_key_pair(values.key_name)
 
+        if values.subnet_id:
+            # Validate the subnet.
+            aws_svc.get_subnet(values.subnet_id)
+
+        if values.security_group_ids:
+            # Validate the security groups.
+            for id in values.security_group_ids:
+                aws_svc.get_security_group(id, retry=False)
+
         if not values.no_validate_ami:
             error = aws_svc.validate_guest_ami(values.ami)
             if error:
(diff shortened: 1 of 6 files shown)


gdmen:
lgtm

brkt_cli/__init__.py
```diff
@@ -98,6 +98,15 @@ def command_encrypt_ami(values, log):
             # Validate the key pair name.
             aws_svc.get_key_pair(values.key_name)
 
+        if values.subnet_id:
```
rohang06: Do we also need this for updating an encrypted AMI?

brkt_cli/aws_service.py
```diff
@@ -230,17 +235,21 @@ def run_instance(self,
                      image_id,
                      security_group_ids=None,
                      instance_type='m3.medium',
-                     block_device_map=None):
+                     block_device_map=None,
+                     subnet_id=None):
         if security_group_ids is None:
             security_group_ids = []
         log.debug('Starting a new instance based on %s', image_id)
+        log.debug('Using security groups %s, subnet %s',
```
rohang06: if security_group_ids = [] or subnet_id = None the debug message won't be very useful.. Only print if both are present?

bburtin: Yes.  I filed a separate bug.

brkt_cli/__init__.py
bburtin: Yes, I filed a separate bug.

brkt_cli/aws_service.py
bburtin: Done.

PR #81, opened 2015-12-08
Create temporary security group in the same VPC as the subnet

bburtin:
When a subnet is specified but security groups are not, create the
temporary security group in the VPC of the given subnet.  When both are
specified, make sure that they are in the same VPC.

Change the return type of create_security_group(), so that it returns an
object instead of an id.  This is consistent with the other create_xxx()
methods in AWSService.

Remove log level cruft from command_encrypt_ami().  We initialize
logging in main().

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -36,18 +36,32 @@ VERSION = '0.9.6'
 log = None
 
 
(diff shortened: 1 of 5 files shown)


gdmen:
lgtm

PR #84, opened 2015-12-09
Fetch latest encryptor_instance data before use

gdmen:
The S3 encryptor version modifies the instance bdm post launch.
Once the encryptor instance is done running, update its local attributes
before reading them.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -775,6 +775,9 @@ def make_encrypted_ami(aws_svc, enc_svc_cls, encryptor_instance, encryptor_ami,
 
     log.info('Encrypted root drive is ready.')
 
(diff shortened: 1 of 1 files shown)


rohang06:
nice

brkt_cli/encrypt_ami.py
```diff
@@ -775,6 +775,9 @@ def make_encrypted_ami(aws_svc, enc_svc_cls, encryptor_instance, encryptor_ami,
 
     log.info('Encrypted root drive is ready.')
 
+    # Making a get_instance call has the same underlying cost as boto's
+    # convenience update() method.
+    encryptor_instance = aws_svc.get_instance(encryptor_instance.id)
```
bburtin: It would be more helpful to explain why we're getting the latest instance properties.  I don't think the current comment is helpful, since it just explains the internals of the API.

PR #85, opened 2015-12-09

jimvoll:
Method:

licensing info in this case.

guest virtualization type.

Testing:
python -m unittest test


brkt_cli/__init__.py
```diff
@@ -167,6 +173,8 @@ def command_update_encrypted_ami(values, log):
         except aws_service.ImageNameError as e:
             print(e.message, file=sys.stderr)
             return 1
+    else:
```
rohang06: encrypted_ami_name is set to None before checking values. encrypted_ami_name

brkt_cli/__init__.py
```diff
@@ -192,36 +200,103 @@ def command_update_encrypted_ami(values, log):
...
-    print(ami)
-    return 0
+    log.info('Updating %s with new MV %s', encrypted_ami, updater_ami)
+    guest = None
+    mv_root_id = None
+    tids = set()
+    try:
+        guest_image = aws_svc.get_image(encrypted_ami)
+        mv_image = aws_svc.get_image(updater_ami)
+        if (guest_image.virtualization_type !=
```
rohang06: looks like this check can go in aws_svc.validate_guest_encrypted_ami()

jimvoll:
Merge from hell

brkt_cli/aws_service.py
```diff
@@ -82,6 +82,10 @@ def create_snapshot(self, volume_id, name=None, description=None):
         pass
 
     @abc.abstractmethod
+    def create_volume(self, size, zone, **kwargs):
```
bburtin: **kwargs makes it hard for the caller to know what the valid options are.  Can we use named arguments for the methods in BaseAWSService?

brkt_cli/encrypt_ami.py
```diff
@@ -389,59 +411,64 @@ def create_encryptor_security_group(svc):
...
+        aws_svc.create_tags(
+            bdm['/dev/sda1'].volume_id, name=NAME_METAVISOR_ROOT_VOLUME)
+        aws_svc.create_tags(
+            bdm['/dev/xvdg'].volume_id, name=NAME_ENCRYPTED_ROOT_VOLUME)
+
     return instance
 
 
-def run_snapshotter_instance(aws_svc, image_id, updater=False):
+def run_guest_instance(aws_svc, image_id, updater=False):
```
bburtin: I've been using "snapshotter" and "snapshot creator", to explain why the guest instance is being started.  Can we keep the existing name, so the terminology is consistent with other parts of the codebase?

brkt_cli/encrypt_ami.py
```diff
@@ -494,6 +521,17 @@ def _snapshot_root_volume(aws_svc, instance, image_id):
     )
     wait_for_snapshots(aws_svc, snapshot.id)
 
+    # Now try to detach the root volume.
+    log.info('Detaching root volume %s %s' % (root_vol.volume_id, instance.id))
```
bburtin: Would it read better to say 'Detaching root volume %s from %s'?

brkt_cli/encrypt_ami.py
```diff
@@ -494,6 +521,17 @@ def _snapshot_root_volume(aws_svc, instance, image_id):
...
+    aws_svc.detach_volume(
+        root_vol.volume_id,
+        instance_id=instance.id,
+        force=True
+    )
+    # And now delete it
+    log.info('Deleting root volume %s %s' % (root_vol.volume_id, instance.id))
```
bburtin: The instance id doesn't seem necessary here, since the volume has been detached.

brkt_cli/encrypt_ami.py
jimvoll: But in the new world, that's not why they are being started. We create the 'guest' image because
it is used as the base to create the new AMI (saving product codes). In the old world, it was used
to create a root volume snapshot we could use later.

brkt_cli/encrypt_ami.py
jimvoll: Sure. I'll fix that

brkt_cli/encrypt_ami.py
jimvoll: O.K. will fix

brkt_cli/aws_service.py
jimvoll: Not sure I understand this comment. You want me to list all the options that are available in boto for this call?

jimvoll:
I thought I got rid of this whole pull request because I had to remege a ton of other changes.
I'll be resubmitting a pull request based on something without conflicts.

brkt_cli/aws_service.py
bburtin: Not all of them.  We only care about a small number of arguments, so just add those.

brkt_cli/encrypt_ami.py
bburtin: If you don't like "snapshotter," can you come up with a better name and update the tags accordingly?  If cleanup fails, I want the customer to know what the resource is from looking at the tags.  "Guest" is kind of generic, but I'm struggling to come up with a better name.

PR #86, opened 2015-12-10
Remove default for brkt-env option

robertabbott:
Brkt-env now defaults to an empty string
instead of 'prod'. The MV will use s3 if
an empty string is passed.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -439,11 +439,14 @@ def tag_encryptor_volumes(aws_svc, instance, update_ami=False):
 
 
 def run_encryptor_instance(aws_svc, encryptor_image_id, snapshot, root_size,
-                           guest_image_id, brkt_env='prod', security_group_ids=None,
+                           guest_image_id, brkt_env=None, security_group_ids=None,
                            subnet_id=None, update_ami=False):
     bdm = BlockDeviceMapping()
     user_data = {}
-    user_data['brkt'] = {'yeti_endpoints':brkt_env}
+    if brkt_env:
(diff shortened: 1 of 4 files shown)


brkt_cli/encrypt_ami.py
```diff
@@ -845,7 +845,7 @@ def make_encrypted_ami(aws_svc, enc_svc_cls, encryptor_instance, encryptor_ami,
     return ami_info
 
 
-def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env='prod',
+def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env='',
```
bburtin: The usual convention is that None means "use default value."  Would that make more sense than an empty string?

brkt_cli/encrypt_ami_args.py
```diff
@@ -51,9 +51,12 @@ def setup_encrypt_ami_args(parser):
     )
 
     # Optional yeti endpoints. Hidden because it's only used for development
+    # if you're using this option, it should be passed as a comma separated list
+    # of endpoints. ie blb.*.*.brkt.net:7002,blb.*.*.brkt.net:7001 the endpoints
+    # must also be in order: api_host,hsmproxy_host
     parser.add_argument(
         '--brkt-env',
-        default='prod',
+        default='',
```
bburtin: If you use brkt_env=None in run_encryptor_instance(), this default becomes unnecessary.

robertabbott: Done. Removed the default values.

PR #89, opened 2015-12-11
Rewrite inline comment.

gdmen:
Replace unnecessary comment with one that explains why we're doing what
we're doing.


bburtin:
Ship it.

PR #90, opened 2015-12-11
Update update-encrypted-ami usage

bburtin:
Use "ID" instead of AMI_ID, UPDATER_AMI_ID to be consistent with the
usage of encrypt-ami.


gdmen:
lgtm

PR #92, opened 2015-12-12
Try both public and private IPs when getting encryption status

bburtin:
When the public and private IP are both specified for an instance, try
both until one works.


brkt_cli/encryptor_service.py
```diff
@@ -58,14 +73,36 @@ def is_encryptor_up(self):
...
+            elif info['bytes_total'] > 0:
+                ratio = float(info['bytes_written']) / info['bytes_total']
+                info['percent_complete'] = int(100 * ratio)
+            successful_hostname = hostname
+            break
+
+        if info:
+            # Don't try the other hostnames again, now that we have one that
+            # is known to work.
+            self.hostnames = [successful_hostname]
```
rohang06: Is there any danger in assuming successful_hostname will always be successful? Would it be safer to try all hosts all the time?

bburtin: Not that I can think of.  Both IPs map to the same host.  Once one works, it should continue to work.  But feel free to convince me that I'm wrong.  Networking is not my strong suit.

rohang06: I can't think of a strongly convincing argument besides the fact that we are try()ing IPs presumably because they are supposed to work, but aren't. So why would we assume that an IP that had worked will continue to do so.

It's up to you though

bburtin:
Closing in order to resolve merge conflicts.

PR #93, opened 2015-12-14

jimvoll:
Method:

licensing info in this case.

guest virtualization type.

Testing:
python -m unittest test


brkt_cli/__init__.py
```diff
@@ -26,12 +27,13 @@
...
 from brkt_cli import encrypt_ami_args
-from brkt_cli import update_encrypted_ami
 from brkt_cli import update_encrypted_ami_args
 from brkt_cli import encryptor_service
 from brkt_cli import util
 
-VERSION = '0.9.7'
+from update_ami import update_ami
+
+VERSION = '0.9.8'
```
bburtin: Can you change the version to '0.9.8pre1' or something else that follows the setuptools versioning spec:

https://pythonhosted.org/setuptools/setuptools.html#specifying-your-project-s-version

I'd like the user to know they're running prerelease code, to avoid tech support confusion.

jimvoll: WIll fix.

brkt_cli/__init__.py
```diff
@@ -211,56 +222,23 @@ def command_update_encrypted_ami(values, log):
...
+        mv_image.virtualization_type):
+        log.error("Encryptor virtualization_type mismatch")
         return 1
+    if not values.encrypted_ami_name:
+        # Replace nonce in AMI name
+        name = guest_image.name
+        m = re.match('(\S+) \(encrypted (\S+)\)', name)
+        encrypted_ami_name = m.group(1) + ' (encrypted %s)' % (nonce,)
+    # Initial validation done
+    log.info('Updating %s with new MV %s', encrypted_ami, updater_ami)
```
bburtin: Since this is visible to customers, we should either spell out "metavisor" or use whatever the official name is.

brkt_cli/encrypt_ami.py
```diff
@@ -59,10 +61,15 @@
...
 # that the user will see in his or her EC2 console.
 
-# Snapshotter instance names.
-NAME_SNAPSHOT_CREATOR = 'Bracket root snapshot creator'
-DESCRIPTION_SNAPSHOT_CREATOR = \
-    'Used for creating a snapshot of the root volume from %(image_id)s'
+# Guest instance names.
+NAME_GUEST_CREATOR = 'Bracket guest'
+DESCRIPTION_GUEST_CREATOR = \
+    'Used to creating a encrypted guest root volume from %(image_id)s'
```
bburtin: Typo.

brkt_cli/encrypt_ami.py
```diff
@@ -858,23 +714,35 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env=None,
...
+    )
+    if (mv_image.virtualization_type == 'hvm' and
+        guest_image.virtualization_type == 'paravirtual'):
+            raise BracketError(
+                    "Encryptor/Guest virtualization type mismatch")
+    legacy = False
+    if (mv_image.virtualization_type == 'paravirtual' and
+        guest_image.virtualization_type == 'hvm'):
+            # This will go away when HVM MV GA's
+            log.info("Legacy mode: Can't preserve guest license data")
```
bburtin: Maybe make this message more specific, so that the user knows why we're not preserving guest license data?

brkt_cli/encrypt_ami.py
jimvoll: Fixed to
'Used to create an encrypted guest root volume from %(image_id)s'

brkt_cli/encrypt_ami.py
jimvoll: Hopefully this is very short-lived, but would this be better:

            log.warn("Must specify a paravirtual AMI type in order to "
                     "preserve guest license information")

bburtin: Sure, sounds good.

PR #95, opened 2015-12-14

jimvoll:
Method:

licensing info in this case.

guest virtualization type.

Testing:
python -m unittest test


bburtin:
Looks good.  Merge away!

PR #97, opened 2015-12-15
Detect duplicate AMI name

bburtin:
When --encrypted-ami-name is specified, see if there's already an AMI
with the given name before proceeding.  Move a bunch of validation code
into a separate function, so that it's called from both
command_encrypt_ami() and command_update_encrypted_ami().  Also, we can
now unit test the validation code.


rohang06:
Nice, lgtm

rohang06:
Ok makes sense.. lgtm

PR #102, opened 2015-12-18
Support subnet and security groups when updating an encrypted AMI

bburtin:
Add --subnet and --security-group options to the update-encrypted-ami
command.

Add a unit test for update_ami(). Update the application logic to always
call aws_svc instead of the .update() method on the object, so that
things work as expected for both production and unit tests.

Returns the new AMI id from update_ami(), and prints it to stdout, like
we do with encrypt().


rohang06:
nice, lgtm

PR #103, opened 2015-12-22
Use 'sd' names for HVM instances. More validation of guest AMIs

jimvoll:
The AWS GUI for HVM instances requires the use of 'sd' names as
attach points even though these names are encoded in Xen and
end up as 'xvd' names on the guest/NetBSD. If we use 'sd' names
we can avoid the common error case of the user picking 'sdf'
as an attach point even though this is being used by the metavisor
as 'xvdf'

Also handle a case where the root_device_name is not found
in the block_device_mapping for an AMI. We can't encrypt this
AMI in that form. Typically the root device name is '/dev/sda1'
and you'll find that in the block device map. If that's not the
case we can't rip out old volumes and replace them with new
volumes due to a bug in AWS (when you remove all the drives
from a stopped instance it still believes the root_device_name
is in use). If we hit this case, we'll encrypt using 'legacy'
mode but not be able to preserve any OS license codes. A warning
to that effect is given prior to encryption.

Don't create explicit snapshots for the boot volume when doing
an update-ami - create_image does that already. Just move
the boot volume from the updater to the guest instead of
creating a volume from a (new) snapshot. Faster this way too.

Finally tag the metavisor boot volume explicitly. It's created
as part of the create_image call.


brkt_cli/__init__.py
```diff
@@ -181,8 +181,17 @@ def command_update_encrypted_ami(values, log):
...
-        encrypted_ami_name = m.group(1) + ' (encrypted %s)' % (nonce,)
+        m = re.match('(.+) \(encrypted (\S+)\)', name)
+        if m:
+            encrypted_ami_name = m.group(1) + ' (encrypted %s)' % (nonce,)
+        else:
+            encrypted_ami_name = name + ' (encrypted %s)' % (nonce,)
+        filters = {'name': encrypted_ami_name}
+        if aws_svc.get_images(filters=filters):
+            raise ValidationError(
+                    'There is already an image named %s' %
```
bburtin: Have you merged the latest code? I see this error message at line 111 in master.

brkt_cli/encrypt_ami.py
```diff
@@ -932,9 +936,10 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env=None,
             aws_svc.attach_volume(
                 mv_root_id,
                 guest_instance.id,
-                root_device_name
+                root_device_name,
             )
             log.info("Done attaching new root: %s" % (root_device_name,))
+            guest_instance.update()
```
bburtin: Please use guest_instance = aws_svc.get_instance(guest_instance.id), so that the code is compatible with unit tests.

brkt_cli/update_ami.py
```diff
@@ -177,32 +175,33 @@ def update_ami(aws_svc, encrypted_ami, updater_ami,
...
+        log.info("Attaching new metavisor boot disk: %s to %s" %
             (mv_root_id, encrypted_guest.id)
         )
-        aws_svc.attach_volume(mv_root_id, encrypted_guest.id, '/dev/sda1')
-        encrypted_guest = aws_svc.get_instance(encrypted_guest.id)
-        guest_bdm['/dev/sda1'] = \
-            encrypted_guest.block_device_mapping['/dev/sda1']
-        guest_bdm['/dev/sda1'].delete_on_termination = True
+        aws_svc.attach_volume(mv_root_id, encrypted_guest.id, root_device_name)
+        encrypted_guest.update()
```
bburtin: Please use aws_svc.get_instance().

bburtin:
This looks good to me except for the issues that I commented on. Can you have someone from storage take a look? I'm not qualified to review the device mapping code.

brkt_cli/__init__.py
jimvoll: I just got done merging the latest code in master. What error are you
seeing??

brkt_cli/encrypt_ami.py
jimvoll: o.k. I guess that doesn't make a difference in real behavior. I'll fix the
code to match the unit test

PR #105, opened 2015-12-22
Use 'sd' names for HVM instances. More validation of guest AMIs

jimvoll:
The AWS GUI for HVM instances requires the use of 'sd' names as
attach points even though these names are encoded in Xen and
end up as 'xvd' names on the guest/NetBSD. If we use 'sd' names
we can avoid the common error case of the user picking 'sdf'
as an attach point even though this is being used by the metavisor
as 'xvdf'

Also handle a case where the root_device_name is not found
in the block_device_mapping for an AMI. We can't encrypt this
AMI in that form. Typically the root device name is '/dev/sda1'
and you'll find that in the block device map. If that's not the
case we can't rip out old volumes and replace them with new
volumes due to a bug in AWS (when you remove all the drives
from a stopped instance it still believes the root_device_name
is in use). If we hit this case, we'll encrypt using 'legacy'
mode but not be able to preserve any OS license codes. A warning
to that effect is given prior to encryption.

Don't create explicit snapshots for the boot volume when doing
an update-ami - create_image does that already. Just move
the boot volume from the updater to the guest instead of
creating a volume from a (new) snapshot. Faster this way too.

Finally tag the metavisor boot volume explicitly. It's created
as part of the create_image call.


agtucker:
Changes look good.

PR #109, opened 2015-12-29
Improve security group validation, add --no-validate option

bburtin:
When the subnet isn't specified, make sure that the specified security
groups are in the default VPC.  Add a get_default_vpc() method to
AWSService.

Replace the --validate-ami and --no-validate-ami options with a new
--no-validate option that controls validation of AMIs, subnets, and
security groups.  Hide the old options for the short term so that we
don't break backward compatibility.

Add support for a default VPC to the test framework.  Create test
security groups in that VPC when the VPC id is not specified.


bburtin:
Squashing commits.  Will reopen.

PR #111, opened 2016-01-03
Make sure we refresh instance state after waitng for instance to stop

jimvoll:
Otherwise the block device mapping shows up as {} and we get a key
error with '/dev/xvda' when snapshotting the root volume


berndtbrkt:
wait for instance should probably just return the updated instance, but this is a OK.

PR #112, opened 2016-01-04
Improve security group validation, add --no-validate option

bburtin:
When the subnet isn't specified, make sure that the specified security
groups are in the default VPC.  Add a get_default_vpc() method to
AWSService.

Replace the --validate-ami and --no-validate-ami options with a new
--no-validate option that controls validation of AMIs, subnets, and
security groups.  Hide the old options for the short term so that we
don't break backward compatibility.

Add support for a default VPC to the test framework.  Create test
security groups in that VPC when the VPC id is not specified.


brkt_cli/aws_service.py
```diff
@@ -529,6 +533,12 @@ def detach_volume(self, vol_id, instance_id=None, force=True):
     def attach_volume(self, vol_id, instance_id, device):
         return self.conn.attach_volume(vol_id, instance_id, device)
 
+    def get_default_vpc(self):
+        vpcs = self.conn.get_all_vpcs(filters={'is-default': 'true'})
```
rohang06: is it possible to return more than one vpc here? i.e. a user misconfiguration

brkt_cli/__init__.py
```diff
@@ -59,7 +59,27 @@ def _validate_subnet_and_security_groups(aws_svc,
...
 
-    return len(vpc_ids) <= 1
+    if len(vpc_ids) > 1:
+        raise ValidationError(
+            'Subnet and security groups must be in the same VPC.')
+
+    if not subnet_id and vpc_ids:
+        (vpc_id, ) = vpc_ids
+        default_vpc = aws_svc.get_default_vpc()
+        log.debug(
```
rohang06: check for default_vpc being None? And are we concerned about the scenario where default_vpc and vpc_id are both None?

brkt_cli/aws_service.py
bburtin: Yeah, but the code will still work.  We'll return the first element in the list.

brkt_cli/__init__.py
bburtin: I check for it on the next line.

rohang06:
Cool, lgtm

bburtin:
I'm going to add some comments to address Rohan's questions and will open a new pull request.

PR #116, opened 2016-01-07
Refactor to expose functions for third party use.

gdmen:
TODO: add a small unit test that checks for the methods & signatures that are used by cephalopod. Test fails if they are changed in a backwards incompatible way.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -274,10 +274,14 @@ def wait_for_encryption(enc_svc,
         # Log progress once a minute.
         now = time.time()
         if now - last_log_time >= 60:
-            state_display = 'Encryption'
-            if state == encryptor_service.ENCRYPT_DOWNLOADING:
-                state_display = 'Download from S3'
-            log.info('%s is %d%% complete', state_display, percent_complete)
+            if state == encryptor_service.ENCRYPT_INITIALIZING:
+                log.info('Encryption process is initializing')
+            else:
+                state_display = 'Encryption'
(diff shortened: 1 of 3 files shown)


brkt_cli/encryptor_service.py
```diff
@@ -17,6 +17,7 @@
 import logging
 import urllib2
 
+ENCRYPT_INITIALIZING = 'initial'
```
bburtin: Nit: "initial" seems inconsistent with "downloading" and "encrypting".

gdmen: That's what the encryptor instance already sends. I'm just adding client side handling for the state that we didn't have before. I'll change this on the avatar side the next time I'm making a change there. Not worth a NUC imo.

bburtin:
Looks good.

gdmen:
Minor fixes after testing HVM & PV encryption. Unit tests coming later today.

gdmen:
Unit test added to check backwards compatibility with current outside usage.

brkt_cli/encrypt_ami.py
```diff
@@ -713,9 +717,229 @@ def log_exception_console(aws_svc, e, id):
         )
 
 
+def get_encrypted_ami_kwargs(aws_svc, enc_svc_cls, encryptor_instance,
```
bburtin: I'm not sure about the kwargs thing.  kwargs make it hard to know what is being returned by the first function, and what's being passed into the second.  Can this function return a tuple instead?

bburtin:
Looks good.  Ship it!

PR #118, opened 2016-01-12
Explictly add public IP to all instances.

gdmen:


bburtin:
Can you please update your commit description so that we remember why you're making this change?

bburtin:
Now that I think about it, it would probably be even better to put the reasoning in the comments in the code.

gdmen:
Shelving this for now unless it turns out we really need it.

bobveznat:
I think most people will want to launch private instances in private subnets with only private IPs. Having public IPs show up on instances will be pretty scary for a lot of people.

Given that this idea is being shelved its kind of no big deal. Before doing something like this in the future I think we should discuss it more.

gdmen:
Yeah, that's the reason I closed it. And the only case this would improve would be when customers have a subnet that they don't have local access to and that defaults to no public IP.

PR #122, opened 2016-01-15
Get encryptor AMI automatically during update

bburtin:
Get the encryptor AMI from the service during update-encrypted-ami.
Remove the --updater-ami option and add a hidden --encryptor-ami option
for development.  Remove the extra call to validate_encryptor_ami().
The validation is already being done inside _connect_and_validate().


rohang06:
Nice, lgtm

PR #123, opened 2016-01-19
Metavisor requires the root disk to be attached at "/dev/sda1".

apoorva-jindal:
If the guest AMI's root disk is attached at at different
attach-point, then treat as legacy mode which will use
the metavisor instance to create the AMI instead of the guest
instance. However, for such AMIs, the guest license information
will be lost.

Testing.
Works for Centos 7.1 AMI ami-7bc3f94b which attaches root disk
at /dev/sda.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -972,6 +972,10 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env=None,
             log.warn("AMI must have root_device_name in block_device_mapping "
                     "in order to preserve guest OS license information")
             legacy = True
+    if (guest_image.root_device_name != "/dev/sda1"):
+        log.warn("The root device for the guest AMI should be /dev/sda1 "
+                 "in order to preserve guest OS license information")
+        legacy = True
     try:
         guest_instance = run_guest_instance(aws_svc,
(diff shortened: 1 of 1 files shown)


brkt_cli/encrypt_ami.py
```diff
@@ -972,6 +972,10 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, brkt_env=None,
             log.warn("AMI must have root_device_name in block_device_mapping "
                     "in order to preserve guest OS license information")
             legacy = True
+    if (guest_image.root_device_name != "/dev/sda1"):
+        log.warn("The root device for the guest AMI should be /dev/sda1 "
```
bburtin: Will the end user know what to do when he sees this error message?

apoorva-jindal: They need to create a new guest AMI with the root disk attached to /dev/sda1. Let me know if it is not clear from the message.

bburtin: I would word it like "Operating system license information will not be preserved because the root disk is <root_device_name> instead of /dev/sda1."  The current wording makes me think "something is wrong, but we're proceeding anyway."  That said, I'm anything but an expert in device mappings and building AMIs.

apoorva-jindal: Sure, will update the error message.

PR #124, opened 2016-01-19
Metavisor requires the root disk to be attached at "/dev/sda1".

apoorva-jindal:
If the guest AMI's root disk is attached at at different
attach-point, then treat as legacy mode which will use
the metavisor instance to create the AMI instead of the guest
instance. However, for such AMIs, the guest license information
will be lost.

Testing.
Works for Centos 7.1 AMI ami-7bc3f94b which attaches root disk
at /dev/sda.


bburtin:
Looks good, thanks!  Merge away.

PR #125, opened 2016-02-01
Add command descriptions

bburtin:
Add descriptions to the top level brkt command and the encrypt-ami and
update-encrypted-ami subcommands.


rohang06:
nice, lgtm

PR #126, opened 2016-02-02
Validate region before getting the encryptor AMI

bburtin:
Validate the region before attempting to get the encryptor AMI.  This
avoids errors like "No AMI for us-west-3 returned" when we attempt to
look up the encryptor AMI for an invalid region.


gdmen:
lgtm, would be nice to squash

PR #129, opened 2016-02-05
Add support for custom tags

bburtin:
Add a new --tag option to encrypt-ami and update-encrypted-ami.  This
allows the user to specify a custom tag that is set on every resource
created by encryption.  --tag may be specified multiple times, in order
to set multiple tags.

Get rid of ImageNameError.  Update the validation code to throw
ValidationError when validation of command-line options fails.  Move
ValidationError into a new validation module, to avoid circular
dependencies between aws_service and brkt_cli.


brkt_cli/__init__.py
```diff
@@ -144,6 +138,25 @@ def _connect_and_validate(aws_svc, values, encryptor_ami_id):
...
+    """
+    if not tag_strings:
+        return {}
+
+    tags = {}
+    for s in tag_strings:
+        m = re.match(r'([^=]+)=(.+)', s)
+        if not m:
+            raise ValidationError('Tag %s is not in the format KEY=VALUE' % s)
+        tags[m.group(1)] = m.group(2)
```
rohang06: should this be 0 and 1, not 1 and 2?

bburtin: Tests pass, so I'm pretty sure it's right.  :-)  I think group 0 is the entire matched string.  Group 1 is the first set of parens.

rohang06:
Nice, lgtm

PR #132, opened 2016-02-05
Remove --validate-ami and --no-validate-ami options

bburtin:
Remove options that were deprecated in 0.9.10.


rohang06:
nice, lgtm

PR #134, opened 2016-02-06
Don't attempt to encrypt an already encrypted image

bburtin:
When checking the guest AMI, fail validation if we find tags that were
created during image encryption.

Refactor the validation code, so that it's more testable.  Move
validation logic out of AWSService and into brkt_cli.  This thins out
the AWSService implementation, and ensures that the validation logic
gets unit tested.

Add switchable retry logic to AWSService.get_image(), so that we don't
retry the operation for AMIs specified on the command line.  These AMIs
were created long enough ago that retry is not necessary.


mpage:
LGTM

PR #135, opened 2016-02-08
Add encryptor AMI tag when updating

bburtin:
When updating an encrypted AMI, set the BrktEncryptor tag on the AMI and
any temporary AWS resources.

Move the call to _validate_guest_ami() into command_encrypt_ami().  This
call started failing for update-encrypted-ami, since we now disallow
double-encryption.


rohang06:
nice, lgtm

PR #136, opened 2016-02-09
Check Python version during setup

bburtin:
Checking inside main() was too late in the process.  Setup would fail on
    Python 2.6 because it couldn't load version 2.7.0 of the requests
    library.


gdmen:
lgtm

PR #138, opened 2016-02-11
Check version on startup

bburtin:
When starting brkt-cli, check the code version against the list of
available versions on PyPI.  Exit if the code version is earlier than
the earliest available version.  Add a --no-check-version option to skip
the version check.


gdmen:
lgtm

bburtin:
Will open a new pull request.

PR #140, opened 2016-02-11
Disable sriov support.

apoorva-jindal:
1. Metavisor does not support sriov. So, if guest instance has sriov 
support, then use legacy mode to create the encrypted ami.
2. Fix the HVM legacy path by removing the unencrypted and encrypted
guests disks from the encryptor instance.


apoorva-jindal:
Closing to fix the tests.

PR #143, opened 2016-02-16
Guest root volume should be set to GP2

krishnanrs:
Read the source volume type from the right place and default the volume type to GP2.


jimvoll:
Looks good to me

PR #145, opened 2016-02-16
Add hidden --hvm flag

dramdass:
Get hvm encryptor amis from the hvm_amis.json file in the S3
bucket when --hvm is specified.


rohang06:
lgtm, nice

jimvoll:
Don't we need to handle the 'update-ami' sub-command also?

dramdass:
@jimvoll I did update that subcommand as well in this review. Are there other places I need to change as well?

PR #146, opened 2016-02-20

apoorva-jindal:

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -191,7 +191,8 @@ def command_encrypt_ami(values, log):
         encrypted_ami_name=values.encrypted_ami_name,
         subnet_id=values.subnet_id,
         security_group_ids=values.security_group_ids,
-        brkt_env=values.brkt_env
+        brkt_env=values.brkt_env,
+        ntp_server=values.ntp_server
     )
     # Print the AMI ID to stdout, in case the caller wants to process
     # the output.  Log messages go to stderr.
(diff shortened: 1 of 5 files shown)


brkt_cli/encrypt_ami_args.py
```diff
@@ -54,6 +54,15 @@ def setup_encrypt_ami_args(parser):
             'May be specified multiple times.'
         )
     )
+    parser.add_argument(
+        '--ntp-server',
+        metavar='DNS Name',
+        dest='ntp_server',
+        action='append',
+        help=(
+            'NTP server to sync Metavisor clock.'
```
bburtin: Can you append "May be specified multiple times." so that the behavior is clearer?

brkt_cli/__init__.py
```diff
@@ -191,7 +191,8 @@ def command_encrypt_ami(values, log):
         encrypted_ami_name=values.encrypted_ami_name,
         subnet_id=values.subnet_id,
         security_group_ids=values.security_group_ids,
-        brkt_env=values.brkt_env
+        brkt_env=values.brkt_env,
+        ntp_server=values.ntp_server
```
bburtin: Is this supposed to be a single value or multiple values?  The argument is specified as "append", which would return a list.

brkt_cli/encrypt_ami_args.py
```diff
@@ -54,6 +54,15 @@ def setup_encrypt_ami_args(parser):
             'May be specified multiple times.'
         )
     )
+    parser.add_argument(
+        '--ntp-server',
+        metavar='DNS Name',
+        dest='ntp_server',
```
bburtin: We should call this "ntp_servers" (plural) if it's possible to specify multiple NTP servers.

PR #147, opened 2016-02-22

apoorva-jindal:


bburtin:
Looks good.  Merge away!

PR #148, opened 2016-02-23

apoorva-jindal:
Error out if ntp-server is not a valid host name or ipv4 address.
Finally, handle errors reported by the Bracket Updater instance.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -93,6 +93,32 @@ def _validate_subnet_and_security_groups(aws_svc,
                 )
 
 
+def _validate_dns_name_ip_address(hostname):
+    """ Verifies that the input hostname is indeed a valid
+    host name or ip address
+
+    :return True if valid, returns False otherwise
+    """
+    # ensure length does not exceed 255 characters
+    if (len(hostname) > 255):
+        return False
+    # remove the last dot from the end
+    if (hostname[-1] == "."):
+        hostname = hostname[:-1]
+    valid = re.compile("(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
+    return all(valid.match(x) for x in hostname.split("."))
+
+
+def _validate_ntp_servers(ntp_servers):
+    if ntp_servers is None:
+        return
+    for server in ntp_servers:
+        if (_validate_dns_name_ip_address(server) is False):
+            raise ValidationError(
+                'Invalid ntp-server %s configured. '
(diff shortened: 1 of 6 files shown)


brkt_cli/__init__.py
```diff
@@ -93,6 +93,32 @@ def _validate_subnet_and_security_groups(aws_svc,
...
+    if (len(hostname) > 255):
+        return False
+    # remove the last dot from the end
+    if (hostname[-1] == "."):
+        hostname = hostname[:-1]
+    valid = re.compile("(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
+    return all(valid.match(x) for x in hostname.split("."))
+
+
+def _validate_ntp_servers(ntp_servers):
```
bburtin: I'm skeptical that the added complexity is necessary.  Even if the patterns match, there's no guarantee that the host really is an NTP server.  And there's the possibility of bugs in the hostname checking code.

If you really want to make sure that the thing is an NTP server, you could use ntplib to try to query it:

https://pypi.python.org/pypi/ntplib

But I'm not sure if the benefit is great enough that it justifies adding another 3rd party dependency.

jimvoll:
I disagree with Boris on this one. I think we can, and should do some error checking here
or the end result will be a failed launch on the resulting instance with little to go on.

For instance, I asked Approva to make sure the argument being passed wasn't IPv6
for instance. We should try to catch obvious typo's.

Using ntplib might not be a good answer here either. You might not be able to contact
the NTP servers specified as part of the encryption process for instance.

brkt_cli/__init__.py
```diff
@@ -93,6 +93,32 @@ def _validate_subnet_and_security_groups(aws_svc,
...
+    if ntp_servers is None:
+        return
+    for server in ntp_servers:
+        if (_validate_dns_name_ip_address(server) is False):
```
bburtin: `if not _validate_dns_name_ip_address(server):`
would read better

brkt_cli/__init__.py
```diff
@@ -93,6 +93,32 @@ def _validate_subnet_and_security_groups(aws_svc,
...
+            raise ValidationError(
+                'Invalid ntp-server %s configured. '
```
bburtin: "configured" is not really the right word here.  I'd say either use "specified" or take it out altogether.

bburtin:
Ok.  Apoorva, if you want to keep the hostname checking code, please add a unit test.

PR #150, opened 2016-02-23

apoorva-jindal:
Error out if ntp-server is not a valid host name or ipv4 address.
Finally, handle errors reported by the Bracket Updater instance.

### DIFF
--- a/brkt_cli/encryptor_service.py
+++ b/brkt_cli/encryptor_service.py
@@ -26,6 +26,7 @@ ENCRYPT_ENCRYPTING = 'encrypting'
 ENCRYPTOR_STATUS_PORT = 8000
 FAILURE_CODE_UNSUPPORTED_GUEST = 'unsupported_guest'
 FAILURE_CODE_AWS_PERMISSIONS = 'insufficient_aws_permissions'
+FAILURE_CODE_INVALID_NTP_SERVERS = 'invalid ntp servers'
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


brkt_cli/encryptor_service.py
```diff
@@ -26,6 +26,7 @@
 ENCRYPTOR_STATUS_PORT = 8000
 FAILURE_CODE_UNSUPPORTED_GUEST = 'unsupported_guest'
 FAILURE_CODE_AWS_PERMISSIONS = 'insufficient_aws_permissions'
+FAILURE_CODE_INVALID_NTP_SERVERS = 'invalid ntp servers'
```
bburtin: Kind of a nit, but this doesn't use underscores like the other failure codes.  Your call.

brkt_cli/encryptor_service.py
```diff
@@ -93,6 +94,8 @@ def get_status(self, timeout_secs=2):
             info['percent_complete'] = 0
             if info['state'] == ENCRYPT_SUCCESSFUL:
                 info['percent_complete'] = 100
+            elif info['state'] == ENCRYPT_FAILED:
+                info['percent_complete'] = 0
```
bburtin: This looks gratuitous.  We already default to 0 three lines above.

apoorva-jindal: The issue is not setting percent_complete to 0, but having a check for state when encryption fails. Currently, code assumes that either encryption succeeds, or bytes_total exists. For a failed scenario for updater instance, bytes_total will not be present, hence will result in an exception and the brkt-cli will keep looping for ever. I can remove the info['percent_complete'] = 0 from there and have an empty elif check to prevent the check for 'bytes_total' or leave it as is. Let me know which one you prefer.

PR #153, opened 2016-02-24
Don't allow updating an encrypted AMI with the same metavisor AMI

bburtin:
When updating an encrypted AMI, check the value of its BrktEncryptorAMI
tag.  Don't proceed with the update if the tag value matches the
proposed metavisor AMI ID.  We skip this check if the user specifies
--no-validate.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -256,21 +256,30 @@ def _validate_guest_ami(aws_svc, ami_id):
     return None
 
 
-def _validate_guest_encrypted_ami(aws_svc, ami_id):
+def _validate_guest_encrypted_ami(ami, encryptor_ami_id):
     """ Validate that this image was encrypted by Bracket by checking
         tags.
-    :return: None if we recognize the image, or an error string if not
+
+    :raise: ValidationError if validation fails
     """
     # Is this encrypted by Bracket?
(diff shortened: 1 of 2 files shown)


brkt_cli/__init__.py
```diff
@@ -256,21 +256,30 @@ def _validate_guest_ami(aws_svc, ami_id):
...
                      TAG_ENCRYPTOR_AMI)
     missing_tags = set(expected_tags) - set(tags.keys())
     if missing_tags:
-        return 'Missing tags %s' % str(missing_tags)
-    return None
+        raise ValidationError(
+            '%s is missing tags: %s' % (ami.id, ', '.join(missing_tags)))
+
+    # See if this image was already encrypted by the given encryptor AMI.
+    original_encryptor_id = tags.get(TAG_ENCRYPTOR_AMI)
```
rohang06: I think we've discussed this before, but this validation would fail if a customer deleted the tags.. do we need to tell them that these tags shouldn't be modified? Other than that lgtm, nice

bburtin: I think it's a bit presumptuous to say, "You modified the tags.  Don't do that."  Probably better to just say that the tags are missing, and let the user draw his own conclusion.

PR #156, opened 2016-02-25
Test --brkt-env

bburtin:
Add a unit test that verifies that we parse the --brkt-env value and
pass the values to user_data when launching the encryptor instance.
Move run_instance_callback arguments into a class, now that we have
seven of them.


robertabbott:
lgtm

PR #157, opened 2016-02-29
Pass --brkt-env value when updating AMI

bburtin:
Parse the --brkt-env value and add it to userdata when updating an
encrypted AMI.  Break out the --brkt-env parsing logic into a separate.
Pass bracket environment data around as an object instead of as a
string.  This allows us to use the parsing code for both encryption and
updating, and also add unit tests.

Move the hostname validation function into util, so that it can be used
to validate the hostnames specified in the --brkt-env value.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -139,6 +139,22 @@ class InstanceError(BracketError):
     pass
 
 
+class BracketEnvironment(object):
+    def __init__(self):
+        self.api_host = None
+        self.api_port = None
+        self.hsmproxy_host = None
+        self.hsmproxy_port = None
+
+    def __repr__(self):
+        return '<BracketEnvironment api=%s:%d, hsmproxy=%s:%d>' % (
+            self.api_host,
+            self.api_port,
+            self.hsmproxy_host,
+            self.hsmproxy_port
+        )
+
+
 def get_default_tags(session_id, encryptor_ami):
     default_tags = {
         TAG_ENCRYPTOR: True,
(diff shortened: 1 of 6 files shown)


brkt_cli/encrypt_ami.py
```diff
@@ -483,10 +499,12 @@ def run_encryptor_instance(aws_svc, encryptor_image_id,
     bdm = BlockDeviceMapping()
     user_data = {}
     if brkt_env:
-        endpoints = brkt_env.split(',')
+        api_host_port = '%s:%d' % (brkt_env.api_host, brkt_env.api_port)
```
robertabbott: would it make sense to have methods in the BracketEnvironment class that return the address rather than having to format the string here?

brkt_cli/update_ami.py
```diff
@@ -72,7 +72,14 @@ def update_ami(aws_svc, encrypted_ami, updater_ami,
         user_data = {'brkt': {'solo_mode': 'updater'}}
         if ntp_servers:
             user_data['ntp-servers'] = ntp_servers
-        user_data = json.dumps(user_data)
+        if brkt_env:
+            api_host_port = '%s:%d' % (brkt_env.api_host, brkt_env.api_port)
```
robertabbott: this is another spot that could use a brkt_env.get_api_host() method

brkt_cli/encrypt_ami.py
bburtin: Ideally, this is a data class and should be treated like application logic.  It shouldn't be aware of the CLI format.  But you're right, it is duplicate code.  I'll break the parsing out into a separate function that returns a BracketEnvironment object.

gdmen:
lgtm

PR #158, opened 2016-03-01
Adds option for encrypting gce images

robertabbott:
Testing: encrypted a google image and verified
the encrypted guest booted

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -30,6 +30,10 @@ from brkt_cli import aws_service
 from brkt_cli import encrypt_ami
 from brkt_cli import encrypt_ami_args
 from brkt_cli import encryptor_service
+from brkt_cli import encrypt_gce_image
+from brkt_cli import encrypt_gce_image_args
+from brkt_cli import launch_gce_image
+from brkt_cli import launch_gce_image_args
 from brkt_cli import update_encrypted_ami_args
 from brkt_cli import util
 from brkt_cli.util import validate_dns_name_ip_address
@@ -196,6 +200,40 @@ def _parse_brkt_env(brkt_env_string):
     (be.api_host, be.api_port) = _parse_endpoint(endpoints[0])
     (be.hsmproxy_host, be.hsmproxy_port) = _parse_endpoint(endpoints[1])
     return be
+
+
+def command_launch_gce_image(values, log):
+    gce_svc = encrypt_gce_image.GCEService(values.project, None, log)
+    launch_gce_image.launch(log,
+                            gce_svc,
+                            values.image,
+                            values.instance_name,
+                            values.zone,
+                            values.delete_boot,
+                            {})
+    return 0
+
+
+def command_encrypt_gce_image(values, log):
+    session_id = util.make_nonce()
+    default_tags = encrypt_ami.get_default_tags(session_id,
+                                                values.encryptor_image)
+    gce_svc = encrypt_gce_image.GCEService(values.project, default_tags, log)
+    log.info('Starting encryptor session %s', gce_svc.get_session_id())
+
+    encrypted_image_id = encrypt_gce_image.encrypt(
+        gce_svc=gce_svc,
+        enc_svc_cls=encryptor_service.EncryptorService,
+        image_id=values.image,
+        encryptor_image=values.encryptor_image,
+        encrypted_image_name=values.encrypted_image_name,
+        zone=values.zone,
+        brkt_env=values.brkt_env
+    )
+    # Print the image name to stdout, in case the caller wants to process
+    # the output.  Log messages go to stderr.
+    print(encrypted_image_id)
+    return 0
 
 
 def command_encrypt_ami(values, log):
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -167,3 +167,31 @@ ami-63733e09
 
 When the process completes, the new AMI id is written to stdout.  All log
 messages are written to stderr.
+
+## Encrypting a GCE Image
+
+EXPERIMENTAL FEATURE. WILL NOT WORK FOR YOU
```
bburtin: I don't think we want this on the front page of the repo if the feature is still experimental.  Maybe we can doc it on the wiki until the code is ready for prime time?

brkt_cli/__init__.py
```diff
@@ -462,7 +507,19 @@ def main():
 
     if values.check_version:
         supported_versions = None
-
+    try:
+        if values.subparser_name == 'launch-gce-image':
+            print('you are trying to use an experimental feature. It wont work')
```
bburtin: Nits:
* Write to stderr instead of stdout.  We only write the image ID to stdout.
* Capitalize the first sentence.
* It's not clear what the experimental feature is.  It would be clearer to say something like "Warning: GCE support is still in development."

brkt_cli/__init__.py
```diff
@@ -462,7 +507,19 @@ def main():
...
+            return command_launch_gce_image(values, log)
```
bburtin: Do we need to pass the log object around?  Usually it's created at the module level.

brkt_cli/__init__.py
```diff
@@ -429,6 +467,12 @@ def main():
     )
     encrypt_ami_args.setup_encrypt_ami_args(encrypt_ami_parser)
 
+    encrypt_gce_image_parser = subparsers.add_parser('encrypt-gce-image')
```
bburtin: Is there a way to make the subparsers hidden?  Or maybe we should make this a separate top-level command while it's still in development?

brkt_cli/encrypt_gce_image.py
```diff
@@ -0,0 +1,335 @@
...
+        return self.tags[encrypt_ami.TAG_ENCRYPTOR_SESSION_ID]
+
+    def get_snapshot(self, name):
+        return self.compute.snapshots().get(project=self.project,
+                snapshot=name).execute()
+
+    def wait_snapshot(self, snapshot):
+        while True:
+            if self.get_snapshot(snapshot)['status'] == 'READY':
+                return
```
bburtin: Do we need a timeout here?  What about a sleep, to avoid hammering the Google API?

brkt_cli/__init__.py
robertabbott: all the other command_* methods are passed the log object

brkt_cli/encrypt_gce_image.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def wait_instance(self, name, zone):
+        while True:
+            instance_data = self.compute.instances().list(project=self.project,
+                    zone=zone).execute()
+            if 'items' in instance_data:
+                for i in instance_data['items']:
+                    if name == i['name'] and i['status'] == 'RUNNING':
+                        return
+            self.log.info( 'waiting for ' + name + ' to become ready')
```
bburtin: Nit: Capitalize the first word, to match the current convention.  Also, will the user know you're talking about an instance from the name?

brkt_cli/encrypt_gce_image.py
```diff
@@ -0,0 +1,335 @@
...
+            try:
+                nw = 'networkInterfaces'
+                instance = self.compute.instances().get(project=self.project,
+                        zone=zone, instance=name).execute()
+                if instance[nw][0]['accessConfigs'][0]['natIP']:
+                    return instance[nw][0]['accessConfigs'][0]['natIP']
+            except:
+                pass
+        self.log.info( "Our world is ending, we couldn't find an")
+        self.log.info( "IP address for this instance after 5 minutes.")
```
bburtin: This should be one message, not two.  Remove the leading whitespace.  And I'd remove the "world is ending" part.  It's an enterprise product, so we should keep the tone of the log messages professional (boring).

brkt_cli/encrypt_gce_image.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+class InstanceError(BracketError):
+    pass
+
+
+class GCE:
```
bburtin: Call this "GCEService", to match the convention of AWSService?

brkt_cli/encrypt_gce_image.py
```diff
@@ -0,0 +1,335 @@
...
+        user_data['brkt'] = {
+            'api_host': endpoints[0],
+            'hsmproxy_host': endpoints[1],
+        }
+    instance_name = 'brkt-guest-' + gce_svc.get_session_id()
+    encryptor = instance_name + '-encryptor'
+    encrypted_image_disk = 'encrypted-image-' + gce_svc.get_session_id()
+
+    gce_svc.run_instance(zone, instance_name, image_id)
+    log.info( 'guest instance terminated')
```
bburtin: Is this supposed to come after the call to delete_instance()?  Also, please capitalize the first letter and remove the leading whitespace.

brkt_cli/__init__.py
bburtin: Ok, do what makes sense.  I'm not sure why we're passing the log object around in other places.

brkt_cli/__init__.py
```diff
@@ -421,7 +459,7 @@ def main():
         help="Don't check whether this version of brkt-cli is supported"
     )
 
-    subparsers = parser.add_subparsers(dest='subparser_name')
+    subparsers = parser.add_subparsers(dest='subparser_name', metavar='{encrypt-ami,update-encrypted-ami}')
```
bburtin: Can you add a comment that explains why we're using metavar?

bburtin:
Looks like a good first cut.  Thanks for cleaning up all of the nits.  Go ahead and merge once the unit tests are working again.

brkt_cli/encrypt_gce_image.py
robertabbott: fixed

brkt_cli/__init__.py
robertabbott: subparsers are now hidden

PR #160, opened 2016-03-07
Encrypt_gce_image handles brkt-env option

robertabbott:
Testing: used the brkt-env option when
encrypting an image and verified the correct
endpoints were used


bburtin:
Looks good.

PR #164, opened 2016-03-14
Wait for root disk to attach

bburtin:
After attaching the encrypted root volume to the guest instance, wait
for the device mapping to become available before continuing.  Add a
unit test for this code path.  Print more details when a virtualization
type mismatch occurs.


bburtin:
Jim approved.  Merging.

PR #165, opened 2016-03-17
Pass HTTP proxy config during encryption

bburtin:
Add a new --proxy option for specifying one or more HTTPS proxies to use
during encryption.  Proxies are specified as host:port.  When proxies
are specified, we generate a proxy.yaml file and pass it to the
encryptor instance via user data.

To do:
* Specify HTTPS proxy when updating an AMI
* Read proxy.yaml from a file


bburtin:
Opening a new pull request based on feedback.

PR #166, opened 2016-03-17
Pass HTTP proxy config during encryption

bburtin:
Add a new --proxy option for specifying one or more HTTPS proxies to use
during encryption.  Proxies are specified as host:port.  When proxies
are specified, we generate a proxy.yaml file and pass it to the
encryptor instance via user data.

To do:
* Specify HTTPS proxy when updating an AMI
* Read proxy.yaml from a file

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -36,6 +36,7 @@ from brkt_cli import launch_gce_image
 from brkt_cli import launch_gce_image_args
 from brkt_cli import update_encrypted_ami_args
 from brkt_cli import util
+from brkt_cli.proxy import Proxy
 from brkt_cli.util import validate_dns_name_ip_address
 from brkt_cli.validation import ValidationError
 from encrypt_ami import (
@@ -175,7 +176,9 @@ def _parse_tags(tag_strings):
 
 
 def _parse_brkt_env(brkt_env_string):
-    """ Parse the --brkt-env value
+    """ Parse the --brkt-env value.  The value is in the following format:
+
+    api_host:port,hsmproxy_host:port
 
     :return: a BracketEnvironment object
     :raise: ValidationError if brkt_env is malformed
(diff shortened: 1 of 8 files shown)


brkt_cli/user_data.py
```diff
@@ -0,0 +1,151 @@
...
+import re
+from email import charset
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+import yaml
+
+# The directory for files saved on the Metavisor. We require that the dest
+# path for all --brkt-files be within this directory.
+BRKT_FILE_PREFIX = '/var/brkt/instance_config'
```
dramdass: Do we use this? Also I think we're technically putting the yaml file in /var/brkt/ami_config/ here

dramdass:
Code looks good! Nice job on the tests too. One minor comment

test.py
```diff
@@ -1248,3 +1252,55 @@ def test_is_later_version_available(self):
...
+
+class TestProxy(unittest.TestCase):
+
+    def test_parse_proxies(self):
+        """ Test parsing host:port strings to Proxy objects.
+        """
+        # Valid
+        proxies = brkt_cli._parse_proxies(
+            'example1.com:8001',
+            'example2.com:8002'
```
bonzotronix: Might as well add an IP address or two here.  That'll be the most common case, FWIW.

brkt_cli/user_data.py
bonzotronix: Yep.  /var/brkt/ami_config/

brkt_cli/__init__.py
```diff
@@ -285,6 +310,11 @@ def command_encrypt_ami(values, log):
     if values.brkt_env:
         brkt_env = _parse_brkt_env(values.brkt_env)
 
+    proxy_config = None
+    if values.proxies:
+        proxies = _parse_proxies(*values.proxies)
```
bonzotronix: Maybe error out (here or elsewhere) if len(proxies) == 0.

bonzotronix:
FTP.  Looks great overall.

brkt_cli/user_data.py
bburtin: Copy/paste cruft.  I'll remove it.

test.py
bburtin: Done.

brkt_cli/__init__.py
bburtin: I don't think this would happen, due to the `if values.proxies` above.

PR #167, opened 2016-03-18
Compress userdata for encryptor instance

bonzotronix:
The encryptor AMI only looks for multi-part MIME userdata in
the case that it finds the userdata to be compressed.  So we'll
compress it, which is probably a good thing to do anyway.


brkt_cli/encrypt_ami.py
```diff
@@ -544,19 +549,32 @@ def run_encryptor_instance(aws_svc, encryptor_image_id,
...
         udc.add_file(
             '/var/brkt/ami_config/proxy.yaml',
             proxy_config,
             BRKT_FILES_CONTENT_TYPE
         )
-        user_data_string += '\n' + udc.to_mime_text()
+
+    user_data_string = udc.to_mime_text()
+    try:
+        log.info('Compressing userdata...')
```
bburtin: Seems gratuitous to log here if it's not a long-running operation.

brkt_cli/encrypt_ami.py
```diff
@@ -544,19 +549,32 @@ def run_encryptor_instance(aws_svc, encryptor_image_id,
...
+        out = StringIO()
```
bburtin: Is StringIO the right thing to use for binary data?

bburtin: Never mind.  Apparently this is the sanctioned way to do it.  Weird, I expected a byte buffer.

brkt_cli/encrypt_ami.py
bonzotronix: OK, will remove.

bonzotronix:
Replaced by squashed commits in PR#168

PR #168, opened 2016-03-18
Compress userdata for encryptor instance

bonzotronix:
[Squashed version of PR#167]

The encryptor AMI only looks for multi-part MIME userdata in the case that it finds the userdata to be compressed.  So we'll compress it, which is probably a good thing to do anyway.
Also, make the test_brkt_env_encrypt() test expect the userdata to be in compressed MIME format, rather than plain JSON.


bburtin:
Ship it!

PR #170, opened 2016-03-22
Update README.md

krishnanrs:
Add the missing CLI arguments to the README file (HVM, tags and Proxy).
Should I also add a note about how to create HVM AMIs or should I just leave it until HVM goes GA and we have formal documentation?


dramdass:
lgtm

PR #171, opened 2016-03-22
Use NO_PROXY talking to private encryptor ip

dramdass:
In some customer envrionments, customers use http_proxy
environment variables in the shell they are encrypting images
and it is needed for the cli when validating version. This
change will ensure we don't try to communicate to encryptor
instances with private ips through the proxy by appending the
encryptor private ip to the NO_PROXY environment variable.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -798,7 +798,7 @@ def snapshot_encrypted_instance(aws_svc, enc_svc_cls, encryptor_instance,
         host_ips.append(encryptor_instance.private_ip_address)
         log.info('Adding %s to NO_PROXY environment variable' %
                  encryptor_instance.private_ip_address)
-        if os.environ.get('NO_PROXY', None):
+        if os.environ.get('NO_PROXY'):
             os.environ['NO_PROXY'] += ",%s" % (
                 encryptor_instance.private_ip_address, )
         else:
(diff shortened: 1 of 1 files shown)


bobveznat:
Gotta love me some proxy! This commit works for me but I'll let Boris take a look at it before merging.

brkt_cli/encrypt_ami.py
```diff
@@ -796,6 +796,13 @@ def snapshot_encrypted_instance(aws_svc, enc_svc_cls, encryptor_instance,
         host_ips.append(encryptor_instance.ip_address)
     if encryptor_instance.private_ip_address:
         host_ips.append(encryptor_instance.private_ip_address)
+        log.info('Adding %s to NO_PROXY environment variable' %
+                 encryptor_instance.private_ip_address)
+        if os.environ.get('NO_PROXY', None):
```
bburtin: Nit: I'm pretty sure that `, None` is not necessary.

dramdass: Fixed

brkt_cli/encrypt_ami.py
```diff
@@ -796,6 +796,13 @@ def snapshot_encrypted_instance(aws_svc, enc_svc_cls, encryptor_instance,
...
+            os.environ['NO_PROXY'] = encryptor_instance.private_ip_address
+        else:
+            os.environ['NO_PROXY'] += ",%s" % (
```
bburtin: The syntax is kind of wonky here.  Would `+= ',' + encryptor_instance.private_ip_address` read better?  Also, you don't have to use a tuple when formatting a string with one variable.

dramdass: Fixed

dramdass:
Addressed code review issues

bburtin:
Ship it!

PR #172, opened 2016-03-22
YETI-715: Updating long named encrypted AMIs fails

krishnanrs:
Updating long named encrypted AMIs fails with an invalid AMI name error.
Added validations to truncate the name if it exceeds the max allowed length (128)


brkt_cli/__init__.py
```diff
@@ -456,6 +457,9 @@ def command_update_encrypted_ami(values, log):
         if m:
             encrypted_ami_name = m.group(1) + ' (encrypted %s)' % (nonce,)
         else:
+            suffix = ' (encrypted %s)' % (nonce,)
+            truncated_length = AMI_NAME_MAX_LENGTH - len(suffix)
+            name = name[:truncated_length]
```
bburtin: We're currently doing this in the encrypt-ami path.  It would be better if both code paths handled this case in a uniform way.

The encrypt path code is in `encrypt_ami.encrypt()`.  This is probably the wrong place.  I'm thinking we should do the validation in `__init__.py`, then pass the final name to `encrypt()` or `update_ami()`.  Feel free to make this change, or assign the bug to me.

bburtin:
I have a fix for this bug in another PR.

PR #176, opened 2016-03-25
Add --ntp-servers to brkt-cli.

apoorva-jindal:
--ntp-servers allow customers to configure locally hosted ntp-servers
for time synchronization. This commit makes this option visible.


bburtin:
Looks good, thanks.

PR #178, opened 2016-03-25
Adds pyasn1 0.1.9 to setup.py

robertabbott:
Oauth2client requires 0.1.7 and something
else requires 0.1.8. If pip resolves Oauth2client
first the brkt cli installs incorrectly. This
ensures we have a sufficiently up to date version
of pyasn1


bburtin:
Ship it.

PR #179, opened 2016-03-28
Load proxy.yaml from a file

bburtin:
Add a --proxy-config-file option to the encrypt-ami subcommand.  This
allows the user to pass in the entire proxy.yaml file, when specifying
proxies as host:port is not enough.


dramdass:
Looks good to me!

PR #180, opened 2016-03-30
Encrypted guest root now resizes itself to fit guest

robertabbott:
The disk to which the encrypted guest is dd'd is now
2x + 1 the size of the guest root instead of always
defaulting to 25GB

Testing: encrypted ubuntu-14.04, centos6 and rhel6
images and verified the guest booted

### DIFF
--- a/brkt_cli/encrypt_gce_image.py
+++ b/brkt_cli/encrypt_gce_image.py
@@ -121,6 +121,10 @@ class GCEService:
                 return
             time.sleep(10)
 
+    def get_disk_size(self, zone, diskName):
+        disk_info = self.compute.disks().get(zone=zone, project=self.project, disk=diskName).execute()
+        return int(disk_info['sizeGb'])
+
     def wait_for_detach(self, zone, diskName):
         while True:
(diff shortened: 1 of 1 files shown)


brkt_cli/encrypt_gce_image.py
```diff
@@ -397,8 +405,9 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
                                 encryptor,
                                 encrypted_image_disk])
 
-        log.info("Image %s successfully created!", encrypted_image_name)
-        return encrypted_image_name
+        if success:
+            log.info("Image %s successfully created!", encrypted_image_name)
```
bburtin: Can't you just do this at the line that says `success = True`?  Then you wouldn't need the variable at all.

robertabbott: that makes more sense.

robertabbott: fixed

bburtin:
Looks good.

PR #181, opened 2016-03-30
Launch-gce-image resizes guest to snapshot size

robertabbott:
Guest root disk now matches encrypted guest
snapshot size instead of defaulting to 25GB

Also adds option for specifying instance type
in the launch-gce-image command

Testing: launched a range of instance types
from 21GB snapshot.


bburtin:
Looks good.

PR #182, opened 2016-04-01
Write user data to temp file

bburtin:
When verbose logging is enabled, AWSService.run_instance() now writes
user data to a temporary file.  This gives us more information when
debugging issues with launching the encryptor instance.


bonzotronix:
LGTM

PR #186, opened 2016-04-06
Add proxy support to update-encrypted-ami

bburtin:
Add --proxy and --proxy-config-file options to the update-encrypted-ami
subcommand.

Add a _get_proxy_config() utility function for loading proxy.yaml
content.  This function is called during encryption and update.

Add a combine_user_data() function, which generates the gzipped MIME
blob that is passed to the metavisor.  This allows us to write dedicated
unit tests for this logic.

Minor terminology changes:
* user data: the actual user data bytes that are sent to the instance
* brkt env: host/port of the api service and HSM proxy
* brkt config: the dictionary that stores brkt env, ntp servers, etc.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -499,15 +499,44 @@ def create_encryptor_security_group(aws_svc, vpc_id=None):
     return sg
 
 
-def add_brkt_env_to_user_data(brkt_env, user_data):
+def add_brkt_env_to_brkt_config(brkt_env, brkt_config):
     if brkt_env:
-        if 'brkt' not in user_data:
-            user_data['brkt'] = {}
+        if 'brkt' not in brkt_config:
+            brkt_config['brkt'] = {}
         api_host_port = '%s:%d' % (brkt_env.api_host, brkt_env.api_port)
         hsmproxy_host_port = '%s:%d' % (
             brkt_env.hsmproxy_host, brkt_env.hsmproxy_port)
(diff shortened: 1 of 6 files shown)


brkt_cli/encrypt_ami.py
```diff
@@ -499,27 +499,56 @@ def create_encryptor_security_group(aws_svc, vpc_id=None):
...
         api_host_port = '%s:%d' % (brkt_env.api_host, brkt_env.api_port)
         hsmproxy_host_port = '%s:%d' % (
             brkt_env.hsmproxy_host, brkt_env.hsmproxy_port)
-        user_data['brkt']['api_host'] = api_host_port
-        user_data['brkt']['hsmproxy_host'] = hsmproxy_host_port
+        brkt_config['brkt']['api_host'] = api_host_port
+        brkt_config['brkt']['hsmproxy_host'] = hsmproxy_host_port
+
+
+def combine_user_data(brkt_config=None, proxy_config=None):
```
bonzotronix: I think it's likely that we'll soon want to add additional files to the userdata, such as the new-ish "logging.yaml" file described here:
https://docs.google.com/document/d/1AqwXJR2Q1PcQUuHuuQyor4g3ov3P8C8tZllD5F_d938/edit#heading=h.zgle73fvea4t

But I'm OK with generalizing combine_user_data at that time.

bonzotronix:
Changes LGTM.  I did not try running brkt-cli with these changes.

PR #187, opened 2016-04-07
Add NO_PROXY when we are updating an AMI

dramdass:
This is similar to what we do when encrypting an AMI. If the
updater has a private IP we don't talk to it using any configured
http proxy.


bburtin:
Thanks Dennis.  This looks good.

PR #189, opened 2016-04-07
Handle empty response when validating the encryptor image

bburtin:
The AWS API can return an empty response when the account doesn't have
access to the given image.  Handle this case when validating the
encryptor image.

Update _validate_encryptor_ami(), so that it raises ValidationError,
like the other validation functions.


bburtin:
Oops, there are other image validation functions that need the same fix.  Closing this PR.

PR #190, opened 2016-04-07
Handle empty response when validating the encryptor image

bburtin:
The AWS API can return an empty response when the account doesn't have
access to the image.  Handle this case when validating the an AMI.

Consolidate the image validation logic, so that we validate all images
in a consistent manner.

Update _validate_encryptor_ami(), so that it raises ValidationError,
like the other validation functions.


krishnanrs:
LGTM

mpage:
LGTM

PR #191, opened 2016-04-08
Fix image name option for encrypt-gce-image

robertabbott:
When no --encrypted_image_name option is passed
the encrypt command now generates a name for the
encrypted image rather than failing silently. It
uses the same log that update-gce-image uses to
generate a same (but the session id will change
so this won't cause collisions).

Removes the validate option from encrypt-gce-image
because it wasn't doing anything.

Testing: encrypted an image without the
encrypted_image_name option and then ran
update-gce-image on that encrypted image


brkt_cli/__init__.py
```diff
@@ -218,32 +218,36 @@ def command_launch_gce_image(values, log):
     return 0
 
 
+def get_image_name(encrypted_image_name, name):
```
bburtin: Can we move this into the GCE codebase?  Or do you see this as something that applies to all CSPs?

bburtin:
LGTM

PR #192, opened 2016-04-11
Handle long image name when updating

bburtin:
When updating an encrypted AMI, handle the possibility that the new name
may exceed 128 characters.  Refactor the image name generation logic
into a separate function, to make it more testable.  Move
append_suffix() into the util module, since it's now called from
multiple places.


bburtin:
Will open another PR to resolve the merge conflict.

PR #194, opened 2016-04-11
Adds checks to gce commands

robertabbott:
Now the cli ensures the encryptor image
exists and no image exists with the same
name as the output image for gce.

### DIFF
--- a/brkt_cli/gce_service.py
+++ b/brkt_cli/gce_service.py
@@ -73,6 +73,13 @@ class GCEService:
         return self.compute.images().get(project=self.project,
                image=image).execute()
 
+    def image_exists(self, image):
+        try:
+            self.get_image(image)
+        except:
+            return False
+        return True
+
     def delete_instance(self, zone, instance):
         return self.compute.instances().delete(project=self.project,
                zone=zone, instance=instance).execute()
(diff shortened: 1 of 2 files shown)


brkt_cli/gce_service.py
```diff
@@ -73,6 +73,13 @@ def get_image(self, image):
         return self.compute.images().get(project=self.project,
                image=image).execute()
 
+    def image_exists(self, image):
```
bburtin: It's not clear from the variable name whether image is a name, an id, or an object.  Not sure whether the problem is that the code is ambiguous, or that I'm not familiar with the GCE APIs.  Is it worth adding a docstring?

robertabbott: Image is a name/id. I named the variable that way because it matches the variable name in the api request. Also the google api never returns python objects like an 'image' object. The api returns json with the requested information.

robertabbott: but I could throw a comment in if that would be helpful

bburtin: That's fine, you can leave it if you think it's clear enough to someone who works with GCE on a regular basis.

bburtin:
Looks good.

PR #196, opened 2016-04-12
Move generic encryption methods to encryptor_service.py

robertabbott:
Some code shared by gce and aws (and likely other csps)
was in encrypt_ami.py. It makes more sense for this code
to be in the generic encryption service file.

Testing: encrypted and updated gce images

### DIFF
--- a/brkt_cli/encryptor_service.py
+++ b/brkt_cli/encryptor_service.py
@@ -15,10 +15,19 @@
 import abc
 import json
 import logging
+import time
 import urllib2
+
+
+from brkt_cli.util import (
+        BracketError,
+        Deadline,
+        sleep
+)
 
 ENCRYPT_INITIALIZING = 'initial'
 ENCRYPT_DOWNLOADING = 'downloading'
+ENCRYPTION_PROGRESS_TIMEOUT = 10 * 60  # 10 minutes
 ENCRYPT_RUNNING = 'encrypting'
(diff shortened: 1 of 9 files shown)


brkt_cli/encryptor_service.py
```diff
@@ -109,3 +166,98 @@ def get_status(self, timeout_secs=2):
             return info
         else:
             raise EncryptorConnectionError(self.port, exceptions_by_host)
+
+
+def sleep(seconds):
```
bburtin: This doesn't look like it's specific to the encryptor service.  Would it be better to put it into util?

brkt_cli/encryptor_service.py
```diff
@@ -27,10 +35,59 @@
...
+    def __repr__(self):
+        return '<BracketEnvironment api=%s:%d, hsmproxy=%s:%d>' % (
+            self.api_host,
+            self.api_port,
+            self.hsmproxy_host,
+            self.hsmproxy_port
+        )
+
+
+def get_default_tags(session_id, encryptor_ami):
```
bburtin: I don't think this is encryptor-specific either, since we set tags on instances, volumes and snapshots.  That said, I'm not immediately sure where to put it.

brkt_cli/encryptor_service.py
robertabbott: done

bburtin:
Looks good, thanks!

PR #198, opened 2016-04-12
Clean up the root snapshot if an error occurs

bburtin:
If an error occurs while waiting for the root snapshot, make sure that
the snapshot is cleaned up.


bburtin:
Closing to add a unit test.

PR #199, opened 2016-04-12
Clean up the root snapshot if an error occurs

bburtin:
If an error occurs while waiting for the root snapshot, make sure that
the snapshot is cleaned up.

### DIFF
--- a/brkt_cli/encrypt_ami.py
+++ b/brkt_cli/encrypt_ami.py
@@ -636,20 +636,24 @@ def _snapshot_root_volume(aws_svc, instance, image_id):
         'Creating snapshot %s of root volume for instance %s',
         snapshot.id, instance.id
     )
-    wait_for_snapshots(aws_svc, snapshot.id)
 
-    # Now try to detach the root volume.
-    log.info('Detaching root volume %s from %s' %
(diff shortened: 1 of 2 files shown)


brkt_cli/encrypt_ami.py
```diff
@@ -636,20 +636,24 @@ def _snapshot_root_volume(aws_svc, instance, image_id):
...
+        log.info('Detaching root volume %s from %s',
+                 root_vol.volume_id, instance.id)
+        aws_svc.detach_volume(
+            root_vol.volume_id,
+            instance_id=instance.id,
+            force=True
+        )
+        # And now delete it
+        log.info('Deleting root volume %s', root_vol.volume_id)
+        aws_svc.delete_volume(root_vol.volume_id)
```
dramdass: not related to this PR but I've seen eventual consistency issues here when we delete the volume and the detach hasn't finished yet. We probably need a retry here

dramdass:
Looks good to me

brkt_cli/encrypt_ami.py
bburtin: Yeah, I noticed that too.  Let's treat this as a separate issue.

PR #200, opened 2016-04-13
Adds gce tests

robertabbott:
Adds DummyGCEService and tests
gce image encryption and update

Testing: ran the tests and then
encrypted a gce image

### DIFF
--- a/brkt_cli/encrypt_gce_image.py
+++ b/brkt_cli/encrypt_gce_image.py
@@ -7,6 +7,7 @@ from brkt_cli.util import (
     Deadline,
 )
 
+from brkt_cli.validation import ValidationError
 from gce_service import gce_metadata_from_userdata
 from encryptor_service import wait_for_encryption
 from encryptor_service import wait_for_encryptor_up
@@ -14,11 +15,23 @@ from encryptor_service import wait_for_encryptor_up
 
 log = logging.getLogger(__name__)
 
+def validate_images(gce_svc, encrypted_image_name,  encryptor, guest_image):
+    # check that image to be updated exists
+    if not gce_svc.image_exists(guest_image):
+        raise ValidationError('Image %s does not exist. Cannot update.' % guest_image)
+
+    # check that encryptor exists
+    if not gce_svc.image_exists(encryptor):
+        raise ValidationError('Encryptor image %s does not exist. Encryption failed.' % encryptor)
+
+    # check that there is no existing image named encrypted_image_name
+    if gce_svc.image_exists(encrypted_image_name):
+        raise ValidationError('An image already exists with name %s. Encryption Failed.' % encrypted_image_name)
+
 
 def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
             encrypted_image_name, zone, brkt_env):
     brkt_data = {}
-    deleted = None
     try:
         add_brkt_env_to_user_data(brkt_env, brkt_data)
         instance_name = 'brkt-guest-' + gce_svc.get_session_id()
@@ -37,9 +50,14 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
         # of the unencrypted guest root
         log.info('Creating disk for encrypted image')
         gce_svc.create_disk(zone, encrypted_image_disk, guest_size * 2 + 1)
+    except Exception as e:
+        gce_svc.cleanup(zone)
+        log.info('Encryption setup failed')
+        raise e
 
-        # run encryptor instance with avatar_creator as root,
-        # customer image and blank disk
+    # run encryptor instance with avatar_creator as root,
+    # customer image and blank disk
+    try:
         log.info('Launching encryptor instance')
         gce_svc.run_instance(zone,
                              encryptor,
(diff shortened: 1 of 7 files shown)


brkt_cli/encrypt_gce_image.py
```diff
@@ -37,9 +36,13 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
         # of the unencrypted guest root
         log.info('Creating disk for encrypted image')
         gce_svc.create_disk(zone, encrypted_image_disk, guest_size * 2 + 1)
+    except Exception as e:
+        gce_svc.cleanup(zone)
+        raise Exception('Encryption setup failed %s' % e)
```
bburtin: I think this would cause you to lose the original traceback.

brkt_cli/gce_service.py
```diff
@@ -31,17 +32,157 @@ class InstanceError(BracketError):
...
+    @abc.abstractmethod
+    def delete_instance(self, zone, instance):
+        pass
+
+    @abc.abstractmethod
+    def delete_disk(self, zone, disk):
+        pass
+
+    @abc.abstractmethod
+    def wait_instance(self, name, zone):
```
bburtin: You may want to make this a utility function that accepts a GCEService object.  This keeps the service implementations shorter, and allows your unit tests to test more of the production code.  Same goes for the other wait_*() methods.

brkt_cli/gce_service.py
```diff
@@ -65,8 +206,11 @@ def wait_snapshot(self, snapshot):
             time.sleep(5)
 
     def get_image(self, image):
-        return self.compute.images().get(project=self.project,
-               image=image).execute()
+        try:
+            return self.compute.images().get(project=self.project,
+                   image=image).execute()
+        except Exception as e:
```
bburtin: Is this necessary?

brkt_cli/gce_service.py
```diff
@@ -31,17 +32,157 @@ class InstanceError(BracketError):
...
         self.credentials = GoogleCredentials.get_application_default()
         self.compute = discovery.build('compute', 'v1',
                 credentials=self.credentials)
         self.storage = discovery.build('storage', 'v1',
                 credentials=self.credentials)
         self.gce_res_uri = "https://www.googleapis.com/compute/v1/"
+        self.disks = []
+        self.instances = []
+
+    def cleanup(self, zone):
```
bburtin: This also looks like something that should be a function and not a class method.

brkt_cli/gce_service.py
robertabbott: removed

test_gce.py
```diff
@@ -0,0 +1,353 @@
...
+    def get_disk(self, zone, disk_name):
+        source_disk = "projects/%s/zones/%s/disks/%s" % (self.project, zone, disk_name)
+        return {
+            'boot': False,
+            'autoDelete': False,
+            'source': self.gce_res_uri + source_disk,
+        }
+
+
+def _build_gce_service():
```
bburtin: Is this necessary, if it's just a one-liner?

test_gce.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def test_long_image_name(self):
+        image_name = 'test-image-with-long-name-encrypted-so-we-hit-63-char-limit'
+        n1 = gce_service.get_image_name(None, image_name)
+        n2 = gce_service.get_image_name(None, image_name)
+        self.assertNotEqual(n1, n2)
+        truncated = True
+        if '63-char-limit' in n1 or '63-char-limit' in n2:
+            truncated = False
+        self.assertEqual(truncated, True)
```
bburtin: More of a nit, but would this read better?
`
self.assertTrue(
    '63-char-limit' in n1 or '63-char-limit' in n2)
`

test_gce.py
```diff
@@ -0,0 +1,353 @@
...
+                gce_svc=gce_svc,
+                enc_svc_cls=test.DummyEncryptorService,
+                image_id='test-ubuntu',
+                encryptor_image=BOGUS_IMAGE,
+                encrypted_image_name='thor',
+                zone='us-central1-a',
+                brkt_env=None,
+            )
+            self.fail('Encryption should have failed')
+        except:
```
bburtin: Check out `with self.assertRaises(ExceptionType):`

brkt_cli/gce_service.py
robertabbott: I'll address this in another change.

brkt_cli/encrypt_gce_image.py
```diff
@@ -49,10 +67,19 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
                              metadata=gce_metadata_from_userdata(brkt_data))
 
         enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
+
         wait_for_encryptor_up(enc_svc, Deadline(600))
         wait_for_encryption(enc_svc)
         gce_svc.delete_instance(zone, encryptor)
-        deleted = True
+    except Exception as e:
+        log.info(e)
```
bburtin: I don't think you need to log here, since the exception is being thrown outside the function.  You are likely to double-log the error.  I would do
```
except:
    gce_svc.cleanup(zone)
    raise
```

But you know the code better.  Your call.

brkt_cli/gce_service.py
```diff
@@ -31,17 +32,157 @@ class InstanceError(BracketError):
...
+    def __init__(self, project, session_id, logger):
+        super(GCEService, self).__init__(project, session_id, logger)
         self.credentials = GoogleCredentials.get_application_default()
         self.compute = discovery.build('compute', 'v1',
                 credentials=self.credentials)
         self.storage = discovery.build('storage', 'v1',
                 credentials=self.credentials)
-        self.gce_res_uri = "https://www.googleapis.com/compute/v1/"
+
+    def cleanup(self, zone):
```
bburtin: Sorry, I missed this last time.  You should probably put a try/except around the cleanup logic.  If you don't, you'd lose context if

1. Exception is raised in application logic
1. The finally block calls clean_up()
1. clean_up raises another exception

For this reason, I always put try/except/log in cleanup code.

bburtin:
Looks good.  Thanks!

PR #201, opened 2016-04-19
Support specifying the guest instance type

bburtin:
Add a --guest-instance-type option to the encrypt-ami and update-
encrypted-ami subcommands.  This allows the user to specify the guest
instance type when the default of m3.medium is incompatible with the
guest AMI.

Clean up the original test_instance_type() test, so that it switches on
the AMI ID instead of making assumptions about the order in which
instances are launched.

### DIFF
--- a/brkt_cli/aws_service.py
+++ b/brkt_cli/aws_service.py
@@ -277,12 +277,11 @@ class AWSService(BaseAWSService):
                      instance_profile_name=None):
         if security_group_ids is None:
             security_group_ids = []
-        log.debug('Starting a new instance based on %s', image_id)
-        if security_group_ids:
-            log.debug(
-                'Using security groups %s', ', '.join(security_group_ids))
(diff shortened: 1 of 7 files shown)


brkt_cli/aws_service.py
```diff
@@ -277,12 +277,11 @@ def run_instance(self,
...
         if security_group_ids is None:
             security_group_ids = []
-        log.debug('Starting a new instance based on %s', image_id)
-        if security_group_ids:
-            log.debug(
-                'Using security groups %s', ', '.join(security_group_ids))
-        if subnet_id:
-            log.debug('Running instance in %s', subnet_id)
+        log.debug(
+            'run_instance: %s, security groups=%s, subnet=%s, '
```
krishnanrs: What happens if security group and subnet are not specified in the command line options?

krishnanrs:
LGTM

PR #202, opened 2016-04-20
Check and ensure that GCE image name is in lower case

krishnanrs:


brkt_cli/__init__.py
```diff
@@ -222,6 +222,9 @@ def command_update_encrypted_gce_image(values, log):
     session_id = util.make_nonce()
     gce_svc = gce_service.GCEService(values.project, session_id, log)
     encrypted_image_name = gce_service.get_image_name(values.encrypted_image_name, values.image)
+    
+    if not encrypted_image_name.islower():
+        raise ValidationError('GCE image name must only be in lower case')
```
bburtin: 'GCE image name must be in lower case' would read better.

robertabbott:
lgtm

PR #204, opened 2016-04-25
Brkt_env defaults to prod for gce commands

robertabbott:
If no brkt_env option is provided to encrypt
or update gce image commands it defaults to
the prod endpoints.

Testing: ran encrypt_gce_image and update_gce_image
without --brkt-env and verified the prod endpoints
were passed through userdata.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -50,6 +50,7 @@ from encryptor_service import BracketEnvironment
 from update_ami import update_ami
 
 VERSION = '0.9.15pre1'
+BRKT_ENV_PROD = 'yetiapi.mgmt.brkt.com:443,hsmproxy.mgmt.brkt.com:443'
 
 log = None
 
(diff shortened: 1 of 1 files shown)


brkt_cli/__init__.py
```diff
@@ -50,6 +50,7 @@
 from update_ami import update_ami
 
 VERSION = '0.9.15pre1'
+BRKT_ENV_PROD = 'yetiapi.mgmt.brkt.com:443,hsmproxy.mgmt.brkt.com:443'
```
bburtin: Would it be more straightforward to make this a BracketEnvironment object instead of a string?  I'm fine either way.

krishnanrs:
LGTM

brkt_cli/__init__.py
robertabbott: This seemed like the simplest, though maybe not the clearest, way to get the BracketEnvironment object and it is similar to how we get the BE object when the --brkt-env option is used. So I'm just going to leave as is

bburtin: Ok, looks good.

PR #205, opened 2016-04-25
Allow specifying a GCE image project

krishnanrs:
YETI-819: Added an optional "image-project" attribute to the brkt-cli for GCE image encryption process. This is required to allow users to be able to directly encrypt a "public" GCE image without have to have a local copy of that image in their account (like we do with AWS). This was tested with the following conditions:
1. Encrypting a public GCE image without specifying an image project (failed)
2. Encrypting a public GCE image from a different image project to the one specified (failed)
3. Encrypting a public GCE image in the same image project specified (successful)
4. Encrypting a private GCE image without specifying an image project (successful)

### DIFF
--- a/brkt_cli/encrypt_gce_image_args.py
+++ b/brkt_cli/encrypt_gce_image_args.py
@@ -37,6 +37,12 @@ def setup_encrypt_gce_image_args(parser):
         required=True
     )
     parser.add_argument(
+        '--image-project',
+        help='GCE project which owns the image',
+        dest='image_project',
+        required=False
+    )
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


brkt_cli/encrypt_gce_image_args.py
```diff
@@ -37,6 +37,12 @@ def setup_encrypt_gce_image_args(parser):
         required=True
     )
     parser.add_argument(
+        '--image-project',
```
bburtin: Would it be more intuitive to specify a metavar here (e.g. NAME, ID, NAME_OR_ID)?

PR #206, opened 2016-04-25
Gce image project

krishnanrs:
Added metavar for image-project attribute


brkt_cli/encrypt_gce_image_args.py
```diff
@@ -37,6 +37,13 @@ def setup_encrypt_gce_image_args(parser):
         required=True
     )
     parser.add_argument(
+        '--image-project',
+        metavar='NAME',
+        help='GCE project name which owns the image (for e.g. centos-cloud)',
```
bburtin: I believe that "e.g." means "for example."  So "for e.g." would mean "for for example."  :-)

PR #207, opened 2016-04-25
Gce image project

krishnanrs:


bburtin:
Looks good to me (except the unit test failures).  I'll let GCE master Bobby make the final call.

PR #211, opened 2016-04-25
Gce image project

krishnanrs:
Updated failing tests and added a new test for invalid project name


robertabbott:
LGTM. Some of these commits that are logically related (and have the same commit message) could probably be squashed, if we care.

PR #212, opened 2016-04-26
Yeti 826 gce launch

krishnanrs:
Need to pass the image_project argument in the correct order.


robertabbott:
LGTM. sorry i didn't catch that in the last one

PR #213, opened 2016-04-26
Provide a means to add a startup-script for GCE instances

krishnanrs:
Added a startup-script option to the launch-gce-image command
to provide an ability to specify a startup script during instance
launch. This can be used for testing cloud-init and similar
testing (see YETI-822).

Jira: YETI-822

### DIFF
--- a/brkt_cli/launch_gce_image_args.py
+++ b/brkt_cli/launch_gce_image_args.py
@@ -42,6 +42,13 @@ def setup_launch_gce_image_args(parser):
         dest='project',
         required=True
     )
+    parser.add_argument(
+        '--startup-script',
+        help='GCE instance startup script',
+        dest='startup_script',
+        metavar='SCRIPT',
+        required=False
+    )
 
     # Optional yeti endpoints. Hidden because it's only used for development.
     # If you're using this option, it should be passed as a comma separated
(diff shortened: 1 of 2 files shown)


robertabbott:
I think this is fine for testing cloud-init etc. However I don't think we should do too much proxying of create instance options.

Most people don't launch their instances manually, they have config somewhere and something that executes that config, in which case they can add the encrypted guest disk themselves. Until people other than us specifically ask for more functionality from this command I don't think we should spend time creating parity between launch-gce-image and `gcloud compute instances create ...`

krishnanrs:
Part of the problem is that the automation framework uses the launch-gce-image script. Once we move away from that, additional work on this front won't be required.

bburtin:
If this option is only for testing, it should be hidden from the usage output.  See the `--encryptor-ami` argument to the `encrypt-ami` subcommand.

brkt_cli/launch_gce_image_args.py
```diff
@@ -43,6 +43,28 @@ def setup_launch_gce_image_args(parser):
...
+        help=argsparse.SUPPRESS,
+        dest='startup_script',
+        metavar='SCRIPT',
+        required=False
+    )
+
+    # Optional startup script. Hidden because it is only used for development
+    # and testing. It should be passed as a string containing a multi-line
+    # script (bash, python etc.)
+    parser.add_argument(
```
robertabbott: duplicate code

robertabbott:
LGTM

PR #217, opened 2016-04-26
Adds retry to gce api calls that need it

robertabbott:
Adds a retry method to utils. Also adds
execute_gce_api_call which wraps a googleapiclient
request object to make it retry-able.

API are wrapped with a retry where appropriate.
Example of an appropriate place: wait_instance
where we are waiting for a resource we created

Example of a bad place to put it: create_disk
if that fails it almost definitely is a real
problem

It's also worth noting that if for some reason
someone wanted to retry create_disk they could
I only added retries to methods in the gce_service
that clearly need it.

Testing: encrypted and updated a gce image


bburtin:
Looks good.  I'm thinking of getting rid of the current retry logic in the AWS code and using this retry() function instead.  :-)

PR #218, opened 2016-04-27
Puts gce encrypt in a try finally block

robertabbott:
Ensures that even if things are interrupted
or exit in a weird place all the resources
created in gce will still be cleaned up

testing: encrypted and updated images
and also interrupted the encryption
processes in a few different places


brkt_cli/encrypt_gce_image.py
```diff
@@ -33,69 +33,70 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
...
-        gce_svc.wait_snapshot(encrypted_image_name)
-        log.info("Image %s successfully created!", encrypted_image_name)
-    except Exception as e:
-        log.info('Image creation failed')
-        raise e
-
-    log.info("Cleaning up")
-    gce_svc.cleanup(zone)
-    return encrypted_image_name
+        return encrypted_image_name
```
krishnanrs: shouldn't this be in the main try block instead of finally? otherwise it gives the impression that the encrypted image has been created.

PR #219, opened 2016-04-27
YETI-836: Better GCE image name validation

krishnanrs:
Better GCE image name validation to match all their nits.


bburtin:
How about moving the validation logic into a validate_image_name() function and adding a test?

PR #220, opened 2016-04-27
YETI-836: Better GCE image name validation

krishnanrs:
Better GCE image name validation along with added unit tests


bburtin:
Looks good.  Thanks for adding the test!

PR #221, opened 2016-04-28
Add proxy support to update-encrypted-ami

bburtin:
Add --proxy and --proxy-config-file options to the update-encrypted-ami
subcommand.

Add a _get_proxy_config() utility function for loading proxy.yaml
content.  This function is called during encryption and update.

Add a combine_user_data() function, which generates the gzipped MIME
blob that is passed to the metavisor.  This allows us to write dedicated
unit tests for this logic.

Minor terminology changes:
* user data: the actual user data bytes that are sent to the instance
* brkt env: host/port of the api service and HSM proxy
* brkt config: the dictionary that stores brkt env, ntp servers, etc.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -339,6 +339,32 @@ def _parse_proxies(*proxy_host_ports):
     return proxies
 
 
+def _get_proxy_config(values):
+    """ Read proxy config specified by either the --proxy or
+    --proxy-config-file option.
+
+    @return: the contents of the proxy.yaml file, or None if not specified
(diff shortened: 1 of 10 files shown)


robertabbott:
looks good to me. Adam might want to look at though

bonzotronix:
Looks *even better* than last time.  We have to have the 4/26 AMIs pushed to prod before merging this, right?

LGTM

bburtin:
Yes.  Danny is going to test this branch with the 4/26 metavisor build.  I'll merge this branch into master once the 4/26 metavisor build is in production.

PR #222, opened 2016-04-29
GCE token auth support

robertabbott:
Testing: encrypted an image
then updated it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: python
 python: "2.7"
-install: pip install boto requests google-api-python-client PyYAML
+install: pip install boto requests google-api-python-client PyYAML oauthlib
 script: python -m unittest test test_gce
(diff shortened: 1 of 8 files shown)


brkt_cli/update_encrypted_gce_image_args.py
```diff
@@ -15,6 +15,20 @@ def setup_update_gce_image_args(parser):
         required=False
     )
     parser.add_argument(
+         '--brkt-email',
+         metavar='EMAIL',
+         dest='api_email',
+         help='Bracket user email',
+         required=True
```
krishnanrs: Do these arguments need to be mandatory here, especially if this was already passed when originally encrypting the image?

krishnanrs:
Except for that one comment and my reservations about using user credentials while authenticating, the changes looks good to me.

brkt_cli/update_encrypted_gce_image_args.py
robertabbott: fixed

robertabbott:
for the time being user credentials is what we have to use. This may change in the future

krishnanrs:
lgtm, except for the test failures now. grant explained that the user credentials are temporary until we have the tagging feature in place.

PR #225, opened 2016-05-03
Remove --updater-ami from the update-encrypted-ami example

bburtin:
Remove the --updater-ami option from the sample output of update-
encrypted-ami.  It's now a hidden option that's meant to be used in
development.


sondring:
LGTM!

PR #226, opened 2016-05-03
Delete guest disk upon instance termination

krishnanrs:
Configure the instance disk to be always deleted upon instance termination. If there is a better way to do this, then let me know.

### DIFF
--- a/brkt_cli/launch_gce_image.py
+++ b/brkt_cli/launch_gce_image.py
@@ -10,10 +10,12 @@ def launch(log, gce_svc, image_id, instance_name, zone, delete_boot, instance_ty
     gce_svc.disk_from_snapshot(zone, image_id, guest)
     gce_svc.wait_for_disk(zone, guest)
     log.info("Starting instance")
+    guest_disk = gce_svc.get_disk(zone, guest)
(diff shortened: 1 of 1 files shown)


brkt_cli/launch_gce_image.py
```diff
@@ -10,10 +10,12 @@ def launch(log, gce_svc, image_id, instance_name, zone, delete_boot, instance_ty
     gce_svc.disk_from_snapshot(zone, image_id, guest)
     gce_svc.wait_for_disk(zone, guest)
     log.info("Starting instance")
+    guest_disk = gce_svc.get_disk(zone, guest)
+    guest_disk['autoDelete'] = True
```
robertabbott: does autoDelete work for non-root disks? I haven't tried that

robertabbott:
I didn't want to make the decision for the guest to delete their disks. If this works and you think it makes things easier for the customer LGTM

krishnanrs:
This is mainly for automation to make sure that we cleanup all resources. I did confirm that autoDelete works for the guest volume as well. The one thing I can do is to configure it for autoDelete only when --delete-boot is set

PR #227, opened 2016-05-03
Add support for subcommand modules

bburtin:
Move AWS argument parsing into a new brkt_cli.aws module.  Load the
module by name and call its interface to initialize command line
arguments.  The next step will be to move all of the AWS-specific code
into brkt_cli.aws so that it isn't intermingled with other CLI and GCE
logic.

### DIFF
--- a/brkt_cli/module.py
+++ b/brkt_cli/module.py
@@ -1,0 +1,60 @@
+# Copyright 2015 Bracket Computing, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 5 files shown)


brkt_cli/module.py
```diff
@@ -0,0 +1,60 @@
...
+    @abc.abstractmethod
+    def get_exposed_subcommands(self):
+        """ Return the list of subcommands that are shown in brkt command
+        usage.
+        @return a list of strings
+        """
+        pass
+
+    @abc.abstractmethod
+    def get_loggers(self):
```
mpage: Do you need this? Shouldn't setting the global log level handle this?

bburtin: We're currently doing this to avoid logging spew from boto and other 3rd party libraries.  I filed a separate issue to clean this up.

PR #229, opened 2016-05-03
Increase retry time

robertabbott:
I found wait_for_detach was failing because
retry wasn't waiting long enough. This should
fix that.


krishnanrs:
LGTM

PR #230, opened 2016-05-04
Puts gce encrypt in a try finally block

robertabbott:
Ensures that even if things are interrupted
or exit in a weird place all the resources
created in gce will still be cleaned up

testing: encrypted and updated images
and also interrupted the encryption
processes in a few different places


krishnanrs:
LGTM

PR #239, opened 2016-05-04
Initialize logging in submodules

bburtin:
Add an init_logging() method to ModuleInterface, which allows submodules
to configure logging.  Move boto logging setup into the aws submodule.
Set the default log level on the root logger, per the usual convention,
and suppress boto logging explicitly.

Import boto.vpc in aws_service.  Somehow an unrelated code change caused
boto.vpc.regions() to start failing.

### DIFF
--- a/brkt_cli/module.py
+++ b/brkt_cli/module.py
@@ -36,10 +36,11 @@ class ModuleInterface(object):
         """
         pass
 
-    @abc.abstractmethod
-    def get_loggers(self):
-        """ Return a list of Logger objects.  brkt_cli sets the log level
-        to INFO or DEBUG, depending on whether the user specifies -v.
+    def init_logging(self, verbose):
+        """ Modules can optionally implement this callback to initialize
+        loggers.
+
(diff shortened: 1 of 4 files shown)


brkt_cli/module.py
```diff
@@ -36,10 +36,11 @@ def get_exposed_subcommands(self):
         """
         pass
 
-    @abc.abstractmethod
-    def get_loggers(self):
-        """ Return a list of Logger objects.  brkt_cli sets the log level
-        to INFO or DEBUG, depending on whether the user specifies -v.
+    def init_logging(self, verbose):
```
mpage: Does this need an `@abc.abstractmethod` decorator?

bburtin: I decided not to make it abstract because it's not a requirement.  But I can see how making it abstract would make the implementation clearer.  I can go either way.

mpage: Ok. It's fine the way it is.

Ship it! 👍

PR #240, opened 2016-05-04
Return exit status from run_subcommand()

bburtin:
Update AWSModuleInterface.run_subcommand() and the callsite to return
the exit status.  Failing to do this caused the function to return None,
which resulted in an exit status of 1.


krishnanrs:
LGTM

PR #244, opened 2016-05-06
Block project wide ssh keys when we launch guest for encryption

robertabbott:
This will prevent us from creating an undesirable authorized_keys
file on the guest root disk.


krishnanrs:
LGTM

PR #246, opened 2016-05-09
Remove dependencies from README

bburtin:
The list of dependencies has become stale.  Removing it, so that we
don't have to keep it current.  It's no longer necessary, now that we're
using pip/setuptools to automatically install dependencies.


sondring:
LGTM

PR #247, opened 2016-05-10
Cleanup GCE image

krishnanrs:
1. Added a keep-encryptor flag to keep the encryptor image, even
if encryption/update fails due to any reason. Default action would
be to delete the encryptor image upon completion or failure. This 
ensures that we do not leave any crud in the customer GCE account.
2. Increased the retries waiting for disk (this fixes YETI-824)
3. Haven't added unit tests (is it needed here?)
4. Ran unit tests locally and verified that it passed


brkt_cli/__init__.py
```diff
@@ -223,41 +231,49 @@ def command_encrypt_gce_image(values, log):
...
-    # create image from file in GCS bucket
-    else:
-        log.info('Retrieving encryptor image from GCS bucket')
-        encryptor = 'encryptor-%s' % gce_svc.get_session_id()
-        if values.image_file:
-            gce_svc.get_latest_encryptor_image(values.zone,
-                                               encryptor,
-                                               values.bucket,
-                                               image_file=values.image_file)
+    try:
```
robertabbott: It would be nice if we could move the image getting logic into its own method and then call that method from inside update_gce_image and encrypt (or somewhere else) and then we don't have to duplicate this code.

Also if get_encryptor_image() was called in update_gce_image and encrypt we wouldn't have the nested try-finally blocks.

PR #248, opened 2016-05-10
Cleanup GCE image

krishnanrs:
1. Added a keep-encryptor flag to keep the encryptor image, even
if encryption/update fails due to any reason. Default action would
be to delete the encryptor image upon completion or failure.
2. Increased the retries waiting for disk and attached a catch exception statement (this fixes YETI-824) 
3. Updated unit tests
4. Ran unit tests locally and verified that it passed


robertabbott:
lgtm

PR #251, opened 2016-05-11
Bobby/404

robertabbott:


brkt_cli/__init__.py
```diff
@@ -215,19 +208,14 @@ def command_encrypt_gce_image(values, log):
...
-        encryptor = 'encryptor-%s' % gce_svc.get_session_id()
 
 
     log.info('Starting encryptor session %s', gce_svc.get_session_id())
     encrypted_image_id = encrypt_gce_image.encrypt(
         gce_svc=gce_svc,
         enc_svc_cls=encryptor_service.EncryptorService,
         image_id=values.image,
-        encryptor_image=encryptor,
+        encryptor_image=values.encryptor_image,
```
krishnanrs: The encryptor_image is an optional parameter. If this is not passed, then the encrypt/update method will fail as it is expecting a non-null value.

robertabbott: The problem with that was that encryptor was always defined. The encrypt method couldn't tell if it should use an existing image or pull down a new one. If the --encryptor-image option was used encrypt would still try to pull an image and name it values.encryptor_image. The resulting image naming conflict crashed the encryption.

Now the check is in the encrypt method. If values.encryptor_image (the value passed to encrypt) is null then it will retrieve an image from GCS, otherwise it will try to use an existing image of that name.

krishnanrs:
lgtm

PR #252, opened 2016-05-11
Log header and payload JSON when -v is specified

bburtin:


bburtin:
Merge got messy.  Will open another PR.

PR #253, opened 2016-05-11
Rename "module interface" to "module connector"

bburtin:
Module interfaces are now module connectors.  This should alleviate the
terminology confusion about an interface in programming terms vs. an
interface to a module.


bburtin:
Still not quite right.  Need to think about terminology some more.

PR #256, opened 2016-05-11
Pass the keep_encryptor flag for image updates

krishnanrs:
Pass the keep_encryptor flag so that the encryptor image does not get cleaned up when specified via command line args.


robertabbott:
lgtm

PR #257, opened 2016-05-11
Refactor subcommand loading

bburtin:
Refactor the subcommand loading, and get rid of the "module" concept.
Each subcommand is now treated separately.  This makes the
implementation of each subcommand much simpler.


mpage:
Looks great! Ship it 👍

PR #258, opened 2016-05-12
Check subcommand result

bburtin:
Make sure that the run() method of the subcommand returns the result
status as an integer.  This should help us find problems with status
codes without having to run the integration tests.


krishnanrs:
lgtm

PR #260, opened 2016-05-12
Preserve volume type for any attached volumes

krishnanrs:
When updating an image, we ignore any attached volumes except for
the "known" one's (sda1, sda2, sda3, sda5 for PV and sda and sdf
for PV). For any attached volume, we should simply preserve the
original volume type configured with the AMI.

This was tested by:
1. Launching an encrypted AMI
2. Attaching a couple of EBS volumes at sdc and sdd
3. Creating a new image snapshot of the running Solo instance
4. Updating the AMI created in step 3 above
5. Verifying that the volume type in the updated AMI matched with the original AMI

Jira: YETI-942 #resolve


brkt_cli/aws/update_ami.py
```diff
@@ -199,11 +199,21 @@ def update_ami(aws_svc, encrypted_ami, updater_ami,
...
             # Use guest_instance as base instance for create_image
             boot_snap_name = NAME_METAVISOR_ROOT_SNAPSHOT
             root_device_name = guest_image.root_device_name
             guest_root = '/dev/sdf'
+            d_list.append(guest_root)
+
+        # Preserve volume type for any additional attached volumes
+        for d in guest_bdm.keys():
+            if d not in d_list:
+                log.info("Preserving volume type for disk %s" % d)
```
bburtin: This doesn't sound like something that users need to be aware of by default. Should it be debug?  Also, the convention is to pass logging arguments by value:
`log.info("Preserving volume type for disk %s", d)`

krishnanrs:
Will send out a new review addressing Boris' comments.

PR #261, opened 2016-05-12
Preserve volume type for any attached volumes

krishnanrs:
When updating an image, we ignore any attached volumes except for
the "known" one's (sda1, sda2, sda3, sda5 for PV and sda and sdf
for PV). For any attached volume, we should simply preserve the
original volume type configured with the AMI.

This was tested by:
1. Launching an encrypted AMI
2. Attaching a couple of EBS volumes at sdc and sdd
3. Creating a new image snapshot of the running Solo instance
4. Updating the AMI created in step 3 above
5. Verifying that the volume type in the updated AMI matched with the original AMI

Jira: YETI-942 #resolve


jimvoll:
Code looks good to me

PR #263, opened 2016-05-13
Add "kid" field to the generated JWT

bburtin:
Add a key id field ("kid") to the generated JWT.  To calculate the key
id, we:

1) Load the public key that's associated with the signing key.
2) Convert the public key to a JWK.
3) Calculate the thumbprint of the JWK, as specified in RFC 7638.


brkt_cli/jwt/__init__.py
```diff
@@ -142,7 +143,8 @@ def generate_jwt(signing_key, exp=None, nbf=None, cnc=None):
     payload_dict = {
         'jti': util.make_nonce(),
         'iss': 'brkt-cli-' + brkt_cli.VERSION,
-        'iat': int(time.time())
+        'iat': int(time.time()),
```
gcallaghan: Is this UTC?

brkt_cli/jwt/jwk.py
```diff
@@ -0,0 +1,55 @@
...
+        'crv': 'P-384',
+        'x': x,
+        'y': y
+    }
+
+
+def get_thumbprint(verifying_key):
+    """ Get the thumbprint of an ecdsa.VerifyingKey. """
+    jwk = ecdsa_to_jwk(verifying_key)
+    thumbprint_json = '{"crv":"%s","kty":"%s","x":"%s","y":"%s"}' % (
```
gcallaghan: ```'{"crv":"%(crv)s","kty":"%(kty)s","x":"%(x)s","y":"%(y)s"}' % jwk```

brkt_cli/jwt/jwk.py
```diff
@@ -0,0 +1,55 @@
...
+    return s.decode("ascii")
+
+
+def ecdsa_to_jwk(verifying_key):
+    """ Convert an ecdsa.VerifyingKey to the equivalent JWK. """
+    x = _long_to_base64(verifying_key.pubkey.point.x())
+    y = _long_to_base64(verifying_key.pubkey.point.y())
+    return {
+        'alg': 'ES384',
+        'use': 'sig',
```
gcallaghan: we can get rid of 'use'

brkt_cli/jwt/__init__.py
bburtin: It's a Unix timestamp.  The API spec says "Return the time in seconds since the epoch as a floating point number."

bburtin:
Opening another PR.

PR #264, opened 2016-05-13
Add "kid" field to the generated JWT

bburtin:
Add a key id field ("kid") to the generated JWT.  To calculate the key
id, we:

1) Load the public key that's associated with the signing key.
2) Convert the public key to a JWK.
3) Calculate the thumbprint of the JWK, as specified in RFC 7638.


gcallaghan:
lgtm

PR #265, opened 2016-05-16
Split gce into its own package

robertabbott:
Should insulate aws users from gce import
issues and make things more modular and nice

Testing: encrypted, updated and launched gce
images


bburtin:
Looks good.  Thanks!

PR #269, opened 2016-05-17
Add --claim option to generate-jwt

bburtin:
Add a --claim option, which allows the user to add JWT claims as
name/value pairs.  Refactor the name/value parsing code, so that the
same code is used for claims and tags.


mpage:
LGTM 👍

PR #270, opened 2016-05-17
Clean up subcommand handling

bburtin:
Require that subcommand lookup succeeds.  We now use the new subcommand
infrastructure for all subcommands.

Move the version check into a separate function, and call it before
running the subcommand.  A previous commit inadvertently switched the
order of these operations.

Testing:
* python -m unittest discover
* hack the hardcoded version and manually test the version check
* manually run encrypt-ami, update-encrypted-ami, and generate-jwt

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -18,7 +18,6 @@ import argparse
 import importlib
 import json
 import logging
-import os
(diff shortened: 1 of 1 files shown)


bonzotronix:
LGTM

PR #271, opened 2016-05-18
Check encryptor_image and image_bucket args

robertabbott:
Raise exception if both args are pass to
encrypt or update gce-image

Testing: checked that providing both options
causes exception and passing one does not

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -17,6 +17,7 @@ from brkt_cli.gce import (
     update_gce_image,
     update_encrypted_gce_image_args,
 )
+from brkt_cli.validation import ValidationError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


brkt_cli/gce/__init__.py
```diff
@@ -160,3 +162,7 @@ def command_encrypt_gce_image(values, log):
     return 0
 
 
+def check_args(values):
+    if values.encryptor_image:
+        if values.bucket != 'prod':
+            raise Exception("you provided both an encryptor image and an image bucket")
```
bburtin: Will this print a traceback to the console?  Should it be ValidationError instead?

bburtin: Also, not sure if this applies, but you could look into using a mutually exclusive option group.  See `encrypt_ami_args.py`.

bburtin:
Ship it!

PR #273, opened 2016-05-18
Improve region validation

bburtin:
When --encryptor-ami is not specified, make sure that the specified
region is one of

* us-east-1
* us-west-1
* us-west-2

When --encryptor-ami is specified, we check against all AWS regions, in
case this is a test build of metavisor.


dramdass:
LGTM

PR #275, opened 2016-05-19
Add a command for generating cloud-init user-data with a JWT

bburtin:
```
$ ./brkt generate-user-data -h
usage: brkt generate-user-data [-h] --jwt JWT [-v]

Generate cloud-init user-data that is passed to the AWS RunInstances
API.

optional arguments:
-h, --help     show this help message and exit
--jwt JWT      JSON Web Token that the encrypted instance will use to
                 authenticate with the Bracket service. Use the generate-jwt
                 subcommand to generate a JWT.
-v, --verbose  Print status information to the console
```


brkt_cli/generate_user_data/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+        udc = UserDataContainer()
+
+        jwt = brkt_cli.validate_jwt(values.jwt)
+        udc.add_file(
+            '/var/brkt/instance_config/token.jwt',
+            jwt,
+            BRKT_FILES_CONTENT_TYPE
+        )
+
+        print udc.to_mime_text()
```
krishnanrs: what is this print statement for? will it display the jwt mime on the terminal?

bburtin: Yes.  We just print it to stdout, like we do with the AMI ID during encryption.  This gives the user the flexibility of sending the output to a file or another command.

brkt_cli/__init__.py
```diff
@@ -37,7 +37,12 @@
 
 # The list of modules that may be loaded.  Modules contain subcommands of
 # the brkt command and CSP-specific code.
-SUBCOMMAND_MODULE_NAMES = ['brkt_cli.aws', 'brkt_cli.jwt', 'brkt_cli.gce']
+SUBCOMMAND_MODULE_NAMES = [
+    'brkt_cli.aws',
+    'brkt_cli.gce',
+    'brkt_cli.generate_user_data',
+    'brkt_cli.jwt',
```
bonzotronix: Obviously, the new module name should be "brkt_cli.gud".  :)

brkt_cli/generate_user_data/__init__.py
krishnanrs: lgtm

gcallaghan:
LGTM

bonzotronix:
Could you and I chat a bit about how to migrate the functionality from brkt_make_userdata to brkt-cli?  I'm not trying to migrate it all right now, but it would be good to know roughly how we'll do it so that it doesn't require changing args and such to this new generate-user-data command.

bburtin:
Sure thing. Let's talk tomorrow.

brkt_cli/generate_user_data/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+        parser = subparsers.add_parser(
+            self.name(),
+            description=(
+                'Generate user-data that is passed to cloud-init when '
+                'running an instance.'
+            )
+        )
+        parser.add_argument(
+            '--jwt',
+            required=True,
```
bonzotronix: In thinking about how to merge brkt_make_userdata functionality into brkt-cli, I think the main thing is that we should expect to add options to this command like:
-- adding arbitrary files to the 'brkt-files' part of the userdata (so they get written to /var/brkt/ami_config)
-- adding other config settings to the JSON object that is sent in the 'text/brkt-config' part
   (e.g., the 'Encrypted image file URL' proposed for BYOK on ESX)

So I'd suggest that we make the '--jwt' optional, either now or when the next option is added to generate-user-data.

Also, I'm wondering if 'make-user-data' might be slightly preferable in the expanded context.

bburtin:
Will open a new PR to address the issues that Adam brought up.

PR #276, opened 2016-05-20
Rename generate-jwt to make-jwt

bburtin:
Change the name of the generate-jwt subcommand to make-jwt, per Adam's
suggestion.  "Make" is shorter and consistent with the existing
make_user_data command in cephalopod.


bonzotronix:
LGTM
(thanks!)

PR #285, opened 2016-05-25
Consolidate the retry logic

bburtin:
Update aws_service so that it uses the retry() function in the util
module.  Add the ability to use a RetryExceptionChecker to determine
whether we should retry.  We need this in order to check
EC2ResponseError.error_code.  Change the retry interval to a time period
instead of counting the number of attempts, so that the caller doesn't
have to do math to determine the timeout.

Remove the retry_boto decorator and replace it with a retry_boto()
function.  This gives us the flexibility of retrying either in
AWSService or in the higher-level application code.  Add unit tests for
util.retry().


robertabbott:
lgtm

PR #289, opened 2016-05-31
adding a retry to the GCE DELETE for the encryptor instance

brktporter:
Using the new refactored util.retry() to solve a persistent issue with the first DELETE on the encryptor. The first attempt always seems to result in httplib.BadStatusLine. Retrying a second time succeeds.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -12,7 +12,8 @@ from brkt_cli.validation import ValidationError
 from gce_service import gce_metadata_from_userdata
 from brkt_cli.encryptor_service import wait_for_encryption
 from brkt_cli.encryptor_service import wait_for_encryptor_up
-
+from brkt_cli import util
+import httplib
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -76,7 +77,8 @@ def do_encryption(gce_svc,
 
         wait_for_encryptor_up(enc_svc, Deadline(600))
         wait_for_encryption(enc_svc)
-        gce_svc.delete_instance(zone, encryptor)
+        util.retry(function=gce_svc.delete_instance,
+            on=[httplib.BadStatusLine])(zone, encryptor)
```
bburtin: Note that this won't retry on the exceptions that are specified in gce_service.retry().  Does httplib.BadStatusLine only apply to this call?  If it applies to all GCE API calls, you could add httplib.BadStatusLine to the list of exceptions in gce_service.retry().  Go ahead and make the call with Bobby.  I just wanted to make sure that the behavior is intentional.

robertabbott: Chris is the only person to encounter this issue so if it fixes his problem LGTM

robertabbott:
https://github.com/brkt/brkt-cli/pull/293

PR #290, opened 2016-05-31
Fix JWT generation

bburtin:
Fix the JWT logic so that it generates a valid token.

* Pass separators when encoding JSON, to avoid whitespace.
* Encode JSON as UTF-8.
* Use SHA-384 when signing.
* Move the kid field from the payload to the header.

Testing: successfully ran verify-jwt in Yeti


gcallaghan:
LGTM

PR #292, opened 2016-06-02
Have named suffix for instance console log

dramdass:
This allows us to easily identify/find/archive console logs
from failed runs.


bburtin:
Looks good, thanks!

PR #296, opened 2016-06-07
Don't abort when version check fails

bburtin:
A version check error shouldn't brkt-cli from working.  If there's an
eror when downloading the list of supported versions, print the error
message and continue.  Also, add a timeout in case the request to
pypi.python.org hangs.


krishnanrs:
lgtm

PR #297, opened 2016-06-07
Add support for encrypted signing key

bburtin:
After reading the PEM file, see if it contains an encrypted key.  If it
does, prompt the user for a password.

Use the cryptography library instead of ecdsa, since ecdsa doesn't
support encrypted keys.  Also, the pyjwt library integrates with the
cryptography library for signing and verification.  This change
introduces a dependency on pip 8 or greater for many users.  The
cryptography library links with openssl.  pip versions 8 and later use
Python Wheels do download supporting binaries instead of compiling them
locally.

Details:
* Move cryptography code out of the core codebase and into a new
brkt_cli.crypto package.  This keeps the low-level code out of the core
codebase and allows us to change the underyling implementation later.
* Use the pyjwt library to do the JWT encoding and signing for us.
* Add requirements.txt, for installing 3rd-party dependencies without
the brkt-cli code.

Testing:
* python -m unittest discover
* Manually test with different keys, malformed key files, etc.
* Use the verify-jwt tool in Yeti to check the JWT that is generated by
the CLI.


brkt_cli/jwt/__init__.py
```diff
@@ -49,7 +52,7 @@ def verbose(self, values):
         return values.make_jwt_verbose
 
     def run(self, values):
-        signing_key = read_signing_key(values.signing_key)
+        crypto = read_signing_key(values.signing_key)
```
gcallaghan: why the rename to crypto here?

gcallaghan:
LGTM

PR #298, opened 2016-06-08
Don't allow cnc value that's less than 1

bburtin:
Also, pass the cnc value to make_jwt().  Oops.

Testing:
* python -m unittest discover
* manual


gcallaghan:
LGTM

PR #299, opened 2016-06-08
Revert "Add support for encrypted signing key"

bburtin:
This reverts commit 26923fb0200abdf5c1654601fc278e1424444deb.

We hit problems with installing the cryptography library
on Linux.  Reverting this change until we figure out what
to do.


bburtin:
Closing this pull request.  We may be able to make this work in QA.

PR #300, opened 2016-06-09
YETI-1059: Keep user provided encryptor image

krishnanrs:
If the user provides an encryptor image, do not delete it


jimvoll:
Dear god that was fast. Thanks Krishnan!

PR #301, opened 2016-06-09
Add --customer option to make-jwk

bburtin:
Add a --customer option, which adds a customer UUID to the JWT payload.

### DIFF
--- a/brkt_cli/jwt/__init__.py
+++ b/brkt_cli/jwt/__init__.py
@@ -16,6 +16,7 @@ import json
 import logging
 import re
 import time
+import uuid
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


brkt_cli/jwt/__init__.py
```diff
@@ -222,6 +237,12 @@ def cnc(value):
         return brkt_cli.validation.min_int_argument(value, 1)
 
     parser.add_argument(
+        '--customer',
+        metavar='UUID',
+        type=uuid.UUID,
+        description='Only used with a third party JWK server'
```
gcallaghan: Required for API access when using a 3rd party JWK server

PR #302, opened 2016-06-10
Refactor unit tests

bburtin:
Break up test.py and move the unit tests into modules that are closer to
the code that's being tested.


gdmen:
lgtm

PR #303, opened 2016-06-10
Update brkt-cli to reflect HVM GA

krishnanrs:
YETI-1064: Update brkt-cli to reflect HVM GA (scheduled as part of 06-07 release)


bburtin:
Looks good, thanks!

PR #305, opened 2016-06-10
Fix typo in make-jwt --customer

bburtin:


bburtin:
Krishnan already fixed it.

PR #307, opened 2016-06-10
Add --network option to update and encrypt gce

robertabbott:
Allows user to specify a non-default network
to launch encryptor instances in

Testing: encrypted and updated gce images

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -94,9 +94,10 @@ def command_launch_gce_image(values, log):
 
 
 def command_update_encrypted_gce_image(values, log):
-    check_args(values)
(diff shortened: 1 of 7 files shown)


krishnanrs:
Where are we validating if the network entered is valid? Or are we just letting GCE APIs error out if that happens to be the case?

robertabbott:
we don't check that the network is valid. It will fail when launching the encryptor if the network is invalid

bburtin:
Bobby's approach should be ok, as long as we fail quickly, we don't show a traceback, and the error message is clear.  If this isn't the case, we should be proactive about validating user input.

krishnanrs:
lgtm

PR #309, opened 2016-06-11
Add http --status-port for encryptor/updater

dramdass:
Default port is 8000 but allow specifying custom port. The
encryptor knows which port to use because we put it in userdata.


dramdass:
This isn't ready to merge yet until the encryptor change is in

bonzotronix:
These changes LGTM.  Do we usually re-submit a new PR with squashed changes?

BTW, our long arg lists to functions like encrypt() are getting way out of hand.  I suspect we'll need to pass around Config objects or something like that soon.

dramdass:
@bonzotronix I'll resubmit. Small issue still needed to be resolved here is some ports cannot be specified  because they are already used by other services. I'll add that validation check and resubmit.

dramdass:
Closed since it is tracked in another PR

PR #310, opened 2016-06-13
Rename brkt_cli.jwt to brkt_cli.brkt_jwt

bburtin:
The brkt_cli.jwt module name was creating problems with our subcommand
loading code.  If the user didn't have pyjwt installed, module loading
silently failed and we wouldn't register the make-jwt command.  This is
because ImportError doesn't give us a way to know which jwt module was
the one that failed to load.

Rename brkt_cli.jwt to brkt_cli.brkt_jwt to avoid this ambiguity.  Also,
change "raise e" in the subcommand loading code to "raise", so that we
preserve the original ImportError traceback.


mpage:
LGTM

PR #312, opened 2016-06-13
Update install instructions

bburtin:
Update README.md with instructions and requirements for installing the
cryptography library.  This library has a binary component, which
requires pip 8 on Windows and OS X, and system packages on Linux.


krishnanrs:
Are we going with the assumption that users will be fairly familiar with virtualenv if they are using Linux? Is it possible to add more information regarding virtualenv installation and using brkt-cli within the same?

bburtin:
Yeah, I wrestled with that issue.  I'm leaning toward keeping the docs terse, since it would take me a paragraph to explain the reasons why you'd want to use virtualenv.  That said, the description on the virtualenv website doesn't do a great job of explaining it.  Let me poke around and see if I can find a good article that explains virtualenv.  Let me know if you already know of one.

bburtin:
I added a link to The Hitchhiker's Guide To Python.  They do a pretty good job of explaining virtualenv.

krishnanrs:
How about a youtube video? 👍 
LGTM

dasein:
LGTM

PR #313, opened 2016-06-14
Handle edge case errors when calling the AWS API

bburtin:
Handle two obscure errors that occasionally pop up in the field.  I
believe that the AWS service returns 503 when the request rate limit has
been exceeded for a given account.  I'm not sure what causes
ssl.SSLError, but it has been reported a couple of times.

To handle these errors, we:

* Check for these cases in BotoRetryExceptionChecker.
* Wrap every AWS call with retry_boto()

Also, make BotoRetryExceptionChecker.error_code_regexp an optional
argument.  If the regexp isn't specified, we check for the two cases
above.

Change "raise e" to "raise" in util.retry(), so that the original
traceback is preserved.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -14,13 +14,14 @@
 
 import abc
 import re
+import ssl
 import tempfile
 
 import boto
 import boto.sts
 import boto.vpc
 import logging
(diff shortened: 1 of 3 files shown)


mpage:
LGTM
:shipit:

PR #315, opened 2016-06-15
Add Docker support

bburtin:
Add a Dockerfile for building a container that runs brkt-cli.  Update
the README with examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,15 @@ In order to use the Bracket service, you must be a
 registered Bracket customer.  Email support@brkt.com for
 more information.
 
-**brkt-cli** requires Python 2.7.  We also recommend using
-[virtualenv](https://virtualenv.pypa.io/), to avoid conflicts between
-brkt-cli dependencies and Python packages that are managed by the operating
-system.  If you're not familiar with virtualenv, check out the
+**brkt-cli** requires Python 2.7.
+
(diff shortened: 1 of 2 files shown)


krishnanrs:
👏

README.md
```diff
@@ -235,3 +238,32 @@ ami-63733e09
...
+```
+$ docker run -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+-e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
+brkt encrypt-ami --region us-west-2 ami-4133cb21
+```
+
+```
+$ docker run -v ~/keys:/keys brkt make-jwt --signing-key /keys/secret.pem
+eyJhbGciOiJFUzM4NCIsInR5cCI6IkpXVCIsImtpZCI6ImU2MTNhYzI0YzRkN2ExY...
+```
```
dasein: I might just add a example here of how to deal with artifacts that are _written_ into the container (e.g. --encrypted-image-target-directory /somedir). Essentially it will be the responsibility of the operator to either mount a volume at the point where they are writing out the artifact, or copy the artifact out of the container after it has run.

bburtin: Currently brkt-cli only outputs to stdout.  If we need to write to disk in the future, I'll update the instructions.

PR #316, opened 2016-06-15
Use https for getting amis from s3

dramdass:
http is liable to be cached by an internet proxy without the
right headers.

https is also the right thing to do from a security perspective.
Note that now, this verifies the s3 server certificate so your
python version should support this.

Testing done:
- ran locally on my Mac and virtualenv


bburtin:
:shipit:

krishnanrs:
Should we explicitly call out that they will need python version 2.7.9 or above? 💭  
🆗

dramdass:
@krishnanrs does it need python > 2.7.9 or above? I don't know that for sure

krishnanrs:
Verified on python 2.7.6 as well. 👍 💯

PR #317, opened 2016-06-16
Moved 'kid' field from payload to header of JWT

bonzotronix:
* Per Grant, the 'kid' field should be in the header and not in the payload.  The JWTs created by 'brkt make-jwt' already put the 'kid' in the header.

Testing:
* Ran unit tests.
* Ran "make-user-data --jwt <data output by 'brkt make-jwt'>"


bburtin:
👍  Thanks for fixing this!

PR #319, opened 2016-06-17
Fix verbose logging check when printing exceptions

bburtin:
Check for debug logging on the Logger object instead of using
values.verbose.  values.verbose only works when -v is specified at the
top level, not at the subcommand level.

Clean up duplicate logging initialization code.


dramdass:
lgtm!

PR #323, opened 2016-06-21
Add http --status-port for encryptor/updater

dramdass:
Default port is 8000 but allow specifying custom port. The
encryptor knows which port to use because we put it in userdata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ In order to use the Bracket service, you must be a
 registered Bracket customer.  Email support@brkt.com for
 more information.
 
+**brkt-cli** requires Python 2.7.
+
 We recommend using [virtualenv](https://virtualenv.pypa.io/), to avoid
 conflicts between **brkt-cli** dependencies and Python packages that are managed
 by the operating system.  If you're not familiar with virtualenv, check out the
--- a/README.md
+++ b/README.md
@@ -26,8 +26,6 @@ In order to use the Bracket service, you must be a
 registered Bracket customer.  Email support@brkt.com for
 more information.
 
-**brkt-cli** requires Python 2.7.
-
 We recommend using [virtualenv](https://virtualenv.pypa.io/), to avoid
 conflicts between **brkt-cli** dependencies and Python packages that are managed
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -26,8 +26,6 @@ In order to use the Bracket service, you must be a
 registered Bracket customer.  Email support@brkt.com for
 more information.
 
-**brkt-cli** requires Python 2.7.
```
bburtin: Is this an intentional change?  I'm pretty sure our stuff won't work with Python 2.6 and 3.x.

dramdass: that's a bad merge. I'll re-add it in

README.md
```diff
@@ -87,7 +85,8 @@ The master branch has the latest features and bug fixes, but is not as thoroughl
 
 The following network connections are established during image encryption:
 
-* **brkt-cli** talks to the Encryptor instance on port 8000.
+* **brkt-cli** talks to the Encryptor instance on port 8000 by default. This can
+be overriden using the --status-port flag which support any port other than port 81.
```
bburtin: "overridden" is misspelled.  Would it make sense to call the option --status-port instead of --encryptor-status-port, so that it's more specific?

brkt_cli/aws/aws_service.py
```diff
@@ -550,6 +550,23 @@ def validate_image_name(name):
     return name
 
 
+def validate_status_port(status_port):
```
bburtin: This should be in encryptor_service.py.

README.md
dramdass: Will fix overridden. The option is called --status-port since it can be used for encryptor and updater.

brkt_cli/aws/aws_service.py
bburtin: Actually... you could move this to the argparse code.  If you use `parser.add_argument(..., type=my_function)`, argparse will call your function to parse the value.  You can also set a default value for the argument.

dramdass: OK I'll move it there

bonzotronix:
LGTM.  Above-and-beyond on the port validation front.  :)

PR #324, opened 2016-06-21
Remove unused status port validation cruft

dramdass:
This code is no longer used since validation is done with
argparse.


bonzotronix:
LGTM

PR #327, opened 2016-06-21
Wait for volume to detach

bburtin:
After detaching a volume, wait for it to be in the "available" state
before proceeding.  This should take care of "volume is still attached
to instance" errors.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -375,7 +375,7 @@ def command_encrypt_ami(values, log):
             parsed_cert = x509.load_pem_x509_certificate(ca_cert_data,
                                                          default_backend())
         except Exception as e:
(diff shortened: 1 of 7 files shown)


krishnanrs:
The default 30s timeout might not be sufficient especially with the eventual consistency problem we keep hitting in quite often and can cause the encryption to abort at various stages, depending on where the problem occurred. In Yolo we have had to set the timeout to 10 mins. to avoid this issue. Otherwise LGTM.

bburtin:
I came up with 30 seconds because I thought that the current code mostly works with no timeout.  Does it really take several minutes to detach a volume, or are you talking about timeouts for other operations?

dramdass:
Attach/detach can take > 30 seconds a fair number of times. I've seen it take > 5 minutes a few times. I think 10 minutes is reasonable.

PR #328, opened 2016-06-22
YETI-1103: Make HVM encryptor the default

krishnanrs:
Now that HVM is GA, make the HVM encryptor the default one.

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -120,9 +120,8 @@ SLEEP_ENABLED = True
 AMI_NAME_MAX_LENGTH = 128
 
 BRACKET_ENVIRONMENT = "prod"
-ENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/amis.json"
-HVM_ENCRYPTOR_AMIS_URL = \
-    "https://solo-brkt-%s-net.s3.amazonaws.com/hvm_amis.json"
+PV_ENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/amis.json"
+HVENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/hvm_amis.json"
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -40,10 +40,10 @@ def setup_encrypt_ami_args(parser):
         default='m3.medium'
     )
     parser.add_argument(
-        '--hvm',
```
bburtin: Are we concerned about backward compatibility?

krishnanrs: The brkt-cli will print errors if there is a virtualization type mis-match. For backwards compatibility there is always the --pv flag still available. Users already having hvm AMIs, no longer have to specify the --hvm flag.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -201,15 +201,15 @@ def get_encrypted_suffix():
...
+def get_encryptor_ami(region, pv=False):
     bracket_env = os.getenv('BRACKET_ENVIRONMENT',
                             BRACKET_ENVIRONMENT)
     if not bracket_env:
         raise BracketError('No bracket environment found')
-    if hvm:
-        bucket_url = HVM_ENCRYPTOR_AMIS_URL % (bracket_env)
-    else:
+    if pv:
         bucket_url = ENCRYPTOR_AMIS_URL % (bracket_env)
```
bburtin: Can you rename this to `PV_ENCRYPTOR_AMIS_URL`?  Probably better to qualify the name, now that it's not the default.

brkt_cli/aws/encrypt_ami_args.py
bburtin: What I meant was, will users be annoyed if their automation scripts start failing because we removed a CLI option?  If you think this is not an issue, keep it as is.

krishnanrs: Good point. That's what documentation is for :) But let me check with the QA as well as field folks just to be sure.

krishnanrs: I checked with Danny on the QA front and he was OK with that (I will notify him so that he can update the encryptor jobs once the merge is done). On the customer front, we are fine (confirmed with Mark and John).

bburtin:
Looks good, aside from the URL naming nit.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -120,9 +120,8 @@
 AMI_NAME_MAX_LENGTH = 128
 
 BRACKET_ENVIRONMENT = "prod"
-ENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/amis.json"
-HVM_ENCRYPTOR_AMIS_URL = \
-    "https://solo-brkt-%s-net.s3.amazonaws.com/hvm_amis.json"
+PV_ENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/amis.json"
+HVENCRYPTOR_AMIS_URL = "https://solo-brkt-%s-net.s3.amazonaws.com/hvm_amis.json"
```
bburtin: Typo?

bburtin:
👍

brkt_cli/aws/encrypt_ami_args.py
dramdass: In the future, I think the standard behavior here is to have the deprecated flag (--hvm) to complain/give a warning for 1-2 releases (e.g. --hvm is deprecated, it will be removed in brkt-cli 1.x) while it continues to work.

PR #329, opened 2016-06-22
Made JWT input be string for make-user-data

bonzotronix:
* In PR#321, I switched the '--jwt' input to make-user-data be a filename instead of a string.  But other commands use JWTs as strings, so make-user-data should follow that convention.


brkt_cli/make_user_data/__init__.py
```diff
@@ -56,10 +56,10 @@ def verbose(self, values):
         return values.make_user_data_verbose
 
     def run(self, values):
-        with open(values.jwt, 'r') as f:
-            token_val = f.read().rstrip()
-
-        print combine_user_data(brkt_config={}, jwt=token_val)
+        if not values.jwt:
```
bburtin: How about making the the argument required, so that it's documented in the `--help` output?

bburtin:
Thanks for catching the inconsistency in the CLI spec.  I totally missed it in PR 321.

brkt_cli/make_user_data/__init__.py
bonzotronix: Sounds good.  See new commit.

bburtin: Cool.  I don't think you need the `if not values.jwt` part any more because argparse will do the validation for you.

bonzotronix: OK.  Submitting a new PR with squashed commits (cuz it would suck to have 3 piddly commits where 1 piddly commit will do).

bonzotronix:
Added latest fix and squashed commits in new PR: #331

PR #321, opened 2016-06-20
Put JWT in brkt_config, not as file

bonzotronix:
* Instead of putting the JWT in a brkt-file with dest path /var/brkt/ami_config/token.jwt, put the JWT in the brkt_config dictionary with key 'identity_token'.

Testing:
* Ran unit tests
* Ran 'brkt make-user-data -jwt <file_with_jwt>'

PR #331, opened 2016-06-23
Made JWT input be string for make-user-data

bonzotronix:
[squashed version of #329]
* In #321, I switched the '--jwt' input to make-user-data be  a filename instead of a string.  But other commands use JWTs as strings, so make-user-data should follow that convention.


bburtin:
👍

PR #332, opened 2016-06-23
Choose metavisor AMI by virtualization type

bburtin:
Default to using the metavisor AMI that has the same virtualization type
as the guest or encrypted image.  Users can continue to use --pv to
override the default behavior with HVM.

Change _validate_guest_ami() to raise ValidationError when validation
fails instead of returning an error message.  This matches the behavior
of other validation functions.

Move _get_encryptor_ami() into brkt_cli.aws, since we call this function
while processing command line arguments.

Remove test_no_validate().  It no longer makes sense, now that
_validate() is being called conditionally based on the state of the
--no-validate option.

Testing:
* python -m unittest discover
* Manually run encrypt-ami and update-encrypted-ami on PV and HVM images
and make sure that the correct metavisor AMI is selected


krishnanrs:
🚨 
👆 
😴 
👍

bburtin:
Are these the four stages that you go through when reading my code?  Alarm, finger-pointing, boredom, and acceptance?

PR #335, opened 2016-07-05
Consolidate instance config arg handling

bonzotronix:
* Created InstanceConfig class with methods to generate userdata and gce_userdata.  Its ctor takes a 'values' list, as returnd by argparse.parse_args().

* Reduced the input args to the 'encrypt' and 'update' functions for aws and gce (substituting an InstanceConfig arg).

* Added helper functions in instance_config_args to be used by commands that launch Metavisor instances.

* This has the effect of making all instance-launching commands (encrypt-ami, update-encrypted-ami, encrypt-gce-image, update-encrypted-gce-image, launch-gce-image) now consistently support these args:
  --brkt-env, --proxy, --proxy-config-file, --ntp-server, --jwt.
  Also, all the 'creator' mode commands (all the above except launch-gce-image) now support the --ca-cert arg as well.

* Added unit tests for the new instance_config stuff.

* Fixed a number of glitches flagged by pep8 and pylint.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image_args.py
+++ b/brkt_cli/gce/encrypt_gce_image_args.py
@@ -28,20 +28,23 @@ def setup_encrypt_gce_image_args(parser):
         default='prod',
         required=False
     )
+
+    # The email and password options will be deprecated soon (per Grant).
     parser.add_argument(
          '--brkt-email',
          metavar='EMAIL',
          dest='api_email',
          help='Bracket user email',
-         required=True
-     )
+         required=False  # arg is optional because the --jwt arg may be used instead
+    )
     parser.add_argument(
          '--brkt-password',
          metavar='PASSWORD',
          dest='api_password',
          help='Bracket user password',
-         required=True
-     )
+         required=False  # arg is optional because the --jwt arg may be used instead
+    )
+
     parser.add_argument(
         '--project',
         help='GCE project name',
@@ -66,16 +69,8 @@ def setup_encrypt_gce_image_args(parser):
         default='default',
         required=False
     )
-
-    # Optional yeti endpoints. Hidden because it's only used for development.
-    # If you're using this option, it should be passed as a comma separated
-    # list of endpoints. ie blb.*.*.brkt.net:7002,blb.*.*.brkt.net:7001 the
-    # endpoints must also be in order: api_host,hsmproxy_host
-    parser.add_argument(
-        '--brkt-env',
-        dest='brkt_env',
-        help=argparse.SUPPRESS
-    )
+    # Optional Image Name that's used to launch the encryptor instance. This
+    # argument is hidden because it's only used for development.
     parser.add_argument(
         '--encryptor-image-file',
         dest='image_file',
@@ -88,6 +83,3 @@ def setup_encrypt_gce_image_args(parser):
         action='store_true',
         help=argparse.SUPPRESS
     )
-
-    # Optional Image Name that's used to launch the encryptor instance. This
-    # argument is hidden because it's only used for development.
--- a/brkt_cli/instance_config.py
+++ b/brkt_cli/instance_config.py
@@ -1,0 +1,87 @@
+# Copyright 2015 Bracket Computing, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# https://github.com/brkt/brkt-cli/blob/master/LICENSE
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import logging
+import os
+
+from brkt_cli.user_data import UserDataContainer
+from brkt_cli.validation import ValidationError
+
+# The directories for files saved on the Metavisor. We require that the dest
+# path for all --brkt-files be within this directory.
+#
+# If the Metavisor is in 'creator' mode, it will only save files
+# to '/var/brkt/ami_config'.
+# If the Metavisor is in 'metavisor' mode, it will only save files
+# to '/var/brkt/instance_config'.
+
+BRKT_FILE_AMI_CONFIG = '/var/brkt/ami_config'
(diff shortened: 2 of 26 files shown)


brkt_cli/instance_config.py
```diff
@@ -0,0 +1,160 @@
...
+        if values.jwt:
+            self.brkt_config['identity_token'] = values.jwt
+        elif 'api_email' in values and 'api_password' in values:
+            if not self.brkt_env:
+                raise ValidationError(
+                    'Must specify --brkt-env when using email and password to get token'
+                )
+            token = _get_identity_token(self.brkt_env, values.api_email,
+                                        values.api_password)
+            self.brkt_config['identity_token'] = token
```
bonzotronix: We could raise a ValidationError if the user supplies both "--jwt" and either the "--brkt-email" or "--brkt-password" options.  Currently, we give the "--jwt" argument precedence and ignore the other args when "--jwt" is supplied.

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -28,20 +28,23 @@ def setup_encrypt_gce_image_args(parser):
...
+    # The email and password options will be deprecated soon (per Grant).
     parser.add_argument(
          '--brkt-email',
          metavar='EMAIL',
          dest='api_email',
          help='Bracket user email',
-         required=True
-     )
+         required=False  # arg is optional because the --jwt arg may be used instead
+    )
```
bonzotronix: I neglected to add a check that the instance config has an 'identity_token' (after the 'values' are processed) for both encrypt-gce-image and update-encrypted-gce-image.  Probably should add that.

brkt_cli/instance_config.py
```diff
@@ -0,0 +1,160 @@
...
+class BrktFile(object):
+    def __init__(self, dest_filename, file_contents):
+        self.dest_filename = dest_filename
+        self.file_contents = file_contents
+
+
+class InstanceConfig(object):
+    """ Class containing common settings for Brkt instances """
+
+    def __init__(self, values=None, mode=INSTANCE_CREATOR_MODE):
```
bburtin: I don't know if it's a good idea for the InstanceConfig constructor to depend on the structure of the values object.  It's a higher-level object, which is created at argument parsing time.  Would it make more sense to pass in `brkt_config`, `brkt_files`, etc. as separate fields?

bonzotronix: Well, the 'values' input is just a Namespace() object.  It looks like
`Namespace(brkt_env='api.example.com:777,hsmproxy.example.com:888', ca_cert='/Users/acain/src/brkt-cli-rw/brkt_cli/assets/ca_cert.pem', ntp_servers=['10.4.5.6'], proxies=None, proxy_config_file=None)`
That seems to me to be a pretty handy input type.  I could add comments clarifying that it's a Namespace (with field names defined in setup_instance_config_args().

bburtin: It's handy, but you're writing lower-level code that makes assumptions about the structure of a higher-level object.  This tends to be brittle because the structure of the higher-level object can change.  What about adding a function at the argument parsing level that returns an InstanceConfig object based on a values object?

bonzotronix: OK, will work on that.

brkt_cli/instance_config.py
```diff
@@ -0,0 +1,160 @@
...
+            add_brkt_env_to_brkt_config(self.brkt_env, self.brkt_config)
+
+        proxy_config = get_proxy_config(values)
+        if proxy_config:
+            proxy_config_filename = self.brkt_files_dest + '/proxy.yaml'
+            self.add_brkt_file(proxy_config_filename, proxy_config)
+
+        if 'ca_cert' in values and values.ca_cert:
+            if not values.brkt_env:
+                raise ValidationError(
```
bburtin: This type of validation logic looks like it should run during argument parsing.  Possibly another reason to move the code into a values->InstanceConfig function?

brkt_cli/instance_config.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def add_brkt_file(self, dest_filename, file_contents):
+        brkt_file = BrktFile(dest_filename, file_contents)
+        self.brkt_files.append(brkt_file)
+
+    def _make_brkt_config_json(self):
+        brkt_config_dict = {'brkt': self.brkt_config}
+        return json.dumps(brkt_config_dict, sort_keys=True)
+
+    def make_userdata(self, do_gzip=False):
```
bburtin: Do we need the do_gzip argument here?  Seems like it would be clearer if the caller called a separate function to do the gzipping.

brkt_cli/instance_config.py
```diff
@@ -0,0 +1,160 @@
...
+            udc.add_file(brkt_file.dest_filename, brkt_file.file_contents,
+                         BRKT_FILES_CONTENT_TYPE)
+
+        user_data_string = udc.to_mime_text()
+
+        if do_gzip:
+            return gzip_user_data(user_data_string)
+        return user_data_string
+
+    def make_gce_metadata(self, extra_items=None):
```
bburtin: Is this something that can be moved into the GCE package?

brkt_cli/instance_config.py
bonzotronix: Yup

brkt_cli/instance_config.py
bonzotronix: Either way is fine with me.  It's here as a convenience.  Splitting it out would mean an extra couple lines added to each of encrypt_ami and update_ami.

brkt_cli/instance_config.py
bonzotronix: I'll give that a try.

bonzotronix:
Made changes from feedback on previous commit.
PTAL

brkt_cli/test_user_data.py
```diff
@@ -11,58 +11,59 @@
...
+        self.assertEqual(bogus_payload, None)
 
-        msg = email.message_from_string(mime_data)
-        found_brkt_config = False
-        found_brkt_files = False
+    def test_add_files_and_config(self):
+        udc = UserDataContainer()
+
+        file1_contents = 'Never gonna give you up.'
+        file2_contents = 'Never\ngonna\tlet you\n\ndown!!'
```
bburtin: You may want to replace this reference with something more critically acclaimed, like Fine Young Cannibals.

bburtin:
👍  Thanks for cleaning this up!

PR #336, opened 2016-07-06
Switch to jwt authentication

robertabbott:
Removes username and password authentication.
The --brkt-token arg is now used to pass a
jwt through userdata to the encryptor.

The jwt must be created with the make-jwt
command

Testing: encrypted an image in gce


krishnanrs:
LGTM. But before we push out this change to customers, we should document the steps for generating JWTs as this will be an additional operation that users will have to perform.

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -29,17 +29,10 @@ def setup_encrypt_gce_image_args(parser):
...
-         dest='api_email',
-         help='Bracket user email',
-         required=True
-     )
-    parser.add_argument(
-         '--brkt-password',
-         metavar='PASSWORD',
-         dest='api_password',
-         help='Bracket user password',
+         '--brkt-token',
```
bburtin: We've been calling this `--jwt` in other subcommands.  I'm not tied to one of the other, but we should have consistent terminology.

brkt_cli/gce/__init__.py
```diff
@@ -118,7 +117,7 @@ def command_update_encrypted_gce_image(values, log):
         encrypted_image_name=encrypted_image_name,
         zone=values.zone,
         brkt_env=brkt_env,
-        token=token,
+        token=values.token,
```
bburtin: Would it make sense to call `brkt_cli.validate_jwt()` instead of blindly passing through the user-specified value?

bonzotronix:
Could you please hold off on merging this PR?
It will conflict mightily with my latest one (#335) and I suspect it will be much easier to make these changes once #335 is merged in (compared to re-working #335 after this one is merged).

robertabbott:
I'll hold off

bonzotronix:
OK, I merged in #335.  I'm hoping rebasing these changes won't be too tricky.

PR #337, opened 2016-07-06
Bobby/token

robertabbott:


brkt_cli/gce/__init__.py
```diff
@@ -167,11 +167,5 @@ def check_args(values, gce_svc):
...
             raise ValidationError("Please provided either an encryptor image or an image bucket")
-    if values.jwt:
-        if values.api_email:
-            log.warning('Since JWT is provided, --brkt-email arg is ignored')
-        if values.api_password:
-            log.warning('Since JWT is provided, --brkt-password arg is ignored')
-    elif not values.api_email or not values.api_password:
-        raise ValidationError(
-            'Must provide either a JWT or both a brkt-email and a brkt-password')
+    if not values.token:
```
bburtin: I believe that this is not necessary, since you pass `required=True` to `add_argument()`.

bburtin:
There are other subcommands that use `--jwt`:

>
$ ack --py -- --jwt
brkt_cli/aws/encrypt_ami_args.py
200:        '--jwt',

brkt_cli/aws/update_encrypted_ami_args.py
190:        '--jwt',

brkt_cli/make_user_data/__init__.py
40:            '--jwt',

robertabbott:
those args are removed in https://github.com/brkt/brkt-cli/pull/335

brkt_cli/gce/__init__.py
```diff
@@ -167,11 +167,5 @@ def check_args(values, gce_svc):
...
+        raise ValidationError('Must provide either a token')
```
bonzotronix: Wording nit: "Must provide a token"

brkt_cli/gce/__init__.py
robertabbott: since the instance_config stuff is also used by aws the token argument can't be required (since it needs to be optional there). So this check is necessary

brkt_cli/instance_config_args.py
```diff
@@ -84,18 +83,19 @@ def setup_instance_config_args(parser, mode=INSTANCE_CREATOR_MODE,
...
+         'JSON Web Token that the encrypted instance will use to '
+         'authenticate with the Bracket service.  Use the make-jwt '
+         'subcommand to generate a JWT.'
+     )
+     """
     parser.add_argument(
-        '--jwt',
-        help=argparse.SUPPRESS
+        '--token',
+        help='Bracket token, created with make-token',
```
bburtin: The docstring above is what I was originally going to use for the help section of this option.  Did you think it was too verbose?  We should either swap it in or delete it.

bburtin:
The code looks good.  Can you update README.md with the latest usage sample?  Also, please add some details to the commit message, so that we have a record of why this change was made.

bonzotronix:
ISTM we should at least get a plan on updating the QA jobs that use --brkt-email & --brkt-password (e.g. yolo/meta/gcp/image_encryptor.py) before we merge this change.

robertabbott:
im working on that now

robertabbott:
those change have been made so I'd like to merge this now

bburtin:
Thanks Bobby.  Can you please address the docs issues:
# Update the commit message and explain what the change is and why you're making it
# Update usage samples in README.md

bburtin:
Never mind.  Ship it!

PR #338, opened 2016-07-07
Move private key loading from JWT to util

bburtin:
Refactor the code that reads a private key from a PEM file, so that it
can be called from both the JWT code and the upcoming key generation
commands.


brkt_cli/brkt_jwt/__init__.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 from __future__ import print_function
 
-import getpass
```
bonzotronix: Looks like you can remove the import of brkt_cli.crypto too.

bonzotronix:
LGTM

PR #339, opened 2016-07-07
Add commands for generating private and public keys

bburtin:
Add two subcommands for generating private and public keys.  These are
just wrappers around OpenSSL for convenience, so that our users don't
have to remember the cryptic OpenSSL options.

```
$ ./brkt make-private-key -h
usage: brkt make-private-key [-h] [--password] [-v]

Generate a 384-bit ECDSA private key (NIST P-384) in OpenSSL PEM format.

optional arguments:
-h, --help     show this help message and exit
--password     Prompt for a password and generate an encrypted private
key
-v, --verbose  Print status information to the console

$ ./brkt get-public-key -h
usage: brkt get-public-key [-h] [-v] PATH

Print the public part of a private key. If the private key is encrypted,
prompt for a password and decrypt the key.

positional arguments:
PATH           Path to a 384-bit ECDSA private key (NIST P-384) in
OpenSSL
                 PEM format

optional arguments:
-h, --help     show this help message and exit
-v, --verbose  Print status information to the console
```


bonzotronix:
Would it be any better to add a single "make-keypair" command that always outputs the public key to a <dest>.pub file (and the password-encrypted private key to the <dest> file)?

bburtin:
I thought about that approach, but opted against it because it makes assumptions about the path to the public key file.  I'm currently leaning toward adding a `\-\-public-out` option to `make-private-key`, and letting the user specify the path to the public key output file.  Does that sound reasonable?

bburtin:
Closing this PR.  Will reopen a new one for the updated CLI spec.

PR #341, opened 2016-07-09
Adds `brkt share-logs`, `brkt diag` 

dramdass:
brkt share-logs subcommand allows users to share a snapshot of the Bracket
system log volume from an instance in the customer AWS account
with the Bracket AWS account. This is intended to be used by customers to
aid Bracket debugging in cases where logs are not reaching the Bracket
servers (e.g. cases with network connecitivity issues from an instance).

brkt diag subcommand will launch an instance with the log volume attached
and readable. It can grab the logs from an instance given instance
id or from a snapshot id. The diag instance is launched with
a specified ssh key with a security group that has port 22 open.

Sample developer use case: 
* Run `brkt diag --instance-id`
* SSH to debug instance and view logs

Sample customer user case: 
* Customer runs `brkt share-logs`
* Bracket support gets notified of new snapshot log (mechanism TBD)
* Bracket support runs `brkt diag --snapshot-id`

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -22,7 +22,11 @@ from boto.exception import EC2ResponseError, NoAuthHandlerFound
 
 import brkt_cli
 from brkt_cli import encryptor_service, util
-from brkt_cli.aws import aws_service, encrypt_ami, share_logs
+from brkt_cli.aws import (
+    aws_service,
+    encrypt_ami,
+    share_logs
+)
 from brkt_cli.instance_config_args import (
     instance_config_from_values,
     setup_instance_config_args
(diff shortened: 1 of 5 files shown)


krishnanrs:
Overall looks good to me. One more suggestion would be to have a standard naming convention for the snapshot so that it is easier to search for the snapshot based on the name. For e.g. <customerAWSID>-mvlogs-<timestamp>.

dramdass:
customerAWSID is not easy to determine from the command line. <instance-id> is possible for sure - is that better?

krishnanrs:
I was just suggesting a common syntax for the name which will make it easier to search based on:
 - Customer
 - Timestamp
 - Possibly some other attributes that we can think of (a reason string, like a NUC or Ticket number etc.)
To begin with, select some naming convention you are comfortable with atleast a timestamp. If this use case expands and I see a need, will suggest changes as necessary (or just make them myself).

dramdass:
The formatting of my previous message got messed up. I meant to say "instance-id" is possible. Getting a customer unique identifier is going be tougher.

krishnanrs:
👏 overall lgtm. I have never tried log mounts with a FreeBSD image, but I presume you have tested that and it works.

dramdass:
Tested a PV snapshot attaching to FreeBSD. I will test an HVM snapshot too to confirm.

PR #343, opened 2016-07-12
Clean up old SLEEP_ENABLED code

bburtin:
We moved the SLEEP_ENABLED global from encrypt_ami to util a while back.
This commit removes the original variable and cleans up the remaining
callsite.


dramdass:
LGTM

PR #344, opened 2016-07-12
Handle temp security group not available

bburtin:


dramdass:
LGTM. The retry should also be applied to the brkt diag command - looks like it should be trivial to do. I'll do that myself in another PR

PR #345, opened 2016-07-13
Sort command line options

bburtin:
Use a custom HelpFormatter to sort options for all commands in
alphabetical order.  Rename share-logs' positional argument to be in all
caps, to follow convention.


dramdass:
LGTM

PR #347, opened 2016-07-14
Snapshots log volume for failed encryption

dramdass:
Adds a flag to specify whether to save logs from encryptor
instance in a snapshot. This is by default True.

Allows `brkt share-logs` to accept a --snapshot-id.

Use case for failed encryption is customer runs an encryption
and if it fails, the snapshot of the log volume is taken and
the brkt share-logs command is logged for users if they then
choose to share with Bracket support.


brkt_cli/aws/encrypt_ami.py
```diff
@@ -597,9 +605,52 @@ def log_exception_console(aws_svc, e, id):
         )
 
 
+def snapshot_log_volume(aws_svc, instance_id):
```
bburtin: Would it make more sense to put this in `brkt_cli/aws/__init__.py`, since it's not specific to encryption?

brkt_cli/aws/share_logs.py
```diff
@@ -79,7 +28,11 @@ def _share_snapshot(snapshot, bracket_aws_account):
     snapshot.share(user_ids=[bracket_aws_account])
 
 
-def share(aws_svc=None, instance_id='', bracket_aws_account=''):
-    snapshot = snapshot_log_volume(aws_svc, instance_id)
+def share(aws_svc=None, instance_id='', bracket_aws_account='',
```
bburtin: I know this is late, but should these be `None` instead of `''`?

brkt_cli/aws/share_logs_args.py
```diff
@@ -16,8 +16,13 @@
 
 def setup_share_logs_args(parser):
     parser.add_argument(
-        'instance_id',
-        metavar='instance-id',
+        '--snapshot-id',
```
bburtin: For a future commit, the convention is `--security-group` and `--subnet`.  So these should probably be `--snapshot` and `--instance`.

brkt_cli/aws/encrypt_ami.py
dramdass: `snapshot_log_volume` uses `wait_for_snapshots` and `cleanup` so just moving `snapshot_log_volume` would create a circular dependency (solvable by inline imports but messy).

Should I move both `wait_for_snapshots` and `cleanup`into `brkt_cli/aws/__init__.py` as well?

bburtin: Yeah, those should be in `__init__.py` as well.  Probably best to do this in a followup commit.

bburtin:
Looks good, thanks!

PR #348, opened 2016-07-14
Generate a default GCE instance name

krishnanrs:


brkt_cli/gce/launch_gce_image_args.py
```diff
@@ -14,7 +14,7 @@ def setup_launch_gce_image_args(parser):
         metavar='NAME',
         dest='instance_name',
         help='Name of the instance',
-        required=True
+        required=False
```
bburtin: You can just omit this.  The default is `False`.

bburtin:
Ship it!

PR #349, opened 2016-07-14
Default don't snapshot failed encryptor log volume

dramdass:
Turns --no-save-encryptor-logs to --save-encryptor-logs which is
False by default and hidden.

Cleanup snapshot and instance id flags for the share-logs and
diag commands.


krishnanrs:
lgtm

PR #350, opened 2016-07-15
Validate the encrypted AMI id when updating

bburtin:
When updating an encrypted AMI, make sure that the AMI exists before
proceeding.


krishnanrs:
👍

PR #354, opened 2016-07-18
Do not use tempdir for console logs

dramdass:
These logs need to be persisted, especially for automated
archival.


brkt_cli/aws/encrypt_ami.py
```diff
@@ -494,9 +494,7 @@ def write_console_output(aws_svc, instance_id):
     try:
         console_output = aws_svc.get_console_output(instance_id)
         if console_output.output:
-            prefix = instance_id + '-'
-            with tempfile.NamedTemporaryFile(
-                    prefix=prefix, suffix='-console.txt', delete=False) as t:
+            with open(instance_id + '-console.txt',  'w') as t:
```
bburtin: If the current directory isn't writable, the user will get an "Unable to write console output" error with no context.  At the very least, we should log the path that we couldn't write to.  I also wonder whether the console logs should go into a known directory (`~/.brkt`?)  That behavior seems clearer and more reliable than writing to the current directory.

Any guess as to how other CLI tools deal with this?

dramdass: The Docker use case really needs something mounted in the current workspace. I'll let @dasein expound on what's possible or not.

.gitignore
```diff
@@ -7,3 +7,6 @@
 *.egg-info/
 build/
 dist/
+
+# Ignore instance console logs from brkt-cli
+*-console.txt
```
bburtin: I believe that this is no longer necessary?

PR #356, opened 2016-07-19
Use NetBSD instead of FreeBSD for diag instances

dramdass:
FreeBSD didn't like our NetBSD disk label so the log partition
for HVM could not be read.

Use NetBSD official images and change the mountpoint and output
to user correspondingly.

Testing done:
- tested diag for a PV and metavisor instance and verified
mounts.

### DIFF
--- a/brkt_cli/aws/diag.py
+++ b/brkt_cli/aws/diag.py
@@ -36,10 +36,12 @@ NAME_DIAG_INSTANCE = 'Bracket Diag for snapshot %(snapshot_id)s'
 DESCRIPTION_DIAG_INSTANCE = \
     'Diag instance with logs from %(snapshot_id)s'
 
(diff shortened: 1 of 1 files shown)


brkt_cli/aws/diag.py
```diff
@@ -121,6 +123,6 @@ def diag(aws_svc=None, region='us-west-2',
         print "IP address: %s" % diag_instance.ip_address
     if diag_instance.private_ip_address:
         print "Private IP address: %s" % diag_instance.private_ip_address
-    print "User: ec2-user"
+    print "User: root"
     print "SSH Keypair: %s" % ssh_keypair
-    print "Log volume mountpoint: /dev/xbd5 for PV, /dev/xbd5s1 for HVM"
+    print "Log volume mountpoint: /dev/xbd2a for PV, /dev/xbd2e for HVM"
```
agtucker: LGTM

PR #358, opened 2016-07-19
Adds arg for updater instance type

dramdass:
I ran into an issue when updating an AMI on behalf of a customer
where AWS did not have the the capacity to launch c3.large
instances for more than an hour. As a workaround I had to hack
the brkt cli to use an m3.medium instead. We should add an
instance type arg for anywhere we launch an instance so we
aren't beholden to AWS capacity issues.

Makes the default updater type to m3.medium. The previous was
c3.large to match the encryptor instance type. However, we
don't need a beefy instance type for updating since we don't
do any resource intensive operations during updating.


brkt_cli/aws/update_encrypted_ami_args.py
```diff
@@ -35,7 +35,16 @@ def setup_update_encrypted_ami(parser):
...
+            'The instance type to use when running the encrypted guest '
+            'instance'),
+        default='m3.medium'
+    )
+    parser.add_argument(
+        '--updater-instance-type',
+        metavar='TYPE',
+        dest='updater_instance_type',
+        help=(
+            'The instance type to use when running the updater '
```
bburtin: Does the usage output automatically print the default for you?

bburtin: I was lazy.  Looks like it doesn't.  Can you document the default values for `--guest-instance-type` and `--updater-instance-type`?

bburtin:
Looks good.  For extra credit, you may want to squash the commits. :shipit:

dramdass:
going to push to a new branch with clean commit history

PR #360, opened 2016-07-20
Adds arg for updater instance type

dramdass:
I ran into an issue when updating an AMI on behalf of a customer
where AWS did not have the the capacity to launch c3.large
instances for more than an hour. As a workaround I had to hack
the brkt cli to use an m3.medium instead. We should add an
instance type arg for anywhere we launch an instance so we
aren't beholden to AWS capacity issues.

Makes the default updater type to m3.medium. The previous was
c3.large to match the encryptor instance type. However, we
don't need a beefy instance type for updating since we don't
do any resource intensive operations during updating.


dramdass:
Boris gave me a LGTM on this code in another branch. This just cleans up the commit history. Going to merge

PR #362, opened 2016-07-20
Lower case tag and value for token claims

robertabbott:
Because Env and env tags being different
would be confusing


brkt_cli/util.py
```diff
@@ -198,7 +198,7 @@ def parse_name_value(name_value):
     if not m:
         raise ValidationError(
             '%s is not in the format NAME=VALUE' % name_value)
-    return m.group(1), m.group(2)
+    return m.group(1).lower(), m.group(2).lower()
```
bburtin: This is a general-purpose function.  I think it would be more appropriate to perform the lower-casing at the callsite.

robertabbott: done

PR #363, opened 2016-07-21
Support specifying a domain for --brkt-env

bburtin:
Support specifying the Yeti environment by its domain

    Add a new hidden --service-domain option that specifies the Yeti
    environment that the metavisor (and soon, the CLI) will talk to.
    When a domain is specified (e.g. stage.mgmt.brkt.com), we infer the
    following hosts and ports:

    * internal API (RPC): yetiapi.<domain>:443
    * HSM proxy: hsmproxy.<domain>:443
    * external API: api.<domain>:80

    Always send the Yeti environment to metavisor, so that behavior is
    clearly defined and does not differ between CSPs.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -33,7 +33,6 @@ from brkt_cli.util import validate_dns_name_ip_address
 from brkt_cli.validation import ValidationError
 
 VERSION = '1.0.3pre1'
-BRKT_ENV_PROD = 'yetiapi.mgmt.brkt.com:443,hsmproxy.mgmt.brkt.com:443'
 
 # The list of modules that may be loaded.  Modules contain subcommands of
 # the brkt command and CSP-specific code.
@@ -50,19 +49,22 @@ log = logging.getLogger(__name__)
 
 
 class BracketEnvironment(object):
-    def __init__(self):
-        self.api_host = None
-        self.api_port = None
-        self.hsmproxy_host = None
-        self.hsmproxy_port = None
+    def __init__(self, api_host=None, api_port=443,
+                 hsmproxy_host=None, hsmproxy_port=443,
+                 public_api_host=None, public_api_port=80):
+        self.api_host = api_host
+        self.api_port = api_port
+        self.hsmproxy_host = hsmproxy_host
+        self.hsmproxy_port = hsmproxy_port
+        self.public_api_host = public_api_host
+        self.public_api_port = public_api_port
 
     def __repr__(self):
-        return '<BracketEnvironment api=%s:%d, hsmproxy=%s:%d>' % (
-            self.api_host,
-            self.api_port,
-            self.hsmproxy_host,
-            self.hsmproxy_port
-        )
+        return (
+            '<BracketEnvironment api={be.api_host}:{be.api_port} '
+            'hsmproxy={be.hsmproxy_host}:{be.hsmproxy_port} '
+            'public_api={be.public_api_host}:{be.public_api_port}>'
+        ).format(be=self)
 
 
 def validate_ntp_servers(ntp_servers):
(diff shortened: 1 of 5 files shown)


bonzotronix:
Changing the interpretation of "--brkt-env" could be OK, since there are (presumably) only a few customers using it.  But maybe it would be cleaner to leave that option and add a new "--brkt-domain" arg?  I'd consider it an error if both "--brkt-env" and "--brkt-domain" were provided.

bburtin:
Thanks Adam.  You're right.  It's better to have two options than to overload the format of one.  I added a new `\-\-service-domain` option (working title, please send suggestions).

brkt_cli/instance_config_args.py
```diff
@@ -63,15 +67,23 @@ def setup_instance_config_args(parser, mode=INSTANCE_CREATOR_MODE,
...
         default=brkt_env_default,
         help=argparse.SUPPRESS
     )
+    # Optional domain that runs the Yeti service (e.g. stage.mgmt.brkt.com).
+    # Hidden because it's only used for development.
+    parser.add_argument(
+        '--service-domain',
+        metavar='DOMAIN',
+        help=argparse.SUPPRESS
+    )
```
bonzotronix: When we add real help text, we should mention that the endpoints are assumed to use port 443. If customers need to use a different port, then they need to use --brkt-env instead.

bburtin: Yeah, good point.  I'm probably going to have to add per-endpoint config sooner rather than later.  I'll address this at that point.

brkt_cli/instance_config_args.py
```diff
@@ -63,15 +67,23 @@ def setup_instance_config_args(parser, mode=INSTANCE_CREATOR_MODE,
...
+    # <rpc-host>:<rpc-port>,<hsmproxy-host>:<hsmproxy-port>
     parser.add_argument(
         '--brkt-env',
         dest='brkt_env',
         default=brkt_env_default,
         help=argparse.SUPPRESS
     )
+    # Optional domain that runs the Yeti service (e.g. stage.mgmt.brkt.com).
+    # Hidden because it's only used for development.
+    parser.add_argument(
```
bonzotronix: Maybe use argparse.add_mutually_exclusive_group for --brkt-env, --service-domain.

brkt_cli/test_instance_config.py
```diff
@@ -61,8 +61,6 @@ def _get_ca_cert_filename():
 def _verify_proxy_config_in_userdata(ut, userdata):
     brkt_config_json = get_mime_part_payload(userdata,
                                              BRKT_CONFIG_CONTENT_TYPE)
-    ut.assertEqual(brkt_config_json, '{"brkt": {}}')
```
bonzotronix: Should we update the check to look for JSON containing the PROD endpoints?

brkt_cli/instance_config_args.py
bburtin: Thanks, I completely forgot about that!

bonzotronix:
As discussed unicast, we might want to use the name "public API" endpoint in place of "external API" endpoint.
FTP

brkt_cli/test_instance_config.py
bburtin: I believe that the other tests already check this case.  Let me know if I'm misunderstanding.

bonzotronix: Ahh, I see that now.  Sounds good.

bburtin:
Thanks Adam.  I changed the terminology from "external API" to "public API."

brkt_cli/__init__.py
```diff
@@ -127,6 +129,27 @@ def _parse_endpoint(endpoint):
     return be
 
 
+def brkt_env_from_domain(domain):
+    """ Return a BracketEnvironment object based on the given domain
+    (e.g. stage.mgmt.brkt.com).
+
+    :return: a BracketEnvironment object
+    :raise: ValidationError if brkt_env is malformed
```
bonzotronix: doesn't look like it'll raise this

bburtin: Thanks for catching this.  Fixed.

bonzotronix:
Code changes LGTM.  Assuming the arg naming discussion has converged, I think it's OK to merge.

PR #365, opened 2016-07-21
Add support for specifying options in a config file

mpage:
This commit adds infrastructure to allow subcommands to read default
values from a config file stored in ~/.brkt/config. In order to use
the new infrastructure, subcommands must implement the setup_config()
method. In it, they should register any options they want to pull
from the config.

A new sub-command, "config", has been added to the cli. This lets
users list, set, get, and unset options in the config. It is
modeled after "kubectl config".

Finally, the first consumer of the config infrastructure is the
"encrypt-gce-image" sub-command. The "project", "network", and
"zone" options may be specified via the config.

Sample usage:

    # Set the default project to use for encrypting images:
    brkt config set encrypt-gce-images.project myproject

    # Retrieve the default project for encrypting images:
    brkt config get encrypt-gce-images.project

    # Unset the default project for encrypting images:
    brkt config unset encrypt-gce-images.project

    # Display all options set in the config file:
    brkt config list


brkt_cli/__init__.py
```diff
@@ -313,15 +313,16 @@ def add_brkt_env_to_brkt_config(brkt_env, brkt_config):
         brkt_config['hsmproxy_host'] = hsmproxy_host_port
 
 
-class SortingHelpFormatter(argparse.HelpFormatter):
+class SortingHelpFormatter(argparse.ArgumentDefaultsHelpFormatter):
```
bburtin: Thanks, I didn't know about this help formatter.  We manually document default values for some of the options.  Can you update the docs so that we don't print the default twice in the usage output?  This can be in a separate commit.

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+                log.exception("Failed unlinking %s", path)
+        except:
+            log.exception("Failed unlinking %s", path)
+
+    def _write_config(self):
+        """Create ~/.brkt if it doesn't exist and safely write out a
+        new config.
+        """
+        try:
+            os.mkdir(CONFIG_DIR, 0755)
```
bburtin: Is there a reason not to use the default permissions here?  Not a huge deal, just that when I read this code, I wonder if the permissions are significant.

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+        except OSError as e:
+            if e.errno == errno.EEXIST:
```
bburtin: Would
```
if e.errno != errno.EEXIST:
    raise
```
be more succinct?

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+                pass
+            else:
+                raise
+        # Construct our own tempfile since NamedTemporaryFile doesn't
```
bburtin: I read the docs for NamedTemporaryFile.  Seems like it should work on Windows as long as you close the file before renaming it?

brkt_cli/util.py
```diff
@@ -217,3 +217,23 @@ def read_private_key(pem_path):
     if crypto.curve != brkt_cli.crypto.SECP384R1:
         raise ValidationError(key_format_err)
     return crypto
+
+
+def render_table_rows(rows, row_prefix=''):
+    """ Render the supplied rows as a table
```
bburtin: Can you provide a simple example in the docstring of how the output will look?

bburtin:
Looks good, aside from the nits that I brought up.  Does ConfigParser handle a list of values for the same key, or specifying the same key multiple times?  We'll need this soon for things like proxy settings.

brkt_cli/__init__.py
mpage: Will do.

brkt_cli/config/__init__.py
mpage: Default mode is 0777. That seems unnecessarily permissive (system-wide rw access).

brkt_cli/config/__init__.py
mpage: Yep. Fixed.

brkt_cli/config/__init__.py
mpage: The docs would lead one to believe that is the case, but Danlin had to rewrite a bunch of code in the metavisor because of NamedTemporaryFile. I'm not clear on the specifics, so it's possible that we were using it in a way that is known to not work. I'd rather avoid the issue entirely.

brkt_cli/util.py
mpage: Will do.

mpage:
> Does ConfigParser handle a list of values for the same key, or specifying the same key multiple times? 

AFAICT, no. Multiple values will need to be stored as a string (i.e. a comma delimited list).

brkt_cli/config/__init__.py
bburtin: I talked to Danlin and confirmed that the Windows problems occur when you try to rename a file that hasn't been closed.  Windows is known to be less flexible about sharing file descriptors than Unix is.  I would lean toward using the library function until we know that there's a problem that needs to be solved.  I'll make sure that the code is tested on Windows before we ship it, and I'll deal with any fallout that occurs.

mpage: I'm not entirely clear what's wrong with the current approach, but I'll defer to you. New diff coming soon.

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+
+VERSION = 1
+
+
+class InvalidOptionError(Exception):
+    def __init__(self, option):
+        self.option = option
+
+
+class CliConfig(object):
```
bburtin: Naming nit: the convention is for acronyms to be in caps (CLIConfig).

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+        attr = levels.pop()
+        cur = self._config['options']
+        for level in levels:
+            if level not in cur:
+                return
+            cur = cur[level]
+        if attr in cur:
+            del cur[attr]
+        # Clean up any empty sub-sections
+        cur = self._config['options']
```
bburtin: Doesn't this need to be done in decreasing order of depth?  Seems like this algorithm won't work with if you have a/b/c, where c is empty and b only contains c.

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+                break
+
+    def read(self):
+        """Read the config from disk"""
+        try:
+            with open(CONFIG_PATH) as f:
+                config = yaml.safe_load(f)
+            self._config = config
+        except IOError as e:
+            if e.errno != errno.ENOENT:
```
bburtin: I would either check for existence before opening or add a comment in the `except` block.  Unless I'm the only one who doesn't immediately know what ENOENT is.

brkt_cli/config/__init__.py
```diff
@@ -0,0 +1,294 @@
...
+    def name(self):
+        return 'config'
+
+    def register(self, subparsers, parsed_config):
+        self.parsed_config = parsed_config
+        config_parser = subparsers.add_parser(
+            self.name(),
+            description=(
+                'Display or update brkt-cli options stored in'
+                ' ~/.brkt/config.')
```
bburtin: Bordering on pedantic, but you may want to remove the trailing dot so there's no ambiguity about whether it's punctuation or part of the filename.  Your call.

brkt_cli/config/test_config.py
```diff
@@ -0,0 +1,88 @@
...
+        self.cmd.register(parser.add_subparsers(), self.cfg)
+        self.cmd._write_config = noop
+
+    def test_get_unknown_option(self):
+        """Verify that we raise an error if the user attempts to fetch an
+        unknown option.
+        """
+        options = ['no-section', 'test-section.no-option']
+        for option in options:
+            self.assertRaises(ValidationError, self.cmd.run, GetValues(option))
```
bburtin: The `with` format reads a bit better.  More of an FYI, if you don't want to deal with rewriting all of your `assertRaises()` calls:
```
with self.assertRaises(ValidationError):
    cmd.run(GetValues(option))
```

bburtin:
Thanks Matt, this looks good.  Can you paste the output out `brkt config -h` into the commit message?  Kind of hard to know what it will look like from reading the code.

brkt_cli/config/test_config.py
mpage: Cool!

brkt_cli/config/__init__.py
mpage: Pedantry, but checking before opening exposing us to a race condition. Also, maybe it's just because I'm a (sorta) systems person, but ENOENT is pretty widely known. Also, google is your friend.

brkt_cli/config/__init__.py
mpage: Will fix.

brkt_cli/config/__init__.py
mpage: Clarity is important. I'll remove the trailing dot.

brkt_cli/config/__init__.py
mpage: You're right. This only cleans up the bottom most empty section. WTF was I smoking.

bburtin: Finally, a legitimate reason to write recursive code! 😄

bburtin:
Looks good, thanks!

mpage:
Going to abandon this and open a new pull request with all the commits squashed.

PR #366, opened 2016-07-26
Change default encryptor/updater status port to 80

dramdass:
This is a more sensible default for customers who have ASAs or
other devices that block certain ports. The metavisor change
that supports custom status ports have been in for more than
a month and a half so this is a safe change.

Updating AMIs older than June 12th to later metavisor AMI will
just work.

If for any reason, someone has to encrypt using a metavisor AMI
older than June 12th, the --encryptor-status-port=8000 flag will
be needed.


bburtin:
👍

PR #369, opened 2016-07-27
Handle encryptor retrieval failure better

robertabbott:
Waits longer for the encryptor image to show
up. fails more gracefully
if the image isn't created for some reason

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -95,7 +95,7 @@ def create_image(gce_svc, zone, encrypted_image_disk, encrypted_image_name, encr
         gce_svc.wait_snapshot(encrypted_image_name)
         log.info("Image %s successfully created!", encrypted_image_name)
     except Exception as e:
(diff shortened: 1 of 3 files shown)


brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -106,8 +106,12 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
         # create metavisor image from file in GCS bucket
         log.info('Retrieving encryptor image from GCS bucket')
         if not encryptor_image:
-            encryptor_image = gce_svc.get_latest_encryptor_image(zone,
-                image_bucket, image_file=image_file)
+            try:
+                encryptor_image = gce_svc.get_latest_encryptor_image(zone,
+                    image_bucket, image_file=image_file)
+            except Exception as e:
+                log.info('Failed to retrieve encryptor image: %s', e)
```
bburtin: Do you need to log the exception here?  I would assume that the caller will log it.

dramdass:
LGTM

PR #370, opened 2016-07-27
Make default encryptor port 8000 again

dramdass:
Since the corresponding metavisor change to accept custom ports
in GCE isn't there. We'll make the default 80 again when the
metavisor change for GCE is there.


bburtin:
👍

jntkflow:
👍

PR #371, opened 2016-07-27
Validate JWT before attempting encryption

bburtin:
When the user specifies --token, check the JWT against the public JWK
endpoint before attempting encryption.  This causes the encryption
process to fail earlier and more intuitively when the public key is not
associated with a Yeti customer.

Fix a bug in BracketEnvironment initialization.  The default https port
is 443, not 80.


brkt_cli/__init__.py
```diff
@@ -318,6 +333,44 @@ def validate_jwt(jwt):
...
+    """ Authenticate with Yeti using the given JWT and make sure that the
+    associated public key is registered with the account.
+
+    :param brkt_env a BracketEnvironment object
+    :param jwt a JWT string
+    :raise ValidationError if the token fails auth or the public key is not
+    registered with the given account
+    """
+    kid = brkt_jwt.get_key_id(jwt)
+    uri = 'https://%s:%d/api/v1/jwk/%s' % (
```
gcallaghan: does the scheme need to be configurable?

gcallaghan:
LGTM

brkt_cli/__init__.py
bburtin: I believe that we require HTTPS.  Let me know if this is not the case.

PR #372, opened 2016-07-28
Remove the requests library

bburtin:
Remove the unused oauth_requests module and all dependencies on the
requests library.  Seems like urllib2 is good enough for what we're
doing, and one less dependency is a good thing.  Use Python logging
instead of print() inside _check_version(), so that we log in a uniform
way.


dramdass:
LGTM

PR #373, opened 2016-07-28
Load the list of supported regions

bburtin:
Load the list of supported regions from the encryptor AMI endpoint,
instead of hardcoding.


gcallaghan:
LGTM

PR #374, opened 2016-07-28
Set public_api_host correctly when brkt-env is used

robertabbott:


bburtin:
What is the problem that you're solving here?  Seems like a hack to make assumptions about the hostname that's passed to `--brkt-env`.

robertabbott:
https://github.com/brkt/brkt-cli/pull/371 assumes that public_api_host is set so brkt-env is completely broken by that change. This fixes that issue by making the same assumption about the public api prefix that service-domain does

PR #375, opened 2016-07-28
Don't require brkt-env when launching a GCE instance

bburtin:
Split BracketEnvironment parsing out from the rest of the values passed
when constructing InstanceConfig.  This allows callsites to determine
whether to specify BracketEnvironment or not.

Background: in commit bfe70c372f95730964a0f9c3366aa94a8806cf64 we
started always sending the Yeti environment to Metavisor.  This created
problems when launching a GCE image.  If the user didn't specify the
environment on the command line, we'd send the prod endpoints, which
would inadvertently override the Metavisor defaults.

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -6,7 +6,7 @@ from brkt_cli.subcommand import Subcommand
 from brkt_cli import encryptor_service, util
 from brkt_cli.instance_config import INSTANCE_METAVISOR_MODE
 from brkt_cli.instance_config_args import (
-    instance_config_from_values,
+    make_instance_config,
     setup_instance_config_args
 )
 from brkt_cli.gce import (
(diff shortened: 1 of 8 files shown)


brkt_cli/gce/__init__.py
```diff
@@ -97,7 +97,8 @@ def _run_subcommand(subcommand, values):
 
 def command_launch_gce_image(values, log):
     gce_svc = gce_service.GCEService(values.project, None, log)
-    instance_config = instance_config_from_values(values)
+    brkt_env = brkt_cli.brkt_env_from_values(values)
+    instance_config = make_instance_config(values, brkt_env)
```
bonzotronix: Not to get too hung up on function naming, but ISTM we'd either want to use make_instance_config() + make_brkt_env() or instance_config_from_values() + brkt_env_from_values().

robertabbott:
in its current form this change doesn't fix the problem

brkt_cli/gce/__init__.py
bburtin: BracketEnvironment can be built from either values or from a domain.  Since Python isn't typed, I can't have multiple `make_brkt_env()` functions that accept different types.  InstanceConfig has a single constructor function that now requires values + BracketEnvironment.  Calling it "from values" sounds like a misnomer.  I went back and forth and decided that this sucks the least.  If you think `instance_config_from_values()` is better, I can change it back.

bonzotronix: The InstanceConfig ctor takes a brkt_config dict.  So I thought instance_config_from_values() would be sensible.  No biggie either way.

PR #377, opened 2016-07-29
Validate the token before attempting to use it

bburtin:
Validate the user-specified token.  Make sure that it's not malformed
and has all of the fields that we require.  Replace manual JWT parsing
with calls to convenience methods in brkt_jwt.


gcallaghan:
LGTM

PR #378, opened 2016-07-29
Expose the make-key, make-token, and make-user-data commands

bburtin:
Expose these commands, now that we have back-end support for JWT auth.


krishnanrs:
LGTM

PR #379, opened 2016-08-01
Display GCE instance name and IP upon successful launch

krishnanrs:
Displays the GCE instance name and public IP which avoids the user from having to lookup the GCE UI (or use other means) to obtain the IP of the launched instance.


mpage:
LGTM

PR #380, opened 2016-08-01
Verify guest image before trying to encrypt it

robertabbott:


krishnanrs:
small nit. can we remove the duplicate error checking that is happening in the validate_images method? Along the same lines, the check for duplicate image name (also within the validate_images method) could be moved in with the new check so that these validations happen much early on.

robertabbott:
validate_image_name verifies that the name of the output image doesn't break gce's image naming rules whereas the new method verifies the guest image exists. I don't think its makes sense to put these two actions in the same method.

brkt_cli/gce/gce_service.py
```diff
@@ -594,6 +594,13 @@ def get_image_name(encrypted_image_name, name):
     return encrypted_image_name
 
 
+def validate_guest_image(gce_svc, image, project):
+    if not gce_svc.image_exists(image, project):
+        raise ValidationError(
+            "Invalid guest image or image project. Cannot encrypt"
```
bburtin: The error message is a bit ambiguous.  It's not clear whether the image doesn't exist, or if it exists and is incompatible with Metavisor.  I'd leave out "Cannot encrypt", since it doesn't add any value and makes assumptions about what the caller is trying to do.

Would it be better to make this a general-purpose function that checks if an image exists?  Or you could go the other way and be more thorough about checking the guest image to make sure that it's compatible.

robertabbott: "if not gce_svc.image_exists(image, project):"

The gce service already has a method to check if images exist. This lightly wraps the gce service method to serve this particular purpose

krishnanrs:
@robertabbott in that case, the first and last check in the below method can be removed since they are now being done twice:

def validate_images(gce_svc, encrypted_image_name, encryptor, guest_image, image_project=None):
    # check that image to be updated exists
    if not gce_svc.image_exists(guest_image, image_project):
        raise ValidationError('Image %s does not exist. Cannot encrypt' % guest_image)

    # check that encryptor exists
    if not gce_svc.image_exists(encryptor):
        raise ValidationError('Encryptor image %s does not exist. Encryption failed.' % encryptor)

    # check that there is no existing image named encrypted_image_name
    if gce_svc.image_exists(encrypted_image_name):
        raise ValidationError('An image already exists with name %s. Encryption Failed.' % encrypted_image_name)

robertabbott:
That makes sense @krishnanrs . Done

bburtin:
Looks good to me.  I'll let @krishnanrs give the +1.

brkt_cli/gce/__init__.py
```diff
@@ -176,6 +181,11 @@ def command_encrypt_gce_image(values, log):
 
     encrypted_image_name = gce_service.get_image_name(values.encrypted_image_name, values.image)
     gce_service.validate_image_name(encrypted_image_name)
+    gce_service.validate_images(gce_svc,
```
krishnanrs: @robertabbott my suggestion was not to move out the validate_images method completely but simply to remove the first check within this method as you had moved it up the chain (and into the __init__.py file). However the new change LGTM so long as the tests pass.

PR #381, opened 2016-08-01
Add support for specifying options in a config file

mpage:
This commit adds infrastructure to allow subcommands to read default
values from a config file stored in ~/.brkt/config. In order to use
the new infrastructure, subcommands must implement the setup_config()
method. In it, they should register any options they want to pull
from the config.

A new sub-command, "config", has been added to the cli. This lets
users list, set, get, and unset options in the config. It is
modeled after "kubectl config".

Finally, the first consumer of the config infrastructure is the
"encrypt-gce-image" sub-command. The "project", "network", and
"zone" options may be specified via the config.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -24,10 +24,8 @@ import urllib2
 from distutils.version import LooseVersion
 from operator import attrgetter
 
-from brkt_cli import (
-    brkt_jwt,
-    util
-)
+from brkt_cli import brkt_jwt, util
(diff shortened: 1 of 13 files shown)


bburtin:
Oh yeah, I just remembered that you'll need to add your new package to `packages` in `setup.py`.  Otherwise setuptools won't install it.

bburtin:
👍

bburtin:
I bet jjv will give you a big hug for this commit.

PR #382, opened 2016-08-01
Dump gce console on encryptor failure

robertabbott:
Testing: killed encryption and verified
the console was written to a file

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -56,10 +56,15 @@ def do_encryption(gce_svc,
                                 gce_svc.get_disk(zone, encrypted_image_disk)],
                          metadata=metadata)
 
-    enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
-
-    wait_for_encryptor_up(enc_svc, Deadline(600))
-    wait_for_encryption(enc_svc)
+    try:
+        enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
+        wait_for_encryptor_up(enc_svc, Deadline(600))
+        wait_for_encryption(enc_svc)
+    except Exception as e:
+        f = gce_svc.write_serial_console_file(zone, encryptor)
+        if f:
+            log.info('Encryption failed. Writing console to %s' % f)
+        raise e
     retry(function=gce_svc.delete_instance,
             on=[httplib.BadStatusLine, socket.error, errors.HttpError])(zone, encryptor)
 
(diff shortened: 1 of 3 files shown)


brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -56,10 +56,15 @@ def do_encryption(gce_svc,
...
-    wait_for_encryptor_up(enc_svc, Deadline(600))
-    wait_for_encryption(enc_svc)
+    try:
+        enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
+        wait_for_encryptor_up(enc_svc, Deadline(600))
+        wait_for_encryption(enc_svc)
+    except Exception as e:
+        log.info('Encryption failed. Writing console to serial_console-%s.out'
+                % gce_svc.get_session_id())
+        gce_svc.write_serial_console_file(zone, encryptor)
```
bburtin: Looks like you'll lose the original exception if this function call raises an exception.

brkt_cli/gce/gce_service.py
```diff
@@ -307,6 +307,17 @@ def get_instance_ip(self, name, zone):
                 pass
         self.log.info("Couldn't find an IP address for this instance.")
 
+    def write_serial_console_file(self, zone, instance):
+        serial_port_out = self.compute.instances().getSerialPortOutput(
+                project=self.project,
+                instance=instance,
+                zone=zone).execute()
+        contents = serial_port_out['contents']
+        fname = 'serial_console-%s.out' % self.session_id
```
bburtin: Should this be a temp file instead of being written to the current directory?  I don't think it's safe to assume that the current directory is writable.

robertabbott: done

brkt_cli/gce/encrypt_gce_image.py
robertabbott: fixed

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -56,10 +56,15 @@ def do_encryption(gce_svc,
...
-    enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
-
-    wait_for_encryptor_up(enc_svc, Deadline(600))
-    wait_for_encryption(enc_svc)
+    try:
+        enc_svc = enc_svc_cls([gce_svc.get_instance_ip(encryptor, zone)])
+        wait_for_encryptor_up(enc_svc, Deadline(600))
+        wait_for_encryption(enc_svc)
+    except Exception as e:
+        log.info('Encryption failed. Writing console to serial-console-%s.out'
```
bburtin: You should print the full path to the file.  Chances are that it's several levels deep under `/tmp`.

PR #383, opened 2016-08-02
Print unexpected failure code

bburtin:
We're improving error handling in the Metavisor.  It's possible that
Metavisor will send an error code that the CLI isn't expecting.  If this
happens, print the failure code, so that the user has an idea of what
went wrong.  Also, print the entire status structure when running in
verbose mode.


dramdass:
Nice! 👍

PR #384, opened 2016-08-02
Remove support for the cnc claim

bburtin:
Remove the cnc option from the make-token command.  We are no longer
using this claim to limit the number of instances.


gcallaghan:
LGTM

PR #385, opened 2016-08-02
Stop encryptor before getting console log

bburtin:
When encryption or update fails, stop the metavisor instance before
attempting to collect the console log.  The current behavior in AWS is
that the console log isn't immediately available unless the instance
state changes.

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -194,6 +194,18 @@ def wait_for_instance(
     )
 
 
+def stop_and_wait(aws_svc, instance_id):
+    """ Stop the given instance and wait for it to be in the stopped state.
+    If an exception is thrown, log the error and return.
+    """
+    try:
+        aws_svc.stop_instance(instance_id)
+        wait_for_instance(aws_svc, instance_id, state='stopped')
+    except:
+        log.exception(
+            'Error while waiting for instance %s to stop', instance_id)
+
+
 def get_encrypted_suffix():
     """ Return a suffix that will be appended to the encrypted image name.
     The suffix is in the format "(encrypted 787ace7a)".  The nonce portion of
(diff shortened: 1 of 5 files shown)


brkt_cli/aws/encrypt_ami.py
```diff
@@ -674,6 +674,14 @@ def snapshot_encrypted_instance(aws_svc, enc_svc_cls, encryptor_instance,
         log.info('Creating encrypted root drive.')
         encryptor_service.wait_for_encryption(enc_svc)
     except (BracketError, encryptor_service.EncryptionError) as e:
+        # Stop the encryptor instance, to make the console log available.
+        try:
+            aws_svc.stop_instance(encryptor_instance.id)
+            wait_for_instance(aws_svc, encryptor_instance.id, state='stopped')
+            pass
```
agtucker: Is the "pass" doing anything here? I don't see it in the analogous update_ami code.

bburtin: It's debugging cruft.  Thanks for catching it.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -674,6 +674,13 @@ def snapshot_encrypted_instance(aws_svc, enc_svc_cls, encryptor_instance,
         log.info('Creating encrypted root drive.')
         encryptor_service.wait_for_encryption(enc_svc)
     except (BracketError, encryptor_service.EncryptionError) as e:
+        # Stop the encryptor instance, to make the console log available.
+        try:
+            aws_svc.stop_instance(encryptor_instance.id)
+            wait_for_instance(aws_svc, encryptor_instance.id, state='stopped')
+        except:
```
agtucker: I think this is going to "clobber" the current exception state - i.e. further down (line 694) you should have "raise e" instead of just "raise" so that the original exception is preserved. Same for update_ami.

bburtin: OMG Python is evil.

agtucker:
LGTM

PR #386, opened 2016-08-03
Clean up JWT validation

bburtin:
Validate the JWT against /api/v1/customer/self instead of
/api/v1/jwk/{kid}.  Turns out that the JWK endpoint may not be there if
the Yeti is using a 3rd party JWK server.

Make the code more forgiving about unexpected errors.  We don't want to
interrupt the encryption process if we're unable to validate the token.
If an IOError is raised or the server returns an unexpected error
code, continue the process and let Metavisor validate the token during
encryption.


krishnanrs:
👍

bonzotronix:
LGTM

PR #387, opened 2016-08-03
Remove support for BRACKET_ENVIRONMENT

bburtin:
Remove support for the BRACKET_ENVIRONMENT environment variable.  This
variable was a switch that would tell brkt-cli whether to use the
staging or production build of Metavisor.  This is no longer being used.
Developers who want to use a specific Metavisor AMI can use the hidden
--encryptor-ami option.


krishnanrs:
👍

PR #388, opened 2016-08-03
Make port 80 default status port for GCE and AWS

dramdass:
This adds --status-port for GCE in addition to AWS and makes
the default port 80.

Testing done:
- encrypted AWS with and without --status-port
- encrypted GCE with and without --status-port


robertabbott:
LGTM

bburtin:
I believe that moving the argument to `instance_config_args` will cause it to show up in `launch-gce-image`.  We should talk to Bobby about this.  I'm working on `--service-domain`, which has the same issue.

robertabbott:
launch-gce-image doesn't use the status port for anything and the status port isn't put in userdata so the launch-gce-image command will just ignore the value and it shouldn't be an issue.

The reason the service-domain thing is an issue is that we put the value in userdata and the metavisor decides who to talk to based on the contents of userdata rather than the file in the image

bburtin:
Yes, these are different cases, but there's still a supportability concern.  It's misleading to have an option for a command, which is documented but ignored.

dramdass:
I'll make it only for encrypt-ami, update-encrypted-ami, encrypt-gce-image, update-gce-image before I merge

dramdass:
Updated to address comment from @bburtin. Now it is not part of `brkt launch-gce-instance` - please take a look

bburtin:
👍  Thanks Dennis.

PR #389, opened 2016-08-03
Don't read --brkt-env inside make_instance_config()

bburtin:
The BracketEnvironment object is now passed in by the caller.  Don't
attempt to read it from values when building the InstanceConfig object.

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -145,10 +145,12 @@ def make_instance_config(values=None, brkt_env=None,
             raise ValidationError(
                 'Can only specify ca-cert for instance in Creator mode'
             )
-        if not values.brkt_env:
+        if not brkt_env:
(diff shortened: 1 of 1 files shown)


brkt_cli/instance_config_args.py
```diff
@@ -145,10 +145,12 @@ def make_instance_config(values=None, brkt_env=None,
             raise ValidationError(
                 'Can only specify ca-cert for instance in Creator mode'
             )
-        if not values.brkt_env:
+        if not brkt_env:
```
bonzotronix: I don't think this will do what we want.  Looks like all the callers of make_instance_config() create the brkt_env first, so it'll never be None here.  We want to throw an error here if the user has chosen PROD endpoints.  So I guess the right check is to compare {code}brkt_env != get_prod_brkt_env(){code}

PR #390, opened 2016-08-04
Don't require project/zone if provided via config

mpage:


dramdass:
LGTM

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -14,12 +14,17 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         help='Specify the name of the generated encrypted Image',
         required=False
     )
+    zone_kwargs = {
```
bburtin: More of a style thing, but wouldn't it read better to break `default` and `required` out into separate variables?  The kwargs thing seems a little indirect.

mpage: I'm fine with how this is. Let me know if you want me to change it.

bburtin:
👍

PR #391, opened 2016-08-04
Remove image project from GCE image update validation

krishnanrs:


bburtin:
Can you update the commit with a short description of why you're making this change?

PR #392, opened 2016-08-04
Add network_host to brkt env passed in

dramdass:
This is needed for metavisor services talking to the Yeti
network service.

Testing done:
- make test


dramdass:
I'm going to hold off on this for a day or so until I make sure everything that uses --brkt-env currently is not going to break

PR #393, opened 2016-08-04
Handle new encryption/update failure codes

bburtin:
Handle new new failure codes returned during encryption or update.
Print a user-friendly message when they occur.  Move failure code
handling into a separate function, to keep wait_for_encryption()
concise.


dramdass:
LGTM!

PR #394, opened 2016-08-04
Change "S3" to "Cloud Storage" in status message

bonzotronix:
We can download encrypted guest images from either AWS S3 or Google Cloud Storage (GCS).  So make the status message generic.

### DIFF
--- a/brkt_cli/encryptor_service.py
+++ b/brkt_cli/encryptor_service.py
@@ -198,7 +198,7 @@ def wait_for_encryption(enc_svc,
             else:
                 state_display = 'Encryption'
                 if state == ENCRYPT_DOWNLOADING:
-                    state_display = 'Download from S3'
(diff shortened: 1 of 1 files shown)


brkt_cli/encryptor_service.py
```diff
@@ -198,7 +198,7 @@ def wait_for_encryption(enc_svc,
             else:
                 state_display = 'Encryption'
                 if state == ENCRYPT_DOWNLOADING:
-                    state_display = 'Download from S3'
+                    state_display = 'Download from Cloud Storage'
```
bburtin: "Cloud Storage" probably shouldn't be capitalized, since it isn't a proper noun.

bonzotronix: Fair enough.  Fixed in force-pushed commit.

bburtin:
👍

PR #395, opened 2016-08-05
Clean up InstanceConfig parsing and arguments

bburtin:
Move BracketEnvironment handling back into make_instance_config() and
rename it to instance_config_from_values().  The behavior was becoming
confusing because BracketEnvironment was handled at the callsite, but
other config stuff was parsed from values.

The current behavior is a bit funky because the logic switches on the
metavisor mode.  I think it's better than the alternative, since
callsites don't have to worry about explicitly handling every option
that goes into InstanceConfig.

Move the add_argument() calls for --brkt-env and --service-domain inside
the mode check.  We don't want to allow these options when launching a
GCE image.  When launching, Metavisor should talk to the Yeti
environment that was specified at encryption time.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import json
 import logging
-import os
 import re
 import urllib2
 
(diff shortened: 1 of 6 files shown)


bonzotronix:
LGTM.  Nicely done.  Great comments too.

PR #397, opened 2016-08-05
Log status when encryption fails

bburtin:
When encryption or update fails, log the entire data structure returned
by Metavisor at error level.  This gives us more context about the
failure, even if the user isn't running with -v.


brkt_cli/encryptor_service.py
```diff
@@ -247,7 +247,7 @@ def wait_for_encryption(enc_svc,
             log.info('Encrypted root drive created.')
             return
         elif state == ENCRYPT_FAILED:
-            log.debug('Encryption failed with status %s', status)
+            log.error('Encryption status: %s', json.dumps(status))
```
agtucker: LGTM

PR #398, opened 2016-08-05
Abort if the token is invalid

bburtin:
If the token auth check returns 400, log the message returned by the
server.  This can happen when Yeti uses a 3rd party JWK server and the
token does not have a customer claim.


bonzotronix:
Seems like an improvement.  So LGTM, I guess.

Though right now I'm more concerned that there is no way to disable this checking for GCE.  I see that we currently have no GCE equivalent to the 'validate' flag for the AWS encrypt/update commands.  Since we don't support a custom CA cert (for brkt-cli talking to YetI), this means that all GCE encrypt/update commands will fail.

gcallaghan:
LGTM

PR #399, opened 2016-08-05
Add network flag to GCE launch image

krishnanrs:
Provide an option to specify a network when launching GCE images via the CLI tool. Defaults to the 'default' network.


robertabbott:
This look fine. In general i don't want to enter the world of proxying instance options with the launch-gce-image command but I get why this is useful for testing so I think its cool to merge away

PR #401, opened 2016-08-05
Fix typo in version check

bburtin:
Clean up cruft that was introduced in the move from the Requests library
to urllib2.


bburtin:
Pyflakes found a bunch of other problems.  Will fix this as part of a bigger commit.

PR #404, opened 2016-08-08
GCE: log the host and port of the encryptor/updater service

bburtin:
Log the host and port that we use to monitor encryption/update progress,
to help the user diagnose networking issues.


robertabbott:
lgtm

PR #409, opened 2016-08-09
Adds gce no-validate option

robertabbott:
Images, networks and JWTs are not
validated when this option is used

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image_args.py
+++ b/brkt_cli/gce/encrypt_gce_image_args.py
@@ -11,7 +11,7 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         '--encrypted-image-name',
         metavar='NAME',
         dest='encrypted_image_name',
-        help='Specify the name of the generated encrypted Image',
+        help='Specify the name of the generated encrypted image',
         required=False
     )
     zone_kwargs = {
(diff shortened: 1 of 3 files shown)


brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -33,6 +33,13 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         default='prod',
         required=False
     )
+    parser.add_argument(
+        '--no-validate',
+        dest='validate',
+        action='store_false',
+        default=True,
+        help="Don't validate Images or JWTs"
```
bburtin: Image isn't a proper noun and shouldn't be capitalized.  We've been using "token" instead of JWT in other commands, so probably better to call it that.

brkt_cli/gce/update_encrypted_gce_image_args.py
```diff
@@ -35,6 +35,13 @@ def setup_update_gce_image_args(parser):
         required=True
     )
     parser.add_argument(
+        '--no-validate',
+        dest='validate',
+        action='store_false',
+        default=True,
+        help="Don't validate AMIs, subnet, and security groups"
```
bburtin: Copy/paste of AWS terms?

bonzotronix:
LGTM.  Thanks!

PR #410, opened 2016-08-09
Set default instance state timeout to 10 minutes

bburtin:
Set the default timeout in wait_for_instance() to 10 minutes, to avoid
intermittent failures when running integration tests.  Clean up
docstring.


dramdass:
👍

PR #412, opened 2016-08-11
launch-gce-image no longer appends 'guest'

robertabbott:
There wasn't really a reason to append 'guest'
to the instance name. Now users have more
flexibility in instance naming


dramdass:
LGTM

PR #413, opened 2016-08-11
Don't append guest to instance name

robertabbott:
launch-gce-image no longer append 'guest'
to the instance name. However because the
root disk and guest disk cant have the same
name '-s' is appended to instance_name and
used as the name for the disk created from
the encrypted guest root snapshot

the launch command now validates the instance
name before trying to launch the instance

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -139,6 +139,9 @@ def command_launch_gce_image(values, log):
     if not values.verbose:
         logging.getLogger('googleapiclient').setLevel(logging.ERROR)
 
+    if values.instance_name:
+        gce_service.validate_image_name(values.instance_name)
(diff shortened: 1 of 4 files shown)


brkt_cli/gce/__init__.py
```diff
@@ -133,6 +133,9 @@ def command_launch_gce_image(values, log):
     if not values.verbose:
         logging.getLogger('googleapiclient').setLevel(logging.ERROR)
 
+    if values.instance_name:
+        gce_service.validate_image_name(values.instance_name + '-s')
```
bburtin: Wouldn't this cause confusion if someone passes an image length that's 62 characters long and we say "Image name may be at most 63 characters"?

bburtin:
👍

PR #414, opened 2016-08-11
Use posixpath for dest files in userdata

bonzotronix:
When adding files with brkt-files parts in userdata, we must use Unix-style path separators for the destination file names, even when the brkt-cli command is being run on Windows.

Testing:
* Ran unit tests on Linux (no failures)
* Ran unit tests on Windows; this fix resolves 3 failures in existing
  tests (still need to fix temp file issue with
  test_util.TestReadPrivateKey.test_read_private_key()

### DIFF
--- a/brkt_cli/instance_config.py
+++ b/brkt_cli/instance_config.py
@@ -14,7 +14,7 @@
 
 import json
 import logging
-import os
+import posixpath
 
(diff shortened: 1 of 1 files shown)


sondring:
LGTM!

bburtin:
:shipit:

PR #419, opened 2016-08-13
Add method to validate if GCE project exists

krishnanrs:
Provide a more meaningful error if the GCE project entered is invalid. This was tested locally.


robertabbott:
lgtm

mpage:
👍

PR #420, opened 2016-08-15
Refactor AWS and GCE subcommands

mpage:
Simplify the run() logic by having the method body contained within each class.


bburtin:
👍

PR #421, opened 2016-08-15
Remove default value for update-encrypted-ami --region

bburtin:
We shouldn't assume that the user wants to run in a given region.


krishnanrs:
👍

PR #422, opened 2016-08-15
Add support for user-specified environments stored in config

mpage:
This commit makes environment a first class concept in the CLI. Environments are
defined as the set of Yeti backends that the CLI/MV talk to. Users declare new
environments via the `set-env` config subcommand. For example, to define the
`foo.com` environment, one does:

    > brkt config set-env foo --service-domain foo.com

Use the `use-env` config subcommand to switch between environments:

    > brkt config use-env stage

Use the `list-envs` config subcommand to list all the declared environments. A
leading `*` indicates the current environment.

    > ./brkt config list-envs
      brkt-hosted
    * foo
    >

The `brkt-hosted` environment corresponds to our hosted production environment.
It is the default environment and may not be changed by users.

Use the `get-env` config subcommand to get the details of an environment:

    > ./brkt config get-env foo
    api-server=yetiapi.foo.com:443
    key-server=hsmproxy.foo.com:443
    network-server=network.foo.com:443
    public-api-server=api.foo.com:443
    >

Finally, use the `unset-env` config subcommand to delete an environment:

    > ./brkt config unset-env foo

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -117,26 +117,23 @@ def parse_brkt_env(brkt_env_string):
     if len(endpoints) != 3:
         raise ValidationError(error_msg)
 
-    def _parse_endpoint(endpoint):
-        host_port_pattern = r'([^:]+):(\d+)$'
-        m = re.match(host_port_pattern, endpoint)
-        if not m:
-            raise ValidationError(error_msg)
-        host = m.group(1)
-        port = int(m.group(2))
-
-        if not util.validate_dns_name_ip_address(host):
-            raise ValidationError('Invalid hostname: ' + host)
-        return host, port
-
     be = BracketEnvironment()
-    (be.api_host, be.api_port) = _parse_endpoint(endpoints[0])
-    # set public api host based on the same prefix assumption
-    # service-domain makes. Hopefully we'll remove brkt-env
-    # soon and we can get rid of it
-    be.public_api_host = be.api_host.replace('yetiapi', 'api')
-    (be.hsmproxy_host, be.hsmproxy_port) = _parse_endpoint(endpoints[1])
-    (be.network_host, be.network_port) = _parse_endpoint(endpoints[2])
+    names = ('api', 'hsmproxy', 'network')
+    for name, endpoint in zip(names, endpoints):
+        try:
+            parts = util.parse_endpoint(endpoint)
+            if 'port' not in parts:
+                raise ValidationError(error_msg)
+            setattr(be, name + '_host', parts['host'])
+            setattr(be, name + '_port', parts['port'])
+            if name == 'api':
+                # set public api host based on the same prefix assumption
+                # service-domain makes. Hopefully we'll remove brkt-env
+                # soon and we can get rid of it
+                be.public_api_host = be.api_host.replace('yetiapi', 'api')
+        except ValueError:
+            raise ValidationError(error_msg)
+
     return be
 
 
@@ -162,7 +159,7 @@ def get_prod_brkt_env():
 def brkt_env_from_values(values):
     """ Return a BracketEnvironment object based on options specified
     on the command line.  If the environment was not specified with
-    --service-domain or --brkt-env, return the production environment.
+    --service-domain or --brkt-env, return None.
     """
     if values.service_domain:
         return brkt_env_from_domain(values.service_domain)
--- a/brkt_cli/config/__init__.py
+++ b/brkt_cli/config/__init__.py
@@ -82,10 +82,13 @@ def _bracket_environment_from_dict(d):
     :return a BracketEnvironment object
     """
     benv = brkt_cli.BracketEnvironment()
-    key_attr = zip(
-        ('api', 'keyserver', 'public-api', 'network'),
-        ('api', 'hsmproxy', 'public_api', 'network'))
-    for k, attr in key_attr:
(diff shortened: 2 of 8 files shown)


brkt_cli/__init__.py
```diff
@@ -117,26 +117,23 @@ def parse_brkt_env(brkt_env_string):
...
+            parts = util.parse_endpoint(endpoint)
+            if 'port' not in parts:
+                raise ValidationError(error_msg)
+            setattr(be, name + '_host', parts['host'])
+            setattr(be, name + '_port', parts['port'])
+            if name == 'api':
+                # set public api host based on the same prefix assumption
+                # service-domain makes. Hopefully we'll remove brkt-env
+                # soon and we can get rid of it
+                be.public_api_host = be.api_host.replace('yetiapi', 'api')
```
bburtin: Any thoughts on how long we're we going to keep `brkt-env` around?  I was hoping that we'd get rid of it sooner rather than later.  If it's not going away any time soon, we should get rid of the `yetiapi` search/replace hack.

brkt_cli/util.py
```diff
@@ -254,3 +255,30 @@ def render_table_rows(rows, row_prefix=''):
...
     table = "\n".join(lines)
     return table
+
+
+def parse_endpoint(endpoint):
+    """Parse a <host>[:<port>] string into its constituent parts.
+
+    :param endpoint a string of the form <host>[:<port>]
+
+    :return a dictionary of the form {"hostname": <hostname>, "port": <port>}
```
bburtin: More of a style thing, but a dictionary sounds gratuitous if you're only returning two things.

brkt_cli/config/__init__.py
```diff
@@ -21,35 +21,169 @@
...
+    """
+    benv = brkt_cli.BracketEnvironment()
+    key_attr = zip(
+        ('api', 'keyserver', 'public-api', 'network'),
+        ('api', 'hsmproxy', 'public_api', 'network'))
+    for k, attr in key_attr:
+        for suff in ('host', 'port'):
+            fk = k + '-' + suff
+            if fk in d:
+                setattr(benv, attr + '_' + suff, d[fk])
```
bburtin: This looks like it will raise a `KeyError` if the config file is malformed.  Is this what we want, given that it's technically user-editable?

brkt_cli/config/__init__.py
```diff
@@ -21,35 +21,169 @@
...
+def _bracket_environment_from_dict(d):
+    """Convert a bracket environment from the config into a BracketEnvironment
+    object
+
+    :param d a dictionary
+
+    :return a BracketEnvironment object
+    """
+    benv = brkt_cli.BracketEnvironment()
+    key_attr = zip(
```
bburtin: Would a dictionary be more intuitive here?
`key_attr = {'api': 'api', 'keyserver': 'hsmproxy', ...}`

brkt_cli/config/__init__.py
```diff
@@ -214,6 +363,95 @@ def register(self, subparsers, parsed_config):
             'option',
             help='The option name (e.g. encrypt-gce-image.project)')
 
+        # Define or update an environment
+        set_env_parser = config_subparsers.add_parser(
+            'set-env',
+            help='update the attributes of an environment',
```
bburtin: I've been starting help strings with a capital letter.  Not sure if that's right or wrong, but starting with lower case would go against the current convention.

brkt_cli/config/__init__.py
```diff
@@ -214,6 +363,95 @@ def register(self, subparsers, parsed_config):
...
+            help='the key server (host[:port]) the metavisor will connect to')
+        set_env_parser.add_argument(
+            '--network-server',
+            help='the network server (host[:port]) the metavisor will connect to')
+        set_env_parser.add_argument(
+            '--public-api-server',
+            help='the public api (host[:port])')
+        set_env_parser.add_argument(
+            '--service-domain',
+            help='set server values from the service domain')
```
bburtin: Should the `--service-domain` docs be more explicit?  Users will want to know what the hosts and ports will be set to when you use this option.  Maybe add more details to the top-level command description?

brkt_cli/config/__init__.py
mpage: I don't think it does. We check if `fk` is in `d` on line 91.

brkt_cli/config/__init__.py
mpage: Sure. This feels like a style change that isn't super important given how short the function is, though.

brkt_cli/config/__init__.py
bburtin: My bad.  This is what happens when you assume...

brkt_cli/config/__init__.py
mpage: Will stay consistent.

brkt_cli/__init__.py
mpage: This warrants a separate discussion outside of this review.

brkt_cli/config/__init__.py
mpage: Actually, should `--service-domain` be hidden? It's hidden elsewhere.

bburtin:
Looks good.  Let's get together to hash out whether we want to expose `--service-domain`, `--brkt-env`, etc.

PR #423, opened 2016-08-16
Copy the IOPS parameter from the right place

krishnanrs:
If the guest root volume is PIOPS, then the IOPS parameter needs to be copied over to the encrypted AMI


brkt_cli/aws/encrypt_ami.py
```diff
@@ -520,7 +520,7 @@ def _snapshot_root_volume(aws_svc, instance, image_id):
         raise
 
     ret_values = (
-        snapshot.id, root_dev, vol.size, vol.type, root_vol.iops)
```
agtucker: LGTM

PR #427, opened 2016-08-16
Copy IOPS properly for PIOPS volumes

krishnanrs:
Copy the IOPS properly for PIOPS volumes. For other volume types, ignore the IOPS settings.
Tested this locally with AMIs using standard, GP2 and PIOPS volumes.


dramdass:
LGTM

PR #428, opened 2016-08-17
Clean up brkt_env defaulting

bonzotronix:
* Delete BRKT_ENV_PROD as it is no longer used.
* Remove the default_brkt_env arg to setup_instance_config_args() as it is no longer used.

Testing:
* Ran unit tests
* Ran 'brkt encrypt-ami'


bburtin:
👍  Thanks for cleaning this up.

PR #429, opened 2016-08-17
Add a hidden option for keeping the encryptor running after failure

bburtin:
Add a hidden --no-terminate-encryptor-on-failure option.  When this
option is specified and encryption fails, we terminate the guest
instance but not the encryptor.  This allows a developer to connect to
the encryptor instance to diagnose the problem.

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -974,7 +974,8 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
             encrypted_ami_name=None, subnet_id=None, security_group_ids=None,
             guest_instance_type='m3.medium', instance_config=None,
             save_encryptor_logs=True,
-            status_port=encryptor_service.ENCRYPTOR_STATUS_PORT):
+            status_port=encryptor_service.ENCRYPTOR_STATUS_PORT,
+            terminate_encryptor_on_failure=True):
     log.info('Starting encryptor session %s', aws_svc.session_id)
 
     encryptor_instance = None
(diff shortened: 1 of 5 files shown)


brkt_cli/aws/encrypt_ami.py
```diff
@@ -1065,26 +1066,27 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
         instance_ids = []
         if guest_instance:
             instance_ids.append(guest_instance.id)
-        if encryptor_instance:
+        if encryptor_instance and terminate_encryptor_on_failure:
             instance_ids.append(encryptor_instance.id)
```
agtucker: Maybe print something like "preserving encryptor instance"?

brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -124,10 +124,21 @@ def setup_encrypt_ami_args(parser):
...
         dest='save_encryptor_logs',
         action='store_true',
         help=argparse.SUPPRESS,
         default=False
     )
+
+    # Optional argument for development: if encryption fails, keep the
+    # encryptor running so that we can debug it.
+    parser.add_argument(
+        '--no-terminate-encryptor-on-failure',
```
agtucker: Just noticed that the GCE code has a --keep-encryptor option that does the same thing - use the same argument name?

bburtin: I think it's different.  From looking at the code, it looks like that flag controls whether we delete the metavisor image in the GCP project, not the encryptor instance.

brkt_cli/aws/encrypt_ami.py
bburtin: Good idea.

brkt_cli/aws/encrypt_ami_args.py
agtucker: Yeah, I think you're right. Then what about GCE implementation of the "no terminate" functionality? Should I file a separate bug for that?

bburtin: Yeah, I guess so.  I know the AWS code a lot better, so I'm able to bang this out pretty quickly.  Go ahead and file the GCE bug and either Bobby or I will look at it separately.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -1065,26 +1066,39 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
         instance_ids = []
         if guest_instance:
             instance_ids.append(guest_instance.id)
-        if encryptor_instance:
+
+        terminate_encryptor = (
+            ami or
+            (encryptor_instance and terminate_encryptor_on_failure)
```
agtucker: Should this be:
    (encryptor_instance and (ami or terminate_encryptor_on_failure))
I.e. don't terminate the encryptor unless there is one. Though I guess ami can't be set unless encryptor_instance is also set, so it works out to be the same. LGTM either way.

bburtin: Yeah, reads better, given that the logic acts as a null check.  I'll fix it.

PR #435, opened 2016-08-22
Make make-user-data useful

bonzotronix:
* Change make_user_data so it does NOT put the part-handler
  Python script in the userdata output (since nobody is using this functionality).
* Omit Yeti endpoints in brkt-config output by make-user-data.
* Added a hidden --brkt-files arg to make-user-data, to facilitate adding files to /var/brkt/ami_config/ via userdata.
* Added unit tests for the above.
* Fixed test_parse_timestamps to use timestamps 10s in the future.

Testing:
* Ran unit tests.
* Ran MV instance with userdata containing a brkt-config blob of '{"brkt": {}}', to make sure the identity server didn't barf.


brkt_cli/make_user_data/__init__.py
```diff
@@ -47,13 +62,26 @@ def register(self, subparsers, parsed_config):
             action='store_true',
             help='Print status information to the console'
         )
+        parser.add_argument(
+            '--brkt-files',
+            metavar='FILENAME',
+            dest='make_user_data_brkt_files',
+            nargs='+',
```
bburtin: Would it be an option to do `--brkt-file file1 --brkt-file file2` instead of `--brkt-files file1 file2`?  It's more verbose, but it avoids ambiguity between a value that's passed to the option vs. a value that's passed to the command.  See the `--claim` argument in `make-token`.

brkt_cli/make_user_data/test_make_user_data.py
```diff
@@ -0,0 +1,86 @@
...
+from brkt_cli.instance_config_args import instance_config_args_to_values
+from brkt_cli.test_util import TestException
+
+
+class TestMakeUserData(unittest.TestCase):
+
+    def setUp(self):
+        super(TestMakeUserData, self).setUp()
+        name_fields = self.id().split('.')
+        self.class_name = name_fields[-2]
```
bburtin: Unused variable?

brkt_cli/make_user_data/test_make_user_data.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+            raise TestException(err_msg)
+        os.unlink(outfile.name)
+
+    def test_token_and_one_brkt_file(self):
+        values = instance_config_args_to_values('')
+        values.token = 'THIS_IS_NOT_A_JWT'
+        infile = os.path.join(self.testdata_dir, 'logging.yaml')
+        values.make_user_data_brkt_files = [ infile ]
+        self.run_cmd(values)
```
bburtin: Can this be done as a function call instead of starting up a separate process?

brkt_cli/make_user_data/test_make_user_data.py
```diff
@@ -0,0 +1,86 @@
...
+    def run_cmd(self, values):
+        muds = MakeUserDataSubcommand()
+        outfile = NamedTemporaryFile(delete=False)   # MAKE TRUE
+        muds.run(values, out=outfile)
+        outfile.close()
+
+        knowngood_file = os.path.join(self.testdata_dir,
+                                      self.test_name + ".out")
+        try:
+            subprocess.check_output(['diff', outfile.name, knowngood_file])
```
bburtin: This will only work on Unix.

brkt_cli/make_user_data/__init__.py
bonzotronix: Will do.

brkt_cli/make_user_data/test_make_user_data.py
bonzotronix: Thought I was gonna use it.  Will remove it.

brkt_cli/make_user_data/test_make_user_data.py
bonzotronix: I'll switch it to use unittest functions, as we discussed.

bonzotronix:
PTAL

bburtin:
I like it.  Thanks!  👍

PR #437, opened 2016-08-23
Copy over IOPS of guest root during image update

krishnanrs:
If the guest root volume type is PIOPS, copy over the IOPS settings during an image update.
This was verified locally by updating AMIs with and without PIOPS volumes.


jimvoll:
LGTM

PR #439, opened 2016-08-26
Use vpn.yaml brkt-file instead of brkt-config

bonzotronix:
To convey the FQDN to use for North-South cert requests, pass in a vpn.yaml file (to be stored in /var/brkt/instance_config/) rather than putting the field in the brkt-config part.


bburtin:
Oops, pressed the button instead of giving you 👍 .  It's definitely Friday.

PR #440, opened 2016-08-30
Allow --ca-cert arg for updating AMI/image

bonzotronix:


brkt_cli/instance_config_args.py
```diff
@@ -186,7 +186,7 @@ def instance_config_from_values(values=None, mode=INSTANCE_CREATOR_MODE,
         ic.add_brkt_file('proxy.yaml', proxy_config)
 
     if 'ca_cert' in values and values.ca_cert:
-        if mode != INSTANCE_CREATOR_MODE:
+        if mode not in (INSTANCE_CREATOR_MODE, INSTANCE_UPDATER_MODE):
```
bburtin: Do we even need this check, since we're selectively adding the option?

bonzotronix: I could remove it.  I guess having that check reminds us of the restriction.  But I'm OK either way.  Remove?

bburtin: Yeah, I generally think that less code is a good thing.  Thanks!

bonzotronix:
OK, PTAL

bburtin:
👍

PR #441, opened 2016-09-06
Add sub-network to GCE image launch command

krishnanrs:
Provide a subnetwork argument to launch a GCE image in a specific subnet. This was tested in my personal account.

### DIFF
--- a/brkt_cli/gce/launch_gce_image_args.py
+++ b/brkt_cli/gce/launch_gce_image_args.py
@@ -42,11 +42,15 @@ def setup_launch_gce_image_args(parser):
     parser.add_argument(
         '--network',
         dest='network',
+        metavar='NAME',
+        help='Launch instance in this network',
         default='default',
         required=False
     )
     parser.add_argument(
         '--subnetwork',
+        metavar='NAME',
+        help='Launch instance in this subnetwork',
         dest='subnetwork',
         required=False
     )
--- a/brkt_cli/gce/launch_gce_image_args.py
+++ b/brkt_cli/gce/launch_gce_image_args.py
@@ -45,6 +45,11 @@ def setup_launch_gce_image_args(parser):
         default='default',
         required=False
     )
+    parser.add_argument(
+        '--subnetwork',
+        dest='subnetwork',
+        required=False
+    )
 
     # Optional startup script. Hidden because it is only used for development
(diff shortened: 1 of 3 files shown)


brkt_cli/gce/launch_gce_image_args.py
```diff
@@ -45,6 +45,11 @@ def setup_launch_gce_image_args(parser):
         default='default',
         required=False
     )
+    parser.add_argument(
```
bburtin: Would it be more intuitive to add metavar='NAME'?

krishnanrs: I'd prefer adding a help text instead, for both subnetwork and network (see below). As is, the launch-gce-image is somewhat was a less used command by customers since the expectation is in the end they would use their own orchestration tools to launch the GCE images. This change however makes it easy for test and automation.

  --network NETWORK     Name of the network in which the instance would be
                        launched (default: default)
  --subnetwork SUBNETWORK
                        Name of the sub-network in which the instance would be
                        launched (default: None)

bburtin: The pattern I've been following to date is that the metavar gives the user a clue about the type of data they're dealing with (NAME, ID, N (integer), etc.).  I think this is better, since "NETWORK" doesn't provide insight into the type of the value that we're expecting.  Then you don't need to put "name" into the help text.  E.g.:
```
    parser.add_argument(
        '--subnet',
        metavar='ID',
        dest='subnet_id',
        help='Launch instances in this subnet'
    )
```

krishnanrs: You're suggesting something like this?
  --network NAME        Name of the network in which the instance would be
                        launched (default: default)
  --subnetwork NAME     Name of the sub-network in which the instance would be
                        launched (default: None)

bburtin: --network NAME The network in which the instance will be
launched
--subnetwork NAME The subnetwork in which the instance will be
launched

bburtin: Or even more concise: "Launch instance in this network", "Launch instance in this subnetwork"

bburtin:
👍

PR #445, opened 2016-09-19

apoorva-jindal:
1. Raise exception if environment variables for vcenter username
and password are not set.
2. Gracefully exit if SCSI controller is not found in the
Encryptor VM.


brkt_cli/esx/__init__.py
```diff
@@ -156,6 +156,10 @@ def command_update_encrypted_vmdk(values, parsed_config, log):
             raise ValidationError("Encrypted image not provided")
     if (values.source_image_path is not None and values.image_name is None):
         raise ValidationError("Specify the Metavisor OVF file.")
+    if os.getenv('VCENTER_USER_NAME') is None:
```
bburtin: `if not os.getenv('VCENTER_USER_NAME)` reads better.  You may also consider adding a utility function that checks for the existence of environment variables.  Your callsite would then look like `_check_env_vars_set('VCENTER_USER_NAME', 'VCENTER_PASSWORD', ...)

brkt_cli/esx/esx_service.py
```diff
@@ -471,10 +471,15 @@ def delete_serial_port_to_file(self, vm, filename):
     def add_disk(self, vm, disk_size=12*1024*1024,
                  filename=None, unit_number=0):
         spec = vim.vm.ConfigSpec()
+        controller = None
         for dev in vm.config.hardware.device:
             if isinstance(dev, vim.vm.device.VirtualSCSIController):
                 controller = dev
                 break
+        if controller is None:
+            log.error("Did not find a SCSI controller in VM %s",
```
bburtin: Do you need to both log the error and raise an Exception?  This won't result in duplicate logging?  Will the untyped exception result in a stack trace being written to the user's console?

PR #446, opened 2016-09-20

apoorva-jindal:
1. Raise exception if environment variables for vcenter username and password are not set.
2. Gracefully exit if SCSI controller is not found in the Encryptor VM.
3. Fix the destination of memory variable.


brkt_cli/esx/__init__.py
```diff
@@ -45,6 +45,12 @@
 log = logging.getLogger(__name__)
 
 
+def _check_env_vars_set(var_names):
```
bburtin: Would read better if you make this `*var_names`.  The callsite would then be `_check_env_vars_set('VCENTER_USER_NAME', 'VCENTER_PASSWORD')`.

brkt_cli/esx/encrypt_vmdk_args.py
```diff
@@ -65,7 +65,7 @@ def setup_encrypt_vmdk_args(parser):
         "--memory",
         help="Memory to assign to Encryptor VM",
         metavar='N',
-        dest="GB",
+        dest="memoryGB",
```
bburtin: Python convention is `memory_gb`.

brkt_cli/esx/update_encrypted_vmdk_args.py
```diff
@@ -58,7 +58,7 @@ def setup_update_vmdk_args(parser):
         "--memory",
         help="Memory to assign to Encryptor VM",
         metavar='N',
```
bburtin: I don't think it will be clear from the command line usage that the value is in gigabytes.  Would it be more intuitive to call the option `"--memory-gb"`, or set `metavar='GB'`?

PR #447, opened 2016-09-20
Re-add subnetwork argument to launch-gce-image

krishnanrs:
Re-adding the subnetwork argument to the launch-gce-image command.
This was accidentally removed in commit: a599ad9a6b4a4afdcdbcd67284efff9c2de34d16.
Tested this locally by launching a new GCE image.


bburtin:
Thanks Krishnan.  Can you add some context to the commit message?  It's not clear why you're adding the argument but not handling it in the code.

bburtin:
👍

PR #448, opened 2016-09-20

apoorva-jindal:
1. Raise exception if environment variables for vcenter username and password are not set.
2. Gracefully exit if SCSI controller is not found in the Encryptor VM.
3. Fix the destination of memory variable.


bburtin:
👍

PR #449, opened 2016-09-20
Allow passing guest FQDN when launching GCE instances

krishnanrs:
The guest FQDN is required for north-south configuration to get enabled on the instance. This option allows setting this user-data value when launching GCE instances. This is very useful in testing and potentially automation of north-south.
Was tested locally in my account.


robertabbott:
once tests pass lgtm

PR #451, opened 2016-09-20
Validate the checksum of the downloaded OVF image

apoorva-jindal:
Validate the checksums of the OVF file as well as all the disks.
This will detect any interrupted downloads or corrupted files.


brkt_cli/esx/esx_service.py
```diff
@@ -744,8 +753,20 @@ def upload_ovf_to_vcenter(self, target_path, ovf_name):
         vm = None
         content = self.si.RetrieveContent()
         manager = self.si.content.ovfManager
-        spec_params = vim.OvfManager.CreateImportSpecParams()
+        # Load checksums for each file
+        mf_checksum = None
+        mf_file_name = ovf_name[:ovf_name.find(".ovf")] + ".mf"
```
bburtin: Looks like this will chop the last character if the filename doesn't end with ".ovf".  Is this a problem?

brkt_cli/esx/esx_service.py
```diff
@@ -744,8 +753,20 @@ def upload_ovf_to_vcenter(self, target_path, ovf_name):
...
+        mf_path = os.path.join(target_path, mf_file_name)
+        with open(mf_path, 'r') as mf_file:
+            mf_checksum = json.load(mf_file)
+        # Validate ovf file
         ovf_path = os.path.join(target_path, ovf_name)
+        ovf_checksum = mf_checksum[(os.path.split(ovf_path))[1]]
+        if ovf_checksum != compute_sha1_of_file(ovf_path):
+            raise Exception("OVF file checksum does not match. "
```
bburtin: Should this be a `ValidationError`, so that the user doesn't see a traceback in their console?

brkt_cli/esx/esx_service.py
```diff
@@ -795,6 +817,12 @@ def upload_ovf_to_vcenter(self, target_path, ovf_name):
                         lease.HttpNfcLeaseComplete()
                         self.destroy_vm(vm)
                     raise Exception("Failed to find VMDKs for the Metavisor OVF")
+                # Validate the checksum of the file
+                file_checksum = mf_checksum[d_file_name]
+                if file_checksum != compute_sha1_of_file(file_path):
+                    raise Exception("Disk file %s checksum does not match. "
```
bburtin: `ValidationError`?

brkt_cli/esx/esx_service.py
apoorva-jindal: No, the ovf_name is derived from having a file ending in ovf - if it does not, an Exception with proper error string will happen way early.

PR #454, opened 2016-09-21
Switch default bucket to prod bucket

apoorva-jindal:
1. Default ovf-image should be prod.
2. Only retain release directories in the bucket.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -850,11 +850,9 @@ def download_ovf_from_s3(bucket_name, image_name=None):
             c_list = list(bucket.list("", "/"))
             dir_list = []
             for content in c_list:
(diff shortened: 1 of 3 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -850,11 +850,9 @@ def download_ovf_from_s3(bucket_name, image_name=None):
...
             dir_list = []
             for content in c_list:
-                dir_list.append(str(content.name))
-            try:
-                dir_list.remove('Users/')
-            except:
-                pass
+                dir_name = str(content.name)
+                if "release" in dir_name:
+                    dir_list.append(str(content.name))
```
bburtin: `dir_list.append(dir_name)`?

bburtin:
👍

PR #455, opened 2016-09-29
Add all common command line options to ESX

gdmen:
Passing userdata in MIME - as AWS & GCE do - allows ESX to use all shared
command line options (e.g. proxy settings). This change also removes
redundant logic in the ESX creation of userdata.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -33,7 +33,7 @@ from brkt_cli.util import (
     retry,
     RetryExceptionChecker
 )
-from brkt_cli import add_brkt_env_to_brkt_config
+from brkt_cli.instance_config import INSTANCE_UPDATER_MODE
 from brkt_cli.validation import ValidationError
 from boto.s3.key import Key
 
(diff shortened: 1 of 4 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -612,34 +613,26 @@ def clone_vm(self, vm, powerOn=False, vm_name=None, template=False):
...
-                            token=None,
-                            ssh_key_file=None, update=False,
+    def create_userdata_str(self, instance_config, update=False,
+                            ssh_key_file=None,
                             rescue_proto=None, rescue_url=None):
         try:
-            if instance_config is None:
-                user_data = dict()
-            else:
-                user_data = json.loads(instance_config.make_brkt_config_json())
```
apoorva-jindal: What happens when instance_config is None?

gdmen: I made this a required param, but just looked again and we send None from the rescue command. Good catch, fixing.

brkt_cli/esx/esx_service.py
```diff
@@ -612,34 +613,28 @@ def clone_vm(self, vm, powerOn=False, vm_name=None, template=False):
...
-                user_data['brkt']['rescue'] = dict()
-                user_data['brkt']['rescue']['protocol'] = rescue_proto
-                user_data['brkt']['rescue']['url'] = rescue_url
-            user_data_str = json.dumps(user_data)
-            return user_data_str
+                brkt_config = dict()
+                brkt_config['rescue'] = dict()
+                brkt_config['rescue']['protocol'] = rescue_proto
+                brkt_config['rescue']['url'] = rescue_url
+            instance_config.set_brkt_config(brkt_config)
```
apoorva-jindal: Will I need instance config to be non-None here? How will it work with rescue command?

apoorva-jindal:
LGTM

PR #458, opened 2016-10-03
Fix teardown implementation

apoorva-jindal:
1. Pass no-teardown value from user argument to esx-service.
Without this change, configuration of no-teardown has no impact.
2. Print final output (template/ovf/ova name) to stdout.
This makes ESX output the same as AWS where the output of the operation
is written to stdout and not stderr.

### DIFF
--- a/brkt_cli/esx/encrypt_vmdk.py
+++ b/brkt_cli/esx/encrypt_vmdk.py
@@ -94,17 +94,18 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
             if target_path is None:
                 raise Exception("Cannot create ova/ovf as target path is None")
             ovf = vc_swc.export_to_ovf(vm, target_path, ovf_name=image_name)
-            log.info("OVF is at %s" % ovf)
             if create_ova is True:
                 if ovftool_path is not None:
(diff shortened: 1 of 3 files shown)


brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -99,12 +99,16 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
                 if ovftool_path is not None:
                     ova = vc_swc.convert_ovf_to_ova(ovftool_path, ovf)
                     log.info("OVA is at %s" % ova)
+                    print(ova)
```
bburtin: Do you need to get rid of the log message above?  Seems like you're logging the same thing twice.

brkt_cli/esx/update_vmdk.py
```diff
@@ -86,6 +86,9 @@ def update_ovf_image_mv_vm(vc_swc, enc_svc_cls, guest_vm, mv_vm,
                     os.system(rm_cmd)
                     ova = vc_swc.convert_ovf_to_ova(ovftool_path, ovf)
                     log.info("OVA is at %s" % ova)
+                    print(ova)
```
bburtin: Another duplicate message?

bburtin:
👍

PR #459, opened 2016-10-03
Make the VM name a positional argument

apoorva-jindal:
The intention of this operation is to change the core upload
location of the MV VM being identified by its name.
Hence, moving the vm-name to a positional argument.

### DIFF
--- a/brkt_cli/esx/rescue_metavisor_args.py
+++ b/brkt_cli/esx/rescue_metavisor_args.py
@@ -31,11 +31,9 @@ def setup_rescue_metavisor_args(parser):
         metavar='NAME',
         required=True)
     parser.add_argument(
-        '--vm-name',
-        metavar='NAME',
-        dest='vm_name',
-        help='Specify the name of the metavisor VM',
-        required=True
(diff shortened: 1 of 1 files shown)


bburtin:
Can you please update the commit message and describe why you're making this change?  Usually a positional argument is the target of the operation that you're performing, as opposed to a property.  If this is the case, I'm fine with this change.

brkt_cli/esx/rescue_metavisor_args.py
```diff
@@ -31,11 +31,9 @@ def setup_rescue_metavisor_args(parser):
         metavar='NAME',
         required=True)
     parser.add_argument(
-        '--vm-name',
+        'vm_name',
         metavar='NAME',
```
bburtin: Should this be `VM_NAME`?  `NAME` is cleaner with other options because you see the option name in front of it (e.g. `--vcenter-datacenter NAME`).  But for positional arguments, I believe you'll just see `NAME` with no context.

Also, do you need to add `nargs=1` for better validation?

PR #460, opened 2016-10-04
Fix pyvmomi version to be between 5.5.0 and 6.0.0

apoorva-jindal:
This avoids using the latest version which is broken on MAC.


bburtin:
:+1:

PR #461, opened 2016-10-04
Hide bucket-name argument

apoorva-jindal:
Since customers should always be using the prod bucket
to fetch MV images from, hiding the bucket-name arguement
from encrypt-vmdk and update-vmdk.


krishnanrs:
👍

bburtin:
👍

PR #462, opened 2016-10-05
Return instance-name on gce-launch-image

Garpetun:
We did not do this before, and it results in us printing "None" to
stdout when someone issues the "brkt --version" command.


bburtin:
👍

Garpetun:
Err I must have zoned out, I mean the output from gce-launch-image.

bburtin:
You can amend the commit locally, then `git push -f origin return-instance-name-on-launch`

bburtin:
👍

PR #463, opened 2016-10-05
Fix typo in install_requires in setup.py

bonzotronix:
Missing a comma after a package named in the 'install_requires' list.


krishnanrs:
👍

PR #465, opened 2016-10-06
Add token to config

mpage:
Users may now specify a token in their config file that will be
used as by default when encrypting, updating, or launching images.


krishnanrs:
👍

bburtin:
👍

PR #466, opened 2016-10-07
Change default cpu-count to 8 and memory to 32G

apoorva-jindal:


krishnanrs:
The changes LGTM. But do you want to wait until tomorrow before merging this pull request? We have a customer scrum and check with the rest of the folks as to what they want for this. If this was an ask from PM, go ahead with the merge.

PR #469, opened 2016-10-21
Make esx encryption use aws credentials to get image from s3

atonderski:
Forcing anonymous s3 access to the bucket locked us out from
using non-public buckets. This change addresses that by allowing
the use aws credentials (if they exist).

If no credentials are found in the environment, we will fall back
to an anonymous connection.


bburtin:
👍

PR #470, opened 2016-10-22
Provide ability to add guest cloud-init to MIME data

krishnanrs:
Provides ability to add guest cloud-init to the MIME instance user-data. See http://cloudinit.readthedocs.io/en/latest/topics/format.html#helper-script-to-generate-mime-messages for more information.

### DIFF
--- a/brkt_cli/instance_config.py
+++ b/brkt_cli/instance_config.py
@@ -47,7 +47,13 @@ class BrktFile(object):
         self.dest_filename = dest_filename
         self.file_contents = file_contents
 
+class GuestFile(object):
+    def __init__(self, dest_file, content_type, file_contents):
+        self.dest_file = dest_file
+        self.content_type = content_type
+        self.file_contents = file_contents
 
+        
 class InstanceConfig(object):
     """ Class containing common settings for Brkt instances """
 
@@ -57,6 +63,7 @@ class InstanceConfig(object):
 
         self.brkt_config = brkt_config
         self._brkt_files = []
+        self._guest_files = []
         self.set_mode(mode)
 
     def brkt_files_dest_dir(self):
@@ -67,6 +74,9 @@ class InstanceConfig(object):
         dest_path = posixpath.join(self._brkt_files_dest_dir, dest_filename)
         brkt_file = BrktFile(dest_path, file_contents)
         self._brkt_files.append(brkt_file)
+
+    def add_guest_file(self, guest_file):
+        self._guest_files.append(guest_file)
 
     def set_mode(self, mode=INSTANCE_CREATOR_MODE):
         self._mode = mode
@@ -93,4 +103,6 @@ class InstanceConfig(object):
         for brkt_file in self._brkt_files:
             udc.add_file(brkt_file.dest_filename, brkt_file.file_contents,
                          BRKT_FILES_CONTENT_TYPE)
+        for guest_file in self._guest_files:
+            udc.add_part(guest_file.content_type, guest_file.file_contents)
         return udc.to_mime_text()
--- a/brkt_cli/make_user_data/__init__.py
+++ b/brkt_cli/make_user_data/__init__.py
@@ -15,10 +15,13 @@ import argparse
 import logging
 import os
 import sys
+import re
 
 import brkt_cli
 from brkt_cli.subcommand import Subcommand
 from brkt_cli.instance_config import INSTANCE_METAVISOR_MODE
+from brkt_cli.instance_config import GuestFile
+from brkt_cli.validation import ValidationError
 from brkt_cli.instance_config_args import (
     instance_config_from_values,
     setup_instance_config_args
@@ -27,6 +30,7 @@ log = logging.getLogger(__name__)
 
 
 def _add_files_to_instance_config(instance_cfg, files_list):
+    
     for fname in files_list:
         try:
             with open(fname, 'r') as f:
@@ -36,7 +40,18 @@ def _add_files_to_instance_config(instance_cfg, files_list):
             raise Exception('Unable to read file: %s' % (fname,))
         instance_cfg.add_brkt_file(os.path.basename(fname), file_contents)
 
-
+        
+def _add_guest_files_to_instance_config(instance_cfg, guest_files):
+    
+    for guest_file in guest_files:
+        try:
+            with open(guest_file.dest_file, 'r') as f:
+                guest_file.file_contents = f.read()
+        except IOError:
+            raise ValidationError('Unable to read file: %s' % guest_file.dest_file)
+        instance_cfg.add_guest_file(guest_file)
+        
+        
 class MakeUserDataSubcommand(Subcommand):
 
     def name(self):
@@ -69,6 +84,14 @@ class MakeUserDataSubcommand(Subcommand):
             action='append',
             help=argparse.SUPPRESS
         )
+        parser.add_argument(
+            '--guest-user-data-file',
+            metavar='FILENAME:TYPE',
+            dest='make_user_data_guest_files',
+            action='append',
+            help=('User-data file and MIME contents type to be passed to the '
+                  'guest instance. Can be specified multiple times.')
+        )
         # Certain customers need to set the FQDN of the guest instance, which
         # is used by Metavisor as the CN field of the Subject DN in the cert
         # requests it submits to an EST server (for North-South VPN tunneling).
@@ -89,6 +112,17 @@ class MakeUserDataSubcommand(Subcommand):
         if values.make_user_data_brkt_files:
             _add_files_to_instance_config(instance_cfg,
                                           values.make_user_data_brkt_files)
+
+        if values.make_user_data_guest_files:
+            guest_files = []
+            for fname in values.make_user_data_guest_files:
+                match = re.match('(.+):(.+)', fname)
+                if match:
+                    guest_file = GuestFile(match.group(1), match.group(2), None)
+                    guest_files.append(guest_file)
+                else:
+                    raise ValidationError('Unable to parse guest file and type: %s' % fname)
+            _add_guest_files_to_instance_config(instance_cfg, guest_files)
 
         if values.make_user_data_guest_fqdn:
             vpn_config = 'fqdn: %s\n' % (values.make_user_data_guest_fqdn,)
--- a/brkt_cli/esx/encrypt_vmdk_args.py
+++ b/brkt_cli/esx/encrypt_vmdk_args.py
@@ -59,14 +59,14 @@ def setup_encrypt_vmdk_args(parser):
         help="Number of CPUs to assign to Encryptor VM",
         metavar='N',
         dest="no_of_cpus",
-        default="1",
+        default="8",
         required=False)
     parser.add_argument(
         "--memory",
         help="Memory to assign to Encryptor VM",
         metavar='GB',
         dest="memory_gb",
-        default="1",
+        default="32",
         required=False)
     parser.add_argument(
         '--encrypted-image-name',
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -867,8 +867,8 @@ def download_ovf_from_s3(bucket_name, image_name=None):
     ovf_name = None
     download_file_list = []
     try:
-        conn = boto.connect_s3(None, None, anon=True,
-                               host="s3.amazonaws.com")
+        anon = not (set(['AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY']) <= set(os.environ))
+        conn = boto.connect_s3(None, None, anon=anon, host="s3.amazonaws.com")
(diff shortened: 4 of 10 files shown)


brkt_cli/instance_config.py
```diff
@@ -68,6 +69,9 @@ def add_brkt_file(self, dest_filename, file_contents):
         brkt_file = BrktFile(dest_path, file_contents)
         self._brkt_files.append(brkt_file)
 
+    def add_guest_file(self, file_contents, format_type):
```
bburtin: Would it be clearer to call this `content_type` instead of `format_type`?

brkt_cli/make_user_data/__init__.py
```diff
@@ -25,16 +25,27 @@
...
+DEFAULT_CLOUDINIT_FORMAT = 'text/x-shellscript'
 
-def _add_files_to_instance_config(instance_cfg, files_list):
+
+def _add_files_to_instance_config(instance_cfg, files_list,
+                                  guest_files=False):
     for fname in files_list:
+        if guest_files:
+            if ':' in fname:
+                (fname, format_type) = fname.split(":", 1)
```
bburtin: How about doing a regular expression match of `(.+):(.+\/.+)`?  That would handle filenames with a colon, and arguments with multiple colons.

bonzotronix:
It should be quick (and illustrative) to add some tests in test_make_user_data.py.  That way we all see what the userdata looks like with guest-files added.

brkt_cli/make_user_data/__init__.py
```diff
@@ -69,6 +80,13 @@ def register(self, subparsers, parsed_config):
             action='append',
             help=argparse.SUPPRESS
         )
+        parser.add_argument(
+            '--guest-userdata-file',
```
bburtin: Should we call this `--user-data-file` instead?  "guest" is an implementation detail, and seems gratuitous, since `--brkt-file` is hidden.  Also, "user data" is usually written as either hyphenated or as separate words in the cloud-init and AWS docs.

brkt_cli/make_user_data/__init__.py
```diff
@@ -69,6 +80,13 @@ def register(self, subparsers, parsed_config):
...
+            metavar='FILENAME:TYPE',
+            dest='make_user_data_guest_files',
+            action='append',
+            help=argparse.SUPPRESS
```
bburtin: Why are we suppressing this?  Seems like something that customers would commonly want to do.

brkt_cli/make_user_data/__init__.py
```diff
@@ -25,16 +25,27 @@
...
 
+DEFAULT_CLOUDINIT_FORMAT = 'text/x-shellscript'
 
-def _add_files_to_instance_config(instance_cfg, files_list):
+
+def _add_files_to_instance_config(instance_cfg, files_list,
+                                  guest_files=False):
     for fname in files_list:
+        if guest_files:
+            if ':' in fname:
```
bburtin: The colon handling should be done during command line parsing.  This would make the behavior of the lower-level function more clearly defined.  The function can accept a list of objects that contain two fields: the path to the file and the content type.  Or at least make it a map of path->content-type.  This would be ok, as long as order isn't important.

brkt_cli/instance_config.py
```diff
@@ -93,4 +97,6 @@ def make_userdata(self):
         for brkt_file in self._brkt_files:
             udc.add_file(brkt_file.dest_filename, brkt_file.file_contents,
                          BRKT_FILES_CONTENT_TYPE)
+        for brkt_file in self._guest_files:
+            udc.add_part(brkt_file[1], brkt_file[0])
```
bonzotronix: For clarity, could we create a GuestFile class (like the BrktFile class) and use that here?  It's easy to get confused with tuples of ordered properties.

brkt_cli/make_user_data/__init__.py
bonzotronix: +1 (though I'd also be fine with `--userdata-file`)

bburtin: Oh.  Now that I think about it, it should probably be a positional argument (no dashed option name) instead of an optional argument: `brkt make-user-data FILE1 FILE2 ...`

brkt_cli/instance_config.py
bburtin: We can't just use `BrktFile`?

bonzotronix: Different properties.  BrktFiles have dest filenames and a non-configurable content-type.

krishnanrs:
Most of the review comments should be addressed. If you feel strongly about specific issues that needs to be addressed, let me know (again).

brkt_cli/instance_config.py
bonzotronix: +1 on boris' suggestion here.

brkt_cli/make_user_data/__init__.py
bonzotronix: +1

brkt_cli/make_user_data/__init__.py
krishnanrs: I want to come around to this if we have an actual use case. The likely hood of a filename having a colon is low and the users of this option are expected to be fairly familiar with cloud-init and hence would know what they are doing.

bburtin: I don't understand.  What's the downside of making the code more robust?  The behavior isn't different if you use a regexp.  Without the regexp, if someone has a colon in their filename, it will lead to silent failure.

brkt_cli/make_user_data/__init__.py
```diff
@@ -26,15 +27,26 @@
 log = logging.getLogger(__name__)
 
 
-def _add_files_to_instance_config(instance_cfg, files_list):
+def _add_files_to_instance_config(instance_cfg, files_list,
+                                  guest_files=False):
     for fname in files_list:
+        if guest_files:
+            match = re.match('(.+)(.+\/.+)?:(.+)', fname)
```
bburtin: The regexp matching should be done outside the function, since it applies to command line parsing.  The function should accept a list of `GuestFile` objects, which specify the path and content type of each file.

brkt_cli/make_user_data/__init__.py
```diff
@@ -69,6 +81,14 @@ def register(self, subparsers, parsed_config):
             action='append',
             help=argparse.SUPPRESS
         )
+        parser.add_argument(
+            '--guest-user-data-file',
```
bburtin: Should we make the list of files positional arguments and get rid of `--guest-user-data-file` altogether?

brkt_cli/make_user_data/__init__.py
```diff
@@ -69,6 +81,14 @@ def register(self, subparsers, parsed_config):
...
+            metavar='FILENAME:TYPE',
```
bburtin: `FILENAME[:TYPE]` would clearly specify that the TYPE part is optional.

brkt_cli/make_user_data/__init__.py
```diff
@@ -69,6 +81,14 @@ def register(self, subparsers, parsed_config):
...
+            dest='make_user_data_guest_files',
+            action='append',
+            help=('User-data file and type to be passed to the guest instance. '
```
bburtin: "MIME content type" would be clearer than "type."  The default should be `text/x-shellscript`.

brkt_cli/make_user_data/__init__.py
krishnanrs: I prefer having separate arguments. What if the user wants to add MV specific user-data, which is not intended for the guest, in which case they need to use the --brkt-file argument?

brkt_cli/make_user_data/__init__.py
krishnanrs: That's what I mentioned in my previous update. With the regex parsing, who am I supposed to know if the last section is the content_type or not? for e.g. file:name vs filename:content_type.

brkt_cli/make_user_data/__init__.py
krishnanrs: Will address this and remove the default text.

brkt_cli/make_user_data/__init__.py
bburtin: It's definitely a judgment call.  If you think about `make-user-data` in an AWS-only context, it would read like "take these files and generate a MIME blob."  I would argue that `--brkt-file` (if we expose it) is the exception, and clearly specifies that the following file is Bracket-internal.  If you really think that we need the distinction, we can call it `--guest-file`.  `user-data` seems redundant, since the command is already called `make-user-data`.

brkt_cli/make_user_data/__init__.py
bburtin: My proposal was to require the content type to have a slash in it.  So you'd only match the content type in the case of `filename:text/x-shellscript`.

brkt_cli/make_user_data/__init__.py
krishnanrs: I accept the first part, however I do not understand the second part. Why should a list of GuestFile objects be passed to this method? The better option would be re-write this function to pass either a BrktFile or GuestFile object as the "file_list" and let the function handle the respective objects. This also means that further down more simplification can be done as these objects can be passed directly (to the instance_config methods).

brkt_cli/make_user_data/__init__.py
krishnanrs: If you look at the original cloud-init script (for which I pasted the link), the "text/" prefix is common for all the cloud-init types and hence treated as optional (it automatically gets prefixed when not specified). I'm just trying to preserve the behavioural consistency of cloud-init itself.

brkt_cli/make_user_data/__init__.py
bonzotronix: Could also change _add_files_to_instance_config() to _add_brkt_files_to_instace_config() and add a _add_guest_files_to_instance_config().  Or just get rid of that function and move the logic to MakeUserDataSubcommand.run() (perhaps with _get_file_contents() utility function).

brkt_cli/make_user_data/__init__.py
krishnanrs: I explained this to @bburtin . The cloud-init contents are all formatted as MIMEText which implies the "text/" prefix by default, if not specified. That is how the sample scripts (and examples that I sent out earlier) work.

brkt_cli/make_user_data/__init__.py
```diff
@@ -36,7 +39,19 @@ def _add_files_to_instance_config(instance_cfg, files_list):
...
+        
+def _add_guest_files_to_instance_config(instance_cfg, guest_files):
+    
+    for guest_file in guest_files:
+        try:
+            with open(guest_file.dest_file, 'r') as f:
+                guest_file.file_contents = f.read()
+        except IOError:
+            log.error('Input file not readable: %s', guest_file.dest_file)
+            raise Exception('Unable to read file: %s' % (guest_file.dest_file,))
```
bburtin: Will this result in double-logging?  Would it be better to just raise `ValidationError`, so that the user doesn't see a traceback?  Also, there's no need to make `guest_file.dest_file` a tuple.

brkt_cli/make_user_data/__init__.py
```diff
@@ -90,6 +113,17 @@ def run(self, values, out=sys.stdout):
             _add_files_to_instance_config(instance_cfg,
                                           values.make_user_data_brkt_files)
 
+        if values.make_user_data_guest_files:
+            guest_files = []
+            for fname in values.make_user_data_guest_files:
+                match = re.match('(.+)(.+\/.+)?:(.+)', fname)
```
bburtin: Seems like the `(.+.\/.+)?` is unnnecessary.  I originally wanted to do that for the purpose of requiring `text/` at the beginning of the content type.  If we're not going to require `text/`, I think you should be able to simplify the regexp.

brkt_cli/make_user_data/__init__.py
```diff
@@ -90,6 +113,17 @@ def run(self, values, out=sys.stdout):
...
+                if match:
+                    guest_file = GuestFile(match.group(1), match.group(3), None)
+                    guest_files.append(guest_file)
+                else:
+                    raise Exception('Unable to parse guest file: %s' % (fname,))
```
bburtin: Will this result in a traceback?  Also, the tuple is not necessary.

brkt_cli/make_user_data/__init__.py
krishnanrs: Addressed this in the new update.

brkt_cli/make_user_data/__init__.py
krishnanrs: Addressed this in the new update. Instead of an Exception, generated a ValidationError

bburtin:
Looks good.  Please squash your commits before checking in.  Supposedly the GitHub UI supports this now.  Not sure if Herb needs to flip a switch to turn it on.  https://github.com/blog/2141-squash-your-commits

PR #471, opened 2016-10-24
Add no-cleanup option to gce encrypt and update

robertabbott:
the no cleanup option prevents the encryptor
from being deleted. In case more debugging
is required than just looking at the logs

Testing: ran encrypt-gce-image with the no-cleanup
option and interrupted it

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -95,7 +95,8 @@ class EncryptGCEImageSubcommand(Subcommand):
             image_bucket=values.bucket,
             network=values.network,
             subnetwork=values.subnetwork,
-            status_port=values.status_port
+            status_port=values.status_port,
+            cleanup=values.cleanup
         )
         # Print the image name to stdout, in case the caller wants to process
         # the output.  Log messages go to stderr.
@@ -158,7 +159,8 @@ class UpdateGCEImageSubcommand(Subcommand):
             image_bucket=values.bucket,
             network=values.network,
             subnetwork=values.subnetwork,
-            status_port=values.status_port
+            status_port=values.status_port,
+            cleanup=values.cleanup
         )
 
         print(updated_image_id)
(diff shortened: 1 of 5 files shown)


brkt_cli/gce/__init__.py
```diff
@@ -95,7 +95,8 @@ def run(self, values):
             image_bucket=values.bucket,
             network=values.network,
             subnetwork=values.subnetwork,
-            status_port=values.status_port
+            status_port=values.status_port,
+            no_cleanup=values.no_cleanup
```
bburtin: Can we call this `cleanup` instead of `no_cleanup`?  I try to avoid negative names in variables, because it leads to double-negatives and makes code harder to read.

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -84,6 +84,13 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         help=argparse.SUPPRESS
     )
     parser.add_argument(
+        '--no-cleanup',
+        dest='no_cleanup',
```
bburtin: Same thing here.  Call the variable `cleanup` and make the default value `True`.  The argument name can still be `--no-cleanup`, and the `action` would be `store_false`.

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -84,6 +84,13 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         help=argparse.SUPPRESS
     )
     parser.add_argument(
+        '--no-cleanup',
+        dest='cleanup',
+        required=False,
+        action='store_false',
+        help=argparse.SUPPRESS
```
bburtin: I think you need a `default=True` here.

bburtin:
👍

PR #473, opened 2016-10-31
Prompt customers for vcenter password.

apoorva-jindal:
If the vcenter password is not provided via the environment variable,
request for it on the customer prompt.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -15,6 +15,7 @@ import brkt_cli
 import logging
 import os
 import subprocess
+import getpass
 
 from brkt_cli.subcommand import Subcommand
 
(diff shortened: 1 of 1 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -169,7 +170,12 @@ def command_update_encrypted_vmdk(values, parsed_config, log):
             raise ValidationError("Encrypted image not provided")
     if (values.source_image_path is not None and values.image_name is None):
         raise ValidationError("Specify the Metavisor OVF file.")
-    _check_env_vars_set('VCENTER_USER_NAME', 'VCENTER_PASSWORD')
+    _check_env_vars_set('VCENTER_USER_NAME')
+    vcenter_password = None
```
bburtin: `vcenter_password = os.getenv('VCENTER_PASSWORD')`
`if not vcenter_password:`
   `vcenter_password = getpass.getpass(...)`

bburtin: Also, seems like you're doing the same thing 3 times.  Can we turn it into a private function?

bburtin:
👍

PR #475, opened 2016-11-01
Replace curl with requests library to upload and download ovfs.

apoorva-jindal:
This change allows ESX brkt-cli to run on Windows as well.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -21,6 +21,7 @@ import atexit
 import os
 import signal
 import hashlib
+import requests
 import boto.s3
 
 from functools import wraps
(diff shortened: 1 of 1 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -867,11 +868,10 @@ def upload_ovf_to_vcenter(self, target_path, ovf_name,
...
                     dev_url = device_url.url.replace("https://*", host_name)
-                curl_cmd = (
-                    "curl -Ss -X POST --insecure -T %s -H 'Content-Type: \
-                    application/x-vnd.vmware-streamVmdk' %s" %
-                    (file_path, dev_url))
-                os.system(curl_cmd)
+                headers = {"Content-Type" : "application/x-vnd.vmware-streamVmdk",
+                           "Connection" : "Keep-Alive"}
+                with open(file_path, 'rb') as f:
+                    requests.post(dev_url, data=f, verify=False, headers=headers)
```
bburtin: Does `verify=False` make sense in production?

apoorva-jindal: Disabling verification as it is expected that customers may provide certificate to vCenter, but not sure if they will distribute a certificate to all ESX hosts.

PR #476, opened 2016-11-01
MV encrypted image name should not include .ovf.

apoorva-jindal:
Not having .ovf extenstion in the MV image name makes
the user interface consistent between getting MV image
from S3 or locally.


bburtin:
What if the user specifies a name that ends with `.ovf` on the command line?  Would we end up with `.ovf.ovf`?  Should we do some validation to handle this case?

bburtin:
👍

PR #477, opened 2016-11-01
Add validation for NAME=VALUE tags

krishnanrs:
The name and value pairs for a key tag must only contain alpha-numeric and hyphen characters.

### DIFF
--- a/brkt_cli/brkt_jwt/__init__.py
+++ b/brkt_cli/brkt_jwt/__init__.py
@@ -62,6 +62,7 @@ class MakeTokenSubcommand(Subcommand):
         if values.claims:
             for name_value in values.claims:
                 name, value = util.parse_name_value(name_value)
+                validate_name_value(name, value)
                 if name in claims:
                     raise ValidationError('Claim %s specified multiple times' % name)
                 claims[name] = value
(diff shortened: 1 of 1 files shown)


brkt_cli/util.py
```diff
@@ -173,6 +173,19 @@ def urlsafe_b64decode(base64_string):
...
 
+def validate_name_value(name, value):
+    """ Validate the format of a NAME=VALUE pair.
+
+    : return: a tuple of name, value
+    : raise: ValidationError if the format is invalid
+    """
+    if not re.match(r'[A-Za-z0-9_\-]*[A-Za-z0-9_\-]$', name) or \
+        not re.match(r'[A-Za-z0-9_\-]*[A-Za-z0-9_\-]$', value):
+        raise ValidationError('NAME=VALUE claim must only contain [A-Za-z0-9\-_]')
```
bburtin: I would say `must only contain letters, numbers, "-", and "_"`.  I don't know if it's safe to assume that the user speaks regex.  :-)

Can we simplify the regex? What about `[A-Za-z0-9_\-]+$`

brkt_cli/util.py
```diff
@@ -173,6 +173,19 @@ def urlsafe_b64decode(base64_string):
     return base64.urlsafe_b64decode(base64_string)
 
 
+def validate_name_value(name, value):
```
bburtin: Since this validation only applies to claims, I would put this function in `brkt_jwt/__init__.py` and call it from there after you parse the name/value.

krishnanrs:
Tags are used both for JWTs and for AWS instances (during encryption/updates). The syntax check is enforced only for JWT tags.

bburtin:
👍

PR #478, opened 2016-11-03
Print only fatal messages from boto

apoorva-jindal:
Boto tries to fetch the AWS credentials from metadata service before
doing a S3 download. If it does not get those credentials (which are
not present in ESX), it prints a confusing info message which looks
like an error. Hence, making the change to print only fatal messages
from boto.


bburtin:
👍

PR #479, opened 2016-11-08
No need to slip into 'legacy' mode for FreeBSD metavisors

jimvoll:
FeeBSD metavisors have support for sriov drivers so we
don't have to play games with guest AMI's that require
sriov.

This preserves any 'pay as you go' licensing for such
guests.


krishnanrs:
👍

PR #481, opened 2016-11-10
Better validation of command line arguments

krishnanrs:
Added better validation of the command line arguments for the
update-gce-image operand.

### DIFF
--- a/brkt_cli/gce/update_encrypted_gce_image_args.py
+++ b/brkt_cli/gce/update_encrypted_gce_image_args.py
@@ -1,7 +1,7 @@
 import argparse
 
 
-def setup_update_gce_image_args(parser):
+def setup_update_gce_image_args(parser, parsed_config):
     parser.add_argument(
         'image',
         metavar='ID',
@@ -14,12 +14,13 @@ def setup_update_gce_image_args(parser):
         help='Specify the name of the generated encrypted image',
         required=False
     )
+    required_zone = parsed_config.get_option('encrypt-gce-image.zone', None)
     parser.add_argument(
         '--zone',
         help='GCE zone to operate in',
         dest='zone',
-        default='us-central1-a',
-        required=True
+        default=required_zone,
+        required=not bool(required_zone)
     )
     parser.add_argument(
         '--encryptor-image-bucket',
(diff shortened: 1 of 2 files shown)


krishnanrs:
Currently I'm overloading the encrypt-gce-image config options for the update-gce-image operation as well, since technically they would be the same. However if the suggestion is to have separate set of config options for those, then let me know.

brkt_cli/gce/update_encrypted_gce_image_args.py
```diff
@@ -14,12 +14,17 @@ def setup_update_gce_image_args(parser):
...
+    }
+    if zone_kwargs['default'] is None:
+        zone_kwargs['required'] = True
     parser.add_argument(
         '--zone',
-        help='GCE zone to operate in',
-        dest='zone',
-        default='us-central1-a',
-        required=True
+        **zone_kwargs
```
bburtin: Is there a way to avoid kwargs here?  Seems like you should be able to call `parsed_config.get_option()`, then set a `required` variable based on that value.

brkt_cli/gce/update_encrypted_gce_image_args.py
```diff
@@ -28,11 +33,17 @@ def setup_update_gce_image_args(parser):
...
+        'required': False,
+    }
+    if proj_kwargs['default'] is None:
+        proj_kwargs['required'] = True
     parser.add_argument(
         '--project',
-        help='GCE project name',
-        dest='project',
-        required=True
+        **proj_kwargs
```
bburtin: Same here.

brkt_cli/gce/update_encrypted_gce_image_args.py
krishnanrs: One way to do that would be something like:
    parser.add_argument(
        '--zone',
        help='GCE zone to operate in',
        dest='zone',
        default=parsed_config.get_option('encrypt-gce-image.zone', None),
        required=False if parsed_config.get_option('encrypt-gce-image.zone', None) else True
    )

If not, then let me know if you have any better suggestions.

bburtin:
👍

PR #482, opened 2016-11-11
Better validation of instance and image names

krishnanrs:
Image name is a mandatory parameter that is passed to the launch-gce-image command. We were not validating that previously.


bonzotronix:
My vote would have been to make the validation function's error messages generic and include the objectionable input name (since the name restrictions are the same for image and instance).  But I could also be OK with the proposed solution.

Leaving this open for additional feedback....

bburtin:
Do we really want to validate for launch?  If the launch fails intuitively when you get the name wrong, the name validation seems gratuitous.

krishnanrs:
The problem is, let's say the user makes a typo with both the instance-name and image-name, but the error message is still generic that says that image or instance name is not in the right format. at this point in time, the user realizes that they typoed the image-name, fix it and retry the command, now they again get the same error since the instance-name is incorrect. Instead if the error was specific that the instance-name or image-name is incorrect, then it is more obvious to the user.

That being said, I do agree that the checking of the image-name is gratuitous since if it is invalid, the command is anyways bound to fail.

PR #484, opened 2016-11-11
Install correct pyvmomi version based on python version

krishnanrs:


bburtin:
It would be helpful if the commit message said why you're making this change.  It will help us remember the reason for the change by looking at the commit logs.

apoorva-jindal:
LGTM

apoorva-jindal:
Pyvmomi versions before 6.0.0 do not verify the ssl certificate, while pyvmomi 6.0.0 uses the variable ssl.sslContext (introduced in python 2.7.9) to determine whether to verify the ssl certificate or not. This makes pyvmomi 6.0.0 incompatible with python versions 2.7.8 and lower. Hence, adding a check to ensure that for python versions 2.7.8 and lower, pyvmomi 5.5.0 is installed.

PR #487, opened 2016-11-11
Handle gce api failure better for project_exists

robertabbott:
If project_exists receives a 400 http error return
false (indicating the project doesn't exists).
For any other exception raise

### DIFF
--- a/brkt_cli/gce/gce_service.py
+++ b/brkt_cli/gce/gce_service.py
@@ -2,6 +2,7 @@
 
 import abc
 import datetime
+import json
 import re
 import socket
 import tempfile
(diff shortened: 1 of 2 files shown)


brkt_cli/gce/gce_service.py
```diff
@@ -276,8 +277,12 @@ def get_project(self, project):
     def project_exists(self, project=None):
         try:
             self.get_project(project)
+        except errors.HttpError as e:
+            if json.loads(e.content)['error']['code'] == 400:
+                return False
+            raise
         except:
-            return False
+            raise
```
bburtin: Is the second `except` necessary?

brkt_cli/gce/gce_service.py
```diff
@@ -276,8 +277,12 @@ def get_project(self, project):
     def project_exists(self, project=None):
         try:
             self.get_project(project)
+        except errors.HttpError as e:
+            if json.loads(e.content)['error']['code'] == 400:
```
krishnanrs: Should this be a 403/404 instead?

bburtin:
👍

PR #488, opened 2016-11-11
Ensure cryptography version is compatible

robertabbott:
1.5.x doesnt work on linux. This ensures
a compatible version of the python cryptography
library is installed


bburtin:
👍

PR #490, opened 2016-11-14
Fix GCE refactor after merge

krishnanrs:
Fix missing change in https://github.com/brkt/brkt-cli/pull/486 causing
errors when running GCE commands.


bburtin:
👍

PR #491, opened 2016-11-14
Attribute error when running brkt-cli

krishnanrs:
An attribute error was being reported when trying to save the
CLI Config. This happens in environments where the CLI Config
has never been initialized. This change ensures that the CLI
Config is attempted to be saved only if needed.

Also moved the _unlink_noraise method so that it can be called as a
common method instead of being restricted to the ConfigSubcommand class.


bburtin:
👍  on the code.  Please update the commit message, so that it's more specific about what went wrong, and explains why you had to move the `_unlink_no_raise()` function.

PR #492, opened 2016-11-14
Update the VMWare command line spec to follow our new convention. Mov…

krishnanrs:
Update the VMWare command line spec to follow our new convention. Move the following commands:

 - encrypt-vmdk (without --use-esx) -> vmware encrypt-with-vcenter
 - encrypt-vmdk (with --use-esx) -> vmware encrypt-with-esx-host
 - update-vmdk (without --user-esx) -> vmware update-with-vcenter
 - update-vmdk (with --use-esx) -> vmware encrypt-with-esx-host
 - rescue-metavisor -> vmware rescue-metavisor

Introduced 2 new environment variables:
 - ESX_USER_NAME to specify the username for the ESX Host
 - ESX_PASSWORD (Optional) to specify the password for the ESX Host.
    If not specified, the user will be prompted for the same.

Hide the old commands and log a deprecation warning when they are used.
Refactor the ESX module so that the run() functions in both the old
and new commands call the same code.

The AWS and GCE commands will be updated in separate commits

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -40,6 +40,7 @@ from brkt_cli.esx import (
     rescue_metavisor_args,
     update_vmdk,
     update_encrypted_vmdk_args,
+    encrypt_with_esx_host_args,
 )
 from brkt_cli.validation import ValidationError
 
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -449,3 +373,182 @@ def command_rescue_metavisor(values, parsed_config, log):
...
+            formatter_class=brkt_cli.SortingHelpFormatter
+        )
+        encrypt_vmdk_args.setup_encrypt_vmdk_args(
+            encrypt_with_vcenter_parser)
+        setup_instance_config_args(encrypt_with_vcenter_parser, parsed_config)
+
+        encrypt_with_esx_parser = vmware_subparsers.add_parser(
+            'encrypt-with-esx-host',
+            description=(
+                'Create an encrypted VMDK on an ESX host'
```
apoorva-jindal: Create an encrypted VMDK from an existing VMDK on an ESX host

brkt_cli/esx/__init__.py
```diff
@@ -449,3 +373,182 @@ def command_rescue_metavisor(values, parsed_config, log):
...
+            ),
+            help='Encrypt a VMDK using vCenter',
+            formatter_class=brkt_cli.SortingHelpFormatter
+        )
+        encrypt_vmdk_args.setup_encrypt_vmdk_args(
+            encrypt_with_vcenter_parser)
+        setup_instance_config_args(encrypt_with_vcenter_parser, parsed_config)
+
+        encrypt_with_esx_parser = vmware_subparsers.add_parser(
+            'encrypt-with-esx-host',
```
apoorva-jindal: We should create new files for the command line arguments for encrypt-with-vcenter and encrypt-with-esx-host as some of the arguments for vcenter are not applicable to esx-host.

brkt_cli/esx/__init__.py
```diff
@@ -58,122 +58,29 @@ def _get_vcenter_password():
...
-            raise ValidationError("Cannot use template VMs for "
-                                  "updation on a single ESX host")
-        if (values.template_vm_name is None):
-            raise ValidationError("Encrypted image not provided")
+        if use_esx is False and values.template_vm_name is None:
+            raise ValidationError("Missing template-vm-name for the "
+                                  "template VM")
     if (values.source_image_path is not None and values.image_name is None):
         raise ValidationError("Specify the Metavisor OVF file.")
     _check_env_vars_set('VCENTER_USER_NAME')
```
krishnanrs: Do we have a good suggestive name? Alternative would be to use different environment variables depending on whether we are using VCENTER v/s ESX.

brkt_cli/esx/__init__.py
apoorva-jindal: The file looks good.

brkt_cli/esx/__init__.py
apoorva-jindal: Can we use different names for vcenter and esx?

krishnanrs: Then users would need to set 4 different environment variables (2 username and 2 password), depending on whether they are using vCenter or ESX (or both). Are we fine with that?

apoorva-jindal: Customers would be using either vcenter or esx host, not both. So, it is still two. Actually it is one because only username is mandatory, if password is not set, customer gets prompted for it.

apoorva-jindal:
LGTM

PR #494, opened 2016-11-15
Hide use-esx option

krishnanrs:
Hide the use-esx option as it is provided only for backward compatibility.
WIth the new command convention, the use-esx flag settings will get
implied automatically.


apoorva-jindal:
LGTM

PR #496, opened 2016-11-16
Traceback error when running rescue-metavisor

krishnanrs:
Traceback error reported when running rescue-metavisor. The _get_vcenter_password
method had been updated to accept one argument, but the invocation from the
rescue-metavisor method was not updated to reflect this change.


bburtin:
👍  interpreted languages FTW!

PR #497, opened 2016-11-18
Fixed usage for update-with-esx-host sub-command

krishnanrs:
- The positional VMDK argument was not required
 - Changed the OVF/OVA option to use update instead of create


Garpetun:
Looks good to me, but since I haven't really contributed to the CLI repo in any meaningful way I'll let someone else merge this.

At a naive first glance it looks like we could split out some of the common arguments, so we don't have to specify --esx-host in both encrypt_with_esx_host_args.py as well as update_with_esx_host_args.py.

apoorva-jindal:
LGTM

PR #498, opened 2016-11-19
Remove the OVF file after converting it to OVA.

apoorva-jindal:
Retaining the OVF file after encryption will cause update of the same ova to fail later as untarring it to create ovf will fail.


bburtin:
Can you please update your commit message and explain why you're making this change?

krishnanrs:
👍

PR #501, opened 2016-11-21
Log the hostname when auth fails

bburtin:
When the token fails authorization, log the name of the host.  This
makes the error a bit more intuitive for users who access multiple
environments.


krishnanrs:
👍

PR #502, opened 2016-11-21
Don't write debug log unless the command starts running

bburtin:
Don't write the debug log to a temp file when a validation error or
keyboard interrupt.  The temp debug log is only useful when diagnosing
unexpected errors during long-running operations.


krishnanrs:
👍

PR #503, opened 2016-11-21
Improve GCE encryption speed by ~8x

jimvoll:
It turns out that the IOPS available to a VM are based on
the aggregate (or perhaps largest) drive attached to the VM.
We are bottlenecked on GCE by the speed of the dd src dest
where both src and dest are smallish (10-20G) drives.

By attaching (but not using) a large drive to the encryptor
instance results in more IOPS (typically 10x) available for
both the src and dest drives.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -35,6 +35,14 @@ def setup_encryption(gce_svc,
         # of the unencrypted guest root
         log.info('Creating disk for encrypted image')
         gce_svc.create_disk(zone, encrypted_image_disk, guest_size * 2 + 1)
+        #
+        # Amazing but true - just attach a big drive to get more IOPS
+        # to be shared by all volumes attached to this VM. We don't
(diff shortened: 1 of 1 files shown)


brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -35,6 +35,14 @@ def setup_encryption(gce_svc,
...
         log.info('Creating disk for encrypted image')
         gce_svc.create_disk(zone, encrypted_image_disk, guest_size * 2 + 1)
+        #
+        # Amazing but true - just attach a big drive to get more IOPS
+        # to be shared by all volumes attached to this VM. We don't
+        # use the dummy-iops drive but get to use it's IOPS for other
+        # drives.
+        #
+        log.info('Creating dummy IOPS disk')
+        gce_svc.create_disk(zone, encrypted_image_disk + "dummy-iops", 500)
```
krishnanrs: Would be a good idea to use the append_suffix method here to ensure that we do not hit any length limitations here. Otherwise the change looks amazingly good! 👍

PR #504, opened 2016-11-21
Improve GCE encryption speed by 10x

jimvoll:
It turns out that the IOPS available to a VM are based on
the aggregate (or perhaps largest) drive attached to the VM.
We are bottlenecked on GCE by the speed of the dd src dest
where both src and dest are smallish (10-20G) drives.

By attaching (but not using) a large drive to the encryptor
instance results in more IOPS (typically 10x) available for
both the src and dest drives.


krishnanrs:
👍

PR #505, opened 2016-11-22
Don't allow "any" for claim name or value

bburtin:
We treat "any" as a reserved word.  Don't allow it for the claim name or
value.  Stop returning True from validate_name_value(), since we throw
ValidationError when validation fails.


krishnanrs:
👍 Remind me to make sure that this is documented in our user guide.

PR #506, opened 2016-11-22
Error when connecting to vCenter

krishnanrs:
CLI reports an error when connecting to vCenter. This happened since the verify flag was initialized in the parent class only. Verified locally and waiting on the automation tests to pass.


Garpetun:
Tested updating from this branch, works for me! I'll let someone else approve it since I haven't checked out the code base.

apoorva-jindal:
LGTM

PR #507, opened 2016-11-22
Dont delete encryptor if no-cleanup is set

robertabbott:


krishnanrs:
We should do this for image updates as well to make it consistent across the two operations. Otherwise LGTM.

PR #508, opened 2016-11-22
Fix disabling of verify certification

apoorva-jindal:
1. Pass the flag down to vcenter service initialization.
2. Set context to None in case --no-verify-cert is used.


krishnanrs:
👍

PR #509, opened 2016-11-22
Clean up make-user-data arguments and usage

bburtin:
Remove the following options from the make-user-data command:

--status-port
--brkt-env
--service-domain

The status port is only used during encryption.  brkt-env and service-
domain shouldn't be used when launching an instance, since we don't want
the same Metavisor talking to two different services.

Clean up the wording for several command line options, to make the
description clearer.

--------------
~~~~
$ ./brkt make-user-data -h
usage: brkt make-user-data [-h] [--ntp-server DNS_NAME]
                           [--proxy HOST:PORT | --proxy-config-file PATH]
                           [--token TOKEN] [-v]
                           [--guest-user-data-file PATH:TYPE]

Generate MIME multipart user-data that is passed to Metavisor and cloud-init
when running an instance.

optional arguments:
  --guest-user-data-file PATH:TYPE
                        User-data file and MIME content type to be passed to
                        cloud-init on the guest instance. Can be specified
                        multiple times. (default: None)
  --ntp-server DNS_NAME
                        NTP server to sync Metavisor clock. May be specified
                        multiple times. (default: None)
  --proxy HOST:PORT     Proxy that Metavisor uses to talk to the Bracket
                        service (default: None)
  --proxy-config-file PATH
                        proxy.yaml file that defines the proxy configuration
                        that metavisor uses to talk to the Bracket service
                        (default: None)
  --token TOKEN         Token (JWT) that Metavisor uses to authenticate with
                        the Bracket service. Use the make-token subcommand to
                        generate a token. (default: None)
  -h, --help            show this help message and exit
  -v, --verbose         Print status information to the console (default:
                        False)
~~~~

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -47,8 +47,8 @@ def setup_instance_config_args(parser, parsed_config,
         dest='ntp_servers',
         action='append',
         help=(
-            'Optional NTP server to sync Metavisor clock. '
-            'May be specified multiple times.'
+            'NTP server to sync Metavisor clock. May be specified multiple '
+            'times.'
         )
     )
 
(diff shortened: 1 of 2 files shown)


krishnanrs:
I agree on the status-port and service-domain completely. However can brkt-env be used to *add* services to the Metavisor by pointing to newly deployed pods/services? I will let someone more familiar in this respect to respond to that.

bburtin:
Sounds like a slippery slope.  If we were to support such a feature, we'd have to test it.

bonzotronix:
LGTM, FWIW.  Thanks for fixing!

PR #512, opened 2016-11-29
Handle omitted file ending

Garpetun:
If you supplied an encryptor ovf image without the .ovf file ending it
chopped off the last characters and you ended up with an invalid file
name. We now handle both cases.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -806,7 +806,10 @@ class VCenterService(BaseVCenterService):
         if validate_mf:
             # Load checksums for each file
             mf_checksum = None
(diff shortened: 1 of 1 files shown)


krishnanrs:
:+1

brkt_cli/esx/esx_service.py
```diff
@@ -806,7 +806,10 @@ def upload_ovf_to_vcenter(self, target_path, ovf_name,
         if validate_mf:
             # Load checksums for each file
             mf_checksum = None
-            mf_file_name = ovf_name[:ovf_name.find(".ovf")] + ".mf"
+            if '.ovf' in ovf_name:
```
bburtin: Looks like this will break if `.ovf` is in the middle of the filename.

bburtin:
👍

PR #514, opened 2016-11-29
Clean up the version check error

bburtin:
If an exception is raised during the version check and the exception
doesn't have a message, use the class name as the message.


krishnanrs:
👍

PR #515, opened 2016-11-29
Remove support for PV Metavisor

bburtin:
We no longer support the paravirtual version of Metavisor.  Remove the
--pv option and related code.


krishnanrs:
Docs also need updates. Otherwise LGTM.

bburtin:
Thanks Krishnan.  I'll update the docs when we ship the next release.

PR #516, opened 2016-11-30
Remove unused register_image code

bburtin:
We are now calling create_image() instead of register_image().  Remove
the unused application code.  We still use this function for registering
dummy images in the unit test code.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -111,14 +111,6 @@ class BaseAWSService(object):
         pass
 
     @abc.abstractmethod
-    def register_image(self,
-                       kernel_id,
-                       block_device_map,
-                       name=None,
-                       description=None):
-        pass
-
-    @abc.abstractmethod
     def get_image(self, image_id, retry=False):
         pass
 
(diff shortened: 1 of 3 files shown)


krishnanrs:
LGTM. For consistency sake, do we also want to change it to create_image in the tests?

bburtin:
Oh... good point.  Let me see if I can do that.  The `create_image()` and `register_image()` test code is pretty similar.

bburtin:
I take it back.  They're different operations.  In the AWS API, `CreateImage` creates an image from an instance.  `RegisterImage` creates an image from snapshots.  In the unit test setup code, we want `RegisterImage`, since we're creating dummy Metavisor and guest images for the tests to use.

krishnanrs:
So why write a test for a functionality that is unused? Also creates the confusion as to why the register_image is missing from the actual aws_svc, unless someone reads this commit description.

bburtin:
It gets weird because `create_image()` expects an instance ID and `register_image()` doesn't.  So it turns into a judgment call between using the correct function and having a few extra lines of code vs. hacking a different function to do something it's not meant to do.

PR #518, opened 2016-12-02
Add HTTP(s) proxy for S3 connections

krishnanrs:
A user may want to use a proxy to connect to the internet
from the host which is running the brkt-cli, while all other
(internal) connections do not require the proxy. The outside
(internet) connection is to S3 while the internal connections
goes to the local vCenter (which at most times has a well known
host name), to an ESX host on which the encryptor/updater VM
is deployed (whose IP most times is not known, but the list
of ESX hosts may available)  and to the encryptor VM itself
(whose IP comes from DHCP, hence is definitely not known
beforehand). This change allows the user to specify a HTTP(s)
proxy and port to allow the brkt-cli to use it for the
outbound connection to S3.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -17,6 +17,8 @@ import os
 import subprocess
 import getpass
 
+from brkt_cli import _parse_proxies
+
 from brkt_cli.subcommand import Subcommand
 
 from brkt_cli import (
@@ -102,6 +104,11 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
         _, brkt_env = parsed_config.get_current_env()
     if not values.token:
         raise ValidationError('Must provide a token')
+
+    proxies = None
+    if values.http_proxy:
+        proxies = _parse_proxies(values.http_proxy)
+
     # Download images from S3
     try:
         if (values.encryptor_vmdk is None and
@@ -109,7 +116,8 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
             (ovf, file_list) = \
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
-                    image_name=values.image_name
+                    image_name=values.image_name,
+                    proxy=proxies[0]
                 )
             if ovf is None:
                 raise ValidationError("Did not find MV OVF images")
@@ -267,6 +275,9 @@ def run_update(values, parsed_config, log, use_esx=False):
     if not values.token:
         raise ValidationError('Must provide a token')
 
+    proxies = None
+    if values.http_proxy:
+        proxies = _parse_proxies(values.http_proxy)
     # Download images from S3
     try:
         if (values.encryptor_vmdk is None and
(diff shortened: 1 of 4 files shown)


bburtin:
Seems like they should be able to use standard proxy environment variables to specify which hosts are internal and external.  Let's make sure these options are really necessary before merging.

bobveznat:
I'm with Boris, we should use the environment variables. boto3 seems to agree:

https://github.com/boto/boto3/issues/338

And then the trick with the env vars is knowing about the NO_PROXY var which you'll in particular need to set when you're out in AWS or GCE to whitelist 169.254.169.254.

apoorva-jindal:
I think it makes sense to state the problem in the commit message. A customer wants to use a proxy to connect to the outside internet on the host running brkt-cli, while all internal connections do not use the proxy. The outside connection is to S3, while the internal connections go to vcenter (which most times has a well known name), to the esx host on which the encryptor VM is deployed (whose IP most times is not known, but the list of ESX hosts may be available) and to the encryptor VM itself (whose IP comes from DHCP, hence is definitely not known). 

Using the proxy environment variables work if all the IPs/names are known up front and are provided to the NO_PROXY variable. Unfortunately, since the encryptor VM IP (and most times the ESX host IP) is not known, using NO_PROXY does not work.

brkt_cli/esx/update_encrypted_vmdk_args.py
```diff
@@ -137,6 +137,24 @@ def setup_update_vmdk_args(parser):
         default=False,
         help=argparse.SUPPRESS
     )
+    # Optional HTTP Proxy argument which can be used in proxied environments
+    # Specifies the HTTP Proxy to use for S3/AWS connections
+    parser.add_argument(
+        '--http-proxy',
```
apoorva-jindal: We have a --proxy variable which gets put in the encryptor VM. Can we change to --http-s3-proxy to explicitly indicate this is for s3 access? Also, can we combine the two options into one similar to the --proxy?

krishnanrs:
Verified that if a http proxy is not specified (when required for the S3 download) or an invalid host/port is specified, then it fails with a timeout:

timeout: timed out
('Failed to download MV image from S3: ', timeout('timed out',))

brkt_cli/esx/__init__.py
```diff
@@ -267,14 +275,18 @@ def run_update(values, parsed_config, log, use_esx=False):
     if not values.token:
         raise ValidationError('Must provide a token')
 
+    proxies = None
+    if values.http_proxy:
+        proxies = _parse_proxies(*values.http_proxy)
```
apoorva-jindal: Do we need a * in front of values.http_proxy here?

krishnanrs: No. Because the argument here is parsed as a string, not a list. I was just reusing the _parse_proxies method which parses a list of proxies which can be specified for the MV.

apoorva-jindal: Thanks, then please remove it.

apoorva-jindal:
LGTM

PR #519, opened 2016-12-07
Provide ability to add SSD scratch disks with instance launch

krishnanrs:
GCE instances support local SSD scratch disks that are physically
attached to the server which hosts the virtual machine (similar to
instance store volumes on AWS). However unlike the persistent disks,
the SSD scratch disks can only be attached at instance launch time.
This change allows the user to specify the number of SSD scratch
disks to attach to the instance at launch time (Default: 0, Max: 8)
using the --ssd-scratch-disks option with the "gce launch" sub-command.

### DIFF
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -113,6 +113,8 @@ def run_update(values, config):
 
 def run_launch(values, config):
     gce_svc = gce_service.GCEService(values.project, None, log)
+    if values.ssd_scratch_disks > 8:
+        raise ValidationError("Maximum of 8 SSD scratch disks are supported")
     instance_config = instance_config_from_values(
         values, mode=INSTANCE_METAVISOR_MODE, cli_config=config)
     if values.startup_script:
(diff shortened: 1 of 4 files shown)


brkt_cli/gce/__init__.py
```diff
@@ -113,6 +113,8 @@ def run_update(values, config):
 
 def run_launch(values, config):
     gce_svc = gce_service.GCEService(values.project, None, log)
+    if values.ssd_scratch_disks > 8:
+        raise ValidationError("Maximum of 8 SSD scrach disks are supported")
```
bburtin: typo

brkt_cli/gce/launch_gce_image_args.py
```diff
@@ -68,3 +68,13 @@ def setup_launch_gce_image_args(parser):
...
     )
+    # Optional (number of) SSD scratch disks because these can only be attached
+    # at instance launch time, compared to the other (persistent) disks
+    parser.add_argument(
+        '--ssd-scratch-disks',
+        metavar='N',
+        type=int,
+        default=0,
+        dest='ssd_scratch_disks',
+        help=argparse.SUPPRESS
```
bburtin: Why are we hiding this option?  It isn't generally useful?

krishnanrs: Fixed. Let me know if the help message looks OK.

PR #520, opened 2016-12-07
Fix error when no S3 Proxy is specified

krishnanrs:
If the http-s3-proxy argument is not defined when encrypting or
updating a vmware image, then the command was failing with an
object attribute error. The change ensures that the proxy
elements are passed only if they are specified, else it is
skipped.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -117,7 +117,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
                     image_name=values.image_name,
-                    proxy=proxies[0]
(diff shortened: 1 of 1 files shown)


jntkflow:
+2

brkt_cli/esx/__init__.py
```diff
@@ -117,7 +117,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
                     image_name=values.image_name,
-                    proxy=proxies[0]
```
bburtin: This seems indirect.  What about doing this above:

```
proxy = None
if values.http_proxy:
  proxy = _parse_proxies(values.http_proxy)[0]
```

This avoids having to play games with if/else later.

PR #521, opened 2016-12-07
Fix error when no S3 Proxy is specified

krishnanrs:
If the http-s3-proxy argument is not defined when encrypting or
updating a vmware image, then the command was failing with an
object attribute error. The change ensures that the proxy
elements are passed only if they are specified, else it is
skipped.


krishnanrs:
@bburtin please review again.

brkt_cli/esx/__init__.py
```diff
@@ -107,7 +107,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
 
     proxies = None
     if values.http_proxy:
-        proxies = _parse_proxies(values.http_proxy)
+        proxies = _parse_proxies(values.http_proxy)[0]
```
bburtin: I believe this is just one object?  If so, the variable name should be singular, not plural.

PR #522, opened 2016-12-07
Fix error when no S3 Proxy is specified for VMWare updates

krishnanrs:
Following up from the previous commit, adding the changes to
ignore the proxy for VMWare updates if it is not specified.


krishnanrs:
@bburtin please review this one. I closed the other pull request.

PR #523, opened 2016-12-08
Close debug log handler before deleting the file

bburtin:
Close the debug log handler before attempting to remove the file.  This
avoids open file descriptor issues on Windows.


krishnanrs:
👍

PR #525, opened 2016-12-12
Hide the encryptor-image-bucket argument

krishnanrs:
The encryptor-image-bucket argument needs to be specified if you
are using a custom (or pre-production) image for encrypting or
updating an image. Since this is required only for internal
testing, this argument is being hidden.


krishnanrs:
Not sure whether we should leave this argument exposed or hide it, although this argument is used only for internal/dev testing.

PR #526, opened 2016-12-12
Add support for setting the network-name for the Encryptor VM

apoorva-jindal:
1. Add support for changing the network name to which the
NIC in the Encryptor VM connects to.
2. Add support for allowing Encryptor VM NIC to be a
VirtualPort or a VirtualPortGroup in a VDS.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -77,7 +77,7 @@ class BaseVCenterService(object):
     def __init__(self, host, user, password, port,
                  datacenter_name, datastore_name, esx_host,
                  cluster_name, no_of_cpus, memoryGB, session_id,
-                 verify=True):
+                 network_name, nic_type, verify=True):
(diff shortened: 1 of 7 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -445,7 +447,7 @@ def create_vm(self, memoryGB=1, numCPUs=1,
         n_intf = vim.vm.device.VirtualVmxnet3()
         n_intf.key = -1
         n_intf.backing = vim.vm.device.VirtualEthernetCard.NetworkBackingInfo()
-        n_intf.backing.deviceName = network_name
+        n_intf.backing.deviceName = self.network_name
```
krishnanrs: Since the network name is optional, what happens if that is not specified?

apoorva-jindal: It will be set to its default value which is "VM Network".

PR #527, opened 2016-12-13
make-token: make the signing key a positional argument

bburtin:
Make the signing key path a positional argument, since it's required and
the token is derived from the key.  Hide the old --signing-key option
and log a deprecation warning when it's used.

```
$ ./brkt make-token -h
usage: brkt make-token [-h] [--claim NAME=VALUE] [--customer UUID]
                       [--exp TIMESTAMP] [--nbf TIMESTAMP] [-v]
                       [SIGNING-KEY-PATH]

Generate a JSON Web Token for encrypting an instance or launching an encrypted
instance. A timestamp can be either a Unix timestamp in seconds or ISO 8601
(2016-05-10T19:15:36Z). Timezone offset defaults to UTC if not specified.

positional arguments:
  SIGNING-KEY-PATH    The private key that is used to sign the JWT. The key
                      must be a 384-bit ECDSA private key (NIST P-384) in PEM
                      format.

optional arguments:
  --claim NAME=VALUE  JWT claim specified by name and value. May be specified
                      multiple times.
  --customer UUID     Required for API access when using a third party JWK
                      server
  --exp TIMESTAMP     Token expiration time
  --nbf TIMESTAMP     Token is not valid before this time
  -h, --help          show this help message and exit
  -v, --verbose       Print status information to the console
```

### DIFF
--- a/brkt_cli/brkt_jwt/__init__.py
+++ b/brkt_cli/brkt_jwt/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from __future__ import print_function
 
+import argparse
 import json
 import logging
 import re
@@ -47,7 +48,17 @@ class MakeTokenSubcommand(Subcommand):
         return values.make_jwt_verbose
 
     def run(self, values):
-        crypto = util.read_private_key(values.signing_key)
+        if values.signing_key_option:
+            log.warn(
+                'The --signing-key option is deprecated and will be removed '
+                'in a future release.'
+            )
+
+        signing_key = values.signing_key or values.signing_key_option
+        if not signing_key:
+            raise ValidationError('Signing key path was not specified')
+
+        crypto = util.read_private_key(signing_key)
         exp = None
         if values.exp:
             exp = parse_timestamp(values.exp)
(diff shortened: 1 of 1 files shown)


brkt_cli/brkt_jwt/__init__.py
```diff
@@ -213,11 +224,23 @@ def setup_make_jwt_args(subparsers):
...
+            'Timezone offset defaults to UTC if not specified.'),
+        help=(
+            'Generate a JSON Web Token for encrypting or launching an '
+            'instance'),
         formatter_class=brkt_cli.SortingHelpFormatter
     )
+
+    # TODO: Make this a required argument when we remove the --signing-key
+    # option.
+    parser.add_argument(
```
krishnanrs: Isn't this argument mandatory? With the current definition, it appears like this argument is optional.

bburtin: It's optional in the short term, to maintain backward compatibility.  I don't want to risk breaking existing scripts that don't specify the positional argument.  Once we remove the legacy option, I'll make the positional argument required.  In the meantime, I raise `ValidationError` if neither argument is specified.

PR #528, opened 2016-12-22
Delete boot disk by default for GCE instances

krishnanrs:
By default, we were not deleting the boot disk for a GCE instance
launched using the "brkt gce launch" command, when the instance was
terminated. This change makes the boot disk deletion the default
behaviour and toggles the purpose of the --delete-boot argument.
Now if this argument is specified, then we will retain the boot
disk on instance termination.

### DIFF
--- a/brkt_cli/gce/gce_service.py
+++ b/brkt_cli/gce/gce_service.py
@@ -540,7 +540,7 @@ class GCEService(BaseGCEService):
                      network='default',
                      disks=[],
                      metadata={},
-                     delete_boot=False,
+                     delete_boot=True,
                      block_project_ssh_keys=False,
                      instance_type='n1-standard-4',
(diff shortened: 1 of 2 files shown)


bburtin:
Seems unintuitive that `--delete-boot` causes us to *not* delete the boot disk.  Would it be better to deprecate `--delete-boot`, make it a no-op, and add a new `no-delete-boot` option?

krishnanrs:
I'd rather prefer that instead of having two contradictory options, we simply remove the old one and create the new one with the right name/format. Will notify the rest of the team of the pending change so that they can update any automation scripts accordingly.

bburtin:
Ok, sounds good.  Just wanted to make sure that you took backward compatibility into consideration.

PR #530, opened 2016-12-27
Don't auto-delete boot disk for encryptor/updater

jimvoll:
The orchestration relies on the boot disk being around after
the updater/encryptor has terminated.

Minor fix to avoid getting console output after updater
has terminated.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -71,6 +71,7 @@ def do_encryption(gce_svc,
                          disks=[gce_svc.get_disk(zone, instance_name),
                                 gce_svc.get_disk(zone, encrypted_image_disk),
                                 gce_svc.get_disk(zone, dummy_name)],
+                         delete_boot=False,
                          metadata=metadata)
 
     try:
--- a/brkt_cli/gce/update_gce_image.py
+++ b/brkt_cli/gce/update_gce_image.py
@@ -70,6 +70,7 @@ def update_gce_image(gce_svc, enc_svc_cls, image_id, encryptor_image,
                              network=network,
                              subnet=subnetwork,
                              disks=[],
+                             delete_boot=False,
                              metadata=user_data)
         ip = gce_svc.get_instance_ip(updater, zone)
         updater_launched = True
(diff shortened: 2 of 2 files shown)


brkt_cli/gce/update_gce_image.py
```diff
@@ -89,6 +90,7 @@ def update_gce_image(gce_svc, enc_svc_cls, image_id, encryptor_image,
         # delete updater instance
         log.info('Deleting updater instance')
         gce_svc.delete_instance(zone, updater)
+        updater_launched = False
```
krishnanrs: Do we really need this? If we set this bool here, then in the event of an error during the update process, the console logs from the updater instance will not be captured.

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -71,6 +71,7 @@ def do_encryption(gce_svc,
                          disks=[gce_svc.get_disk(zone, instance_name),
                                 gce_svc.get_disk(zone, encrypted_image_disk),
                                 gce_svc.get_disk(zone, dummy_name)],
+                         delete_boot=False,
```
krishnanrs: Instead of explicitly settings this for encrypt_gce_image and update_gce_image, can we set the default as False in the gce_service? For the launch command we are anyways defaulting delete_boot to True so that should not be impacted?

brkt_cli/gce/update_gce_image.py
jimvoll: This is exactly the problem though. In the event of an error, the instance is already deleted at this point and getting the console output results in another exception.

brkt_cli/gce/encrypt_gce_image.py
jimvoll: Either way. I can change it if you want.

brkt_cli/gce/update_gce_image.py
jimvoll: This is the problem Neha reported, so I fixed that issue while I was fixing the other one

krishnanrs:
Got it. I leave it to you to decide if you want to submit this as is. 👍

PR #532, opened 2017-01-06
MV-257: Add crypto policy as an option in brkt cli

ramyabrkt:
crypto option can take xts, gcm and it is an optional argument

Testing:
1. gce freebsd with xts, gcm and default(gcm)
2. aws freebsd with xts, gts and default(gcm)

This review is dependent a corresponding review in freebsd code base.
would like to commit that first and then this can be merged

Jira: MV-257

### DIFF
--- a/brkt_cli/aws/encrypt_ami_args.py
+++ b/brkt_cli/aws/encrypt_ami_args.py
@@ -130,13 +130,9 @@ def setup_encrypt_ami_args(parser):
     parser.add_argument(
         '--crypto',
         dest='crypto',
-<<<<<<< HEAD
         required=False,
         help=argparse.SUPPRESS,
         default='gcm'
-=======
-        required=False
->>>>>>> d053007937ba6b5451c9f37e35b5e84c850131bd
     )
 
     # Optional argument for development: if encryption fails, keep the
--- a/brkt_cli/aws/encrypt_ami_args.py
+++ b/brkt_cli/aws/encrypt_ami_args.py
@@ -127,6 +127,14 @@ def setup_encrypt_ami_args(parser):
         default=False
     )
 
+    parser.add_argument(
+        '--crypto',
+        dest='crypto',
+        required=False,
+        help=argparse.SUPPRESS,
+        default='gcm'
+    )
+
     # Optional argument for development: if encryption fails, keep the
     # encryptor running so that we can debug it.
     parser.add_argument(
--- a/brkt_cli/gce/__init__.py
+++ b/brkt_cli/gce/__init__.py
@@ -32,6 +32,13 @@ def run_encrypt(values, config):
     gce_svc = gce_service.GCEService(values.project, session_id, log)
     check_args(values, gce_svc, config)
 
+    instance_config = instance_config_from_values(
+            values, mode=INSTANCE_CREATOR_MODE, cli_config=config)
+    if instance_config.brkt_config and 'crypto' in instance_config.brkt_config:
+	crypto_policy = values.crypto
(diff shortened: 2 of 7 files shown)


brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -127,6 +127,12 @@ def setup_encrypt_ami_args(parser):
         default=False
     )
 
+    parser.add_argument(
```
krishnanrs: we should add some help context here so that the user knows what to enter

brkt_cli/gce/__init__.py
```diff
@@ -32,6 +32,13 @@ def run_encrypt(values, config):
     gce_svc = gce_service.GCEService(values.project, session_id, log)
     check_args(values, gce_svc, config)
 
+    instance_config = instance_config_from_values(
```
krishnanrs: Should add a validation method for the values so that we can simply accept it blindly post validation, instead of having to compare upper/lower case strings

brkt_cli/aws/encrypt_ami_args.py
krishnanrs: Also, if we are going to default to GCM, we can set a default value here so that we can ensure that there is always a value associated with this argument, instead of having to check whether it is defined or not.

brkt_cli/gce/__init__.py
jimvoll: Why is crypto policy upper case for GCE?

bonzotronix:
That merge commit shouldn't be there, right?  Maybe rebase the branch?

brkt_cli/aws/encrypt_ami_args.py
ramyabrkt: sounds good

brkt_cli/gce/__init__.py
ramyabrkt: the check for xts or gcm crypto policy converts to lowercase anyways. however will change it to lower case for consistency. no reason for upper case here

brkt_cli/aws/encrypt_ami_args.py
bonzotronix: Shouldn't this new option be 'hidden' (i.e., help=argparse.SUPPRESS)?
And can we think of a more specific arg name than "crypto"?

brkt_cli/instance_config_args.py
```diff
@@ -176,6 +175,9 @@ def instance_config_from_values(values=None, mode=INSTANCE_CREATOR_MODE,
     if values.ntp_servers:
         brkt_config['ntp_servers'] = values.ntp_servers
 
+    if values and 'crypto' in values:
+        brkt_config['crypto'] = values.crypto
+
```
bonzotronix: This field should not be part of instance_config.  The instance_config settings should be independent of the command/operation.  This new setting should really be passed in a new dict named something like "encrypt_params", in the same way that on-prem image encryption passes its storage-related settings in a "create_from_storage" dict.

brkt_cli/gce/__init__.py
ramyabrkt: done

brkt_cli/aws/encrypt_ami_args.py
ramyabrkt: hidden or not was getting debated when i wrote this. its decided now to have this hidden

ramyabrkt:
Another pull request has been created with same contents with review comments addressed + some tests.

PR #534, opened 2017-01-07
Provide ability to specify crypto mode

krishnanrs:
Bracket supports GCM and XTS encryption mode for the disks. By default
the guest root volume always was encrypted using GCM. This change
adds a new (hidden) --boot-disk-crypto-policy argument which allows
the user to override the default encryption policy for the (guest)
root volume. If 'default' is specified, then the default encryption
policy is determined from the encryptor image type (NetBSD vs FreeBSD).
Note that XTS is not supported with the NetBSD encryptor.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -39,14 +39,18 @@ from brkt_cli.aws.encrypt_ami import (
 from brkt_cli.aws.update_ami import update_ami
 from brkt_cli.instance_config import (
     INSTANCE_CREATOR_MODE,
-    INSTANCE_UPDATER_MODE
+    INSTANCE_UPDATER_MODE,
 )
 from brkt_cli.instance_config_args import (
     instance_config_from_values,
     setup_instance_config_args
 )
 from brkt_cli.subcommand import Subcommand
-from brkt_cli.util import BracketError
+from brkt_cli.util import (
+    BracketError,
+    CRYPTO_GCM,
+    CRYPTO_XTS
+)
 from brkt_cli.validation import ValidationError
 
 log = logging.getLogger(__name__)
@@ -227,6 +231,27 @@ def run_encrypt(values, config, verbose=False):
         _validate(aws_svc, values, encryptor_ami)
         brkt_cli.validate_ntp_servers(values.ntp_servers)
 
+    mv_image = aws_svc.get_image(encryptor_ami)
+    if values.crypto is None:
+        if mv_image.name.startswith('brkt-avatar-freebsd'):
+            crypto_policy = CRYPTO_XTS
+        elif mv_image.name.startswith('brkt-avatar'):
+            crypto_policy = CRYPTO_GCM
+        else:
+            log.warn(
+                "Unable to determine encryptor type for image %s. "
+                "Default boot volume crypto policy to %s",
+                mv_image.name, CRYPTO_XTS
+            )
+            crypto_policy = CRYPTO_XTS
+    else:
+        crypto_policy = values.crypto
+        if crypto_policy == CRYPTO_XTS and not mv_image.name.startswith('brkt-avatar-freebsd'):
+            raise ValidationError(
+                'Unsupported crypto policy %s for encryptor %s' %
+                crypto_policy, mv_image.name
+            )
+
     instance_config = instance_config_from_values(
         values, mode=INSTANCE_CREATOR_MODE, cli_config=config)
     if verbose:
@@ -243,6 +268,7 @@ def run_encrypt(values, config, verbose=False):
         image_id=guest_image.id,
         encryptor_ami=encryptor_ami,
         encrypted_ami_name=values.encrypted_ami_name,
+        crypto_policy=crypto_policy,
         subnet_id=values.subnet_id,
         security_group_ids=values.security_group_ids,
         guest_instance_type=values.guest_instance_type,
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -59,7 +59,9 @@ from brkt_cli.util import (
     Deadline,
     make_nonce,
     sleep,
-    append_suffix)
+    append_suffix,
+    CRYPTO_XTS
+)
 from datetime import datetime
 
 # End user-visible terminology.  These are resource names and descriptions
@@ -325,13 +327,14 @@ def create_encryptor_security_group(aws_svc, vpc_id=None, status_port=\
 
 def _run_encryptor_instance(
         aws_svc, encryptor_image_id, snapshot, root_size, guest_image_id,
-        security_group_ids=None, subnet_id=None, zone=None,
+        crypto_policy, security_group_ids=None, subnet_id=None, zone=None,
         instance_config=None,
         status_port=encryptor_service.ENCRYPTOR_STATUS_PORT):
     bdm = BlockDeviceMapping()
 
     if instance_config is None:
         instance_config = InstanceConfig()
+    instance_config.brkt_config['crypto_policy_type'] = crypto_policy
 
     # Use gp2 for fast burst I/O copying root drive
     guest_unencrypted_root = EBSBlockDeviceType(
@@ -345,7 +348,10 @@ def _run_encryptor_instance(
     guest_encrypted_root = EBSBlockDeviceType(
         volume_type='gp2',
         delete_on_termination=True)
-    guest_encrypted_root.size = 2 * root_size + 1
+    if crypto_policy == CRYPTO_XTS:
+        guest_encrypted_root.size = root_size + 1
+    else:
+        guest_encrypted_root.size = 2 * root_size + 1
 
     # Use 'sd' names even though AWS maps these to 'xvd'
     # The AWS GUI only exposes 'sd' names, and won't allow
@@ -898,7 +904,7 @@ def register_ami(aws_svc, encryptor_instance, encryptor_image, name,
     return ami_info
 
 
-def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
+def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
             encrypted_ami_name=None, subnet_id=None, security_group_ids=None,
             guest_instance_type='m3.medium', instance_config=None,
             save_encryptor_logs=True,
@@ -935,6 +941,7 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
                  "preserved because the root disk is attached at %s "
                  "instead of /dev/sda1", guest_image.root_device_name)
         legacy = True
+
     try:
         guest_instance = run_guest_instance(aws_svc,
             image_id, subnet_id=subnet_id, instance_type=guest_instance_type)
@@ -959,6 +966,7 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
             snapshot=snapshot_id,
             root_size=size,
             guest_image_id=image_id,
+            crypto_policy=crypto_policy,
             security_group_ids=security_group_ids,
             subnet_id=subnet_id,
             zone=guest_instance.placement,
--- a/brkt_cli/aws/encrypt_ami_args.py
+++ b/brkt_cli/aws/encrypt_ami_args.py
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 import argparse
+from brkt_cli.util import (
+    CRYPTO_GCM,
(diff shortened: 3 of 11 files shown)


krishnanrs:
I'm trying to see if I can add more tests. In the meantime, if there are any comments, let me know.

krishnanrs:
I've added tests to validate:
1. Error is raised for an invalid crypto option
2. Confirm if the user_data contains the crypto policy

brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -136,3 +136,12 @@ def setup_encrypt_ami_args(parser):
         default=True,
         help=argparse.SUPPRESS
     )
+    # Optional argument for root disk crypto policy. The supported values
+    # currently are "gcm" and "xts" with "gcm" being the default
+    parser.add_argument(
```
bburtin: How about adding a `choices` argument here?

brkt_cli/aws/__init__.py
```diff
@@ -744,6 +746,12 @@ def _validate(aws_svc, values, encryptor_ami_id):
     if values.encrypted_ami_name:
         aws_service.validate_image_name(values.encrypted_ami_name)
 
+    if values.crypto.lower() not in SUPPORTED_CRYPTO:
```
bburtin: I think this code won't be necessary if you add `choices` to the argument spec.

brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -136,3 +136,12 @@ def setup_encrypt_ami_args(parser):
...
+        '--boot-disk-crypto-policy',
```
bburtin: The name is a bit verbose.  Are we considering using different crypto policy for different disks?

brkt_cli/aws/test_aws.py
```diff
@@ -42,6 +43,7 @@ def __init__(self):
         self.proxies = []
         self.proxy_config_file = None
         self.status_port = None
+        self.crypto = SUPPORTED_CRYPTO[0]
```
bburtin: Would it make sense to add `CRYPTO_GCM` and `CRYPTO_XTS` constants?  The person reading the code won't have to look at the source to figure out what `[0]` is.

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -113,10 +118,10 @@ def create_image(gce_svc, zone, encrypted_image_disk, encrypted_image_name, encr
 
 
 def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
-            encrypted_image_name, zone, instance_config, image_project=None,
-            keep_encryptor=False, image_file=None, image_bucket=None,
-            network=None, subnetwork=None, status_port=ENCRYPTOR_STATUS_PORT,
-            cleanup=True):
+            encrypted_image_name, zone, instance_config, crypto_policy,
```
bburtin: Should we default this to `gcm`, to keep the callsites more concise?  I guess it depends on how common it will be to use XTS.

test_gce.py
```diff
@@ -275,6 +278,20 @@ def test_cleanup_on_fail(self):
         self.assertEqual(len(gce_svc.disks), 0)
         self.assertEqual(len(gce_svc.instances), 0)
 
+#    def test_invalid_crypto_policy(self):
```
bburtin: Cruft?

brkt_cli/aws/encrypt_ami_args.py
krishnanrs: Agreed. Testing out with suggested changes.

brkt_cli/aws/encrypt_ami_args.py
krishnanrs: At this point in time, the brkt-cli only encrypts the guest boot volume, which is why I kept it as verbose as possible. If anyone else has other suggestions, then I would be happy to incorporate the same.

brkt_cli/aws/__init__.py
krishnanrs: Agreed. Testing out with suggested changes.

brkt_cli/aws/test_aws.py
krishnanrs: Agreed. Testing out with suggested changes.

brkt_cli/gce/encrypt_gce_image.py
krishnanrs: If we setup the argument with choices and to always contain a default value, then we are assured that the crypto_policy would be always set to some value. No need to explicitly set it to `gcm` here in that case?

test_gce.py
krishnanrs: Yes. Will be removed. Also with the changes, I see no need to add validation tests for the crypto policy choices. Will only add a test to validate that the crypto policy is passed in the instance user data.

brkt_cli/instance_config_args.py
```diff
@@ -176,6 +176,9 @@ def instance_config_from_values(values=None, mode=INSTANCE_CREATOR_MODE,
     if values.ntp_servers:
         brkt_config['ntp_servers'] = values.ntp_servers
 
+    if 'crypto' in values:
+        brkt_config['crypto_policy_type'] = values.crypto
+
```
bonzotronix: I'm hoping we can leave it to the encryptor function to set this, and keep it out of instance_config (as this is really a setting pertaining to the encryption, not pertaining to how the instance behaves).

brkt_cli/instance_config.py
```diff
@@ -39,6 +39,10 @@
 INSTANCE_CREATOR_MODE   = 2
 INSTANCE_UPDATER_MODE   = 3
 
+# Supported crypto options for the disks
+CRYPTO_GCM = 'gcm'
+CRYPTO_XTS = 'xts'
```
bonzotronix: Could we move this to brkt_cli/__init__.py or maybe util.py?

brkt_cli/aws/encrypt_ami_args.py
```diff
@@ -136,3 +140,13 @@ def setup_encrypt_ami_args(parser):
...
+        dest='crypto',
```
bonzotronix: Stylistic nit: I'd tend to make the internal name for this arg as specific as the user-visible name: boot_disk_crypto_policy.

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -132,19 +143,31 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
...
             keep_encryptor = True
 
+        if crypto_policy == CRYPTO_XTS and 'metavisor' not in encryptor_image.name:
+            raise ValidationError(
+                'Unsupported crypto policy %s for encryptor %s' %
+                crypto_policy, encryptor_image.name
+            )
+
+        if crypto_policy == 'default':
+            if 'metavisor' not in encryptor_image.name:
```
bonzotronix: Is it possible someone will have a NetBSD image with 'metavisor' in its name?

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -132,19 +143,31 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
             # Keep user provided encryptor image
             keep_encryptor = True
 
+        if crypto_policy == CRYPTO_XTS and 'metavisor' not in encryptor_image.name:
+            raise ValidationError(
+                'Unsupported crypto policy %s for encryptor %s' %
+                crypto_policy, encryptor_image.name
```
bonzotronix: This looks OK to me.  We could mention NetBSD somehow, but no biggie.

brkt_cli/gce/encrypt_gce_image.py
krishnanrs: All bracket `published` images will follow the naming convention. However it is possible for users building and using their own encryptor images where the name may not follow convention. However in that case, the user probably already know what they need to do.

brkt_cli/aws/encrypt_ami_args.py
krishnanrs: Would make long variable names that has to be passed. Unless you are very particular about this, I would like to take a pass on this one.

bonzotronix: OK, no prob

brkt_cli/aws/encrypt_ami_args.py
bburtin: I was thinking just `--crypto-policy` would be enough, unless we really want to specify different policy for different disks.

brkt_cli/gce/encrypt_gce_image.py
bburtin: Ok, that works.  It was mostly an emotional reaction to the number of callsites that were affected.  I'm fine either way.

ramyabrkt:
looks good to me

brkt_cli/aws/encrypt_ami_args.py
krishnanrs: Updated per suggestion

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -99,3 +103,13 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
...
     )
+    # Optional argument for root disk crypto policy. The supported values
+    # currently are "gcm" and "xts" with "gcm" being the default
+    parser.add_argument(
+        '--crypto-policy',
+        dest='crypto',
+        metavar='NAME',
+        choices=['default', CRYPTO_GCM, CRYPTO_XTS],
+        help=argparse.SUPPRESS,
+        default='default'
```
bburtin: `'default'` is a little funky.  Seems like it would be clearer to remove the default and handle `None` during argument parsing?

brkt_cli/aws/encrypt_ami.py
```diff
@@ -345,7 +350,10 @@ def _run_encryptor_instance(
     guest_encrypted_root = EBSBlockDeviceType(
         volume_type='gp2',
         delete_on_termination=True)
-    guest_encrypted_root.size = 2 * root_size + 1
+    if crypto_policy == 'xts':
```
bburtin: Use the constant here?

bonzotronix:
LGTM

brkt_cli/aws/encrypt_ami.py
```diff
@@ -935,6 +943,26 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami,
...
                  "instead of /dev/sda1", guest_image.root_device_name)
         legacy = True
+
+    if crypto_policy == CRYPTO_XTS and not mv_image.name.startwith('brkt-avatar-freebsd'):
+        raise ValidationError(
+            'Unsupported crypto policy %s for encryptor %s' %
+            crypto_policy, mv_image.name
+        )
+
+    if crypto_policy == 'default':
```
bburtin: Can we establish the crypto policy and do the validation during command line parsing?  I think that would divide responsibility more clearly.  That way, when you call `encrypt()`, it just does what it's told, as opposed to switching behavior and doing validation.

brkt_cli/gce/encrypt_gce_image_args.py
krishnanrs: `default` is a supported option in the MV although the CLI updates it to one of the supported values. in other words, even if this escapes validation in the CLI, the MV would still try to do the correct thing. Using `None` makes it seem passe`

brkt_cli/aws/encrypt_ami.py
krishnanrs: If I do that, then I will have to retrieve the encryptor AMI details during the command line parsing, instead of doing it as part of the encryption method. Is that OK with you?

bburtin: Sounds fine to me.  This is validation logic, and validation should ideally happen before we call the application code.

brkt_cli/gce/encrypt_gce_image_args.py
bburtin: I see two separate issues: Python code structure and Metavisor behavior.  From the Python side, a default of `None` means that the user didn't say what the crypto policy should be.  That's the usual convention for the CLI spec.  It's unusual to be passing around the string `"default"`.  I'd expect us to pass around a `None` value, and then turn it into `"default"` at the very end, if that's what Metavisor wants.

brkt_cli/aws/encrypt_ami.py
krishnanrs: I moved the validation logic for AWS. But for GCE, it will require major surgery.

PR #535, opened 2017-01-11
Use correct attribute to identify encryptor type

krishnanrs:
For GCE, it is the `image_name` which determines the encryptor
type (NetBSD vs FreeBSD). However the `image_name` would vary
depending on how it was determined:
 - User provided image file: whatever the user provided
 - Published images in public buckets: the corresponding image name
 - Default latest image: always returned as `latest.tar.gz`

The last case above is the most common use case where users will
simply used the `released` latest image to encrypt their GCE image.
For this case, the boot volume crypto policy would be set to 
`default` allowing the encryptor to determine the policy itself.
In all other case, the crypto policy will be set based on the image
name.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image.py
+++ b/brkt_cli/gce/encrypt_gce_image.py
@@ -12,7 +12,6 @@ from brkt_cli.encryptor_service import (
 from brkt_cli.gce.gce_service import gce_metadata_from_userdata
 from brkt_cli.util import Deadline, retry, append_suffix
 from googleapiclient import errors
-from brkt_cli.validation import ValidationError
 from brkt_cli.util import (
     CRYPTO_XTS,
     CRYPTO_GCM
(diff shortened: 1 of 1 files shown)


bburtin:
Code deletion makes me happy.  Looks good to me, but let's get at least one more +1 from others who were involved in this decision.

brkt_cli/gce/encrypt_gce_image.py
```diff
@@ -143,24 +142,10 @@ def encrypt(gce_svc, enc_svc_cls, image_id, encryptor_image,
             # Keep user provided encryptor image
             keep_encryptor = True
 
+        # For GCE there is no way to identify the encryptor type
+        # Default to GCM
```
bonzotronix: I thought the only ambiguous case is when the GCE encryptor is from a *released* image.  Wrong?

krishnanrs: No, two possible cases: from a `released` image as well as a `dev`/`self` generated encryptor image.

PR #538, opened 2017-01-17
Default guest root crypto mode to XTS for GCE

krishnanrs:
Default the guest root volume crypto mode for GCE to XTS. This keeps
it consistent with the AWS encryption process.


bonzotronix:
LGTM

PR #539, opened 2017-01-17
Adding share-logs to GCE

markvoll:
This creates a new instance, attaches the metavisors root disk,
uploads all files found in /var to a google storage (bucket,file)
and shares the result with a specified account

Usage: brkt gce share-logs --instance <metavisor instance>
 --zone <zone> --project <project> --bucket <bucket>
 --file <file> --account <account>


brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+def share_logs(values, gce_svc):
+    snap = None
+    snapshot_name = 'brkt-diag-snapshot-' + gce_svc.get_session_id()
+    instance_name = 'brkt-diag-instance-' + gce_svc.get_session_id()
+    disk_name = 'sdb-' + gce_svc.get_session_id()
+    log.info('Sharing logs')
+
+    try:
+        image_project = 'ubuntu-os-cloud'
+        ubuntu_image = 'ubuntu-1404-trusty-v20160222'
```
krishnanrs: GCE frequently obsoletes images. Would be better to either have a local (static) image or find the latest published image for a given OS, instead of hard-coding the image here.

brkt_cli/gce/gce_service.py
```diff
@@ -202,10 +203,47 @@ class GCEService(BaseGCEService):
...
-        self.storage = discovery.build('storage', 'v1',
-                credentials=self.credentials)
+        self.compute = discovery.build(
+            'compute', 'v1', credentials=self.credentials)
+
+        self.storage = discovery.build(
+            'storage', 'v1', credentials=self.credentials)
+
+    # if bucket exists do we have permission?
+    def check_bucket_name(self, bucket):
```
krishnanrs: GCE is very strict about naming conventions. Maybe we can add a validation method to validate the bucket name, file name etc.?

brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+        else:
+            log.info("Can't upload log files")
+            return 1
+
+        log.info(
+            'Logs available at https://storage.cloud.google.com/%s/%s'
+            % (values.bucket, values.file))
+
+    except Exception as e:
+        log.info('share_logs error: %s ' % e)
```
bburtin: Is this an unexpected exception?  If so, we should let the higher-level code catch and handle it.  If we log the error at info level, we'll lose the traceback, making it harder to debug the cause of the error.  If you really want to catch it here, you should probably be using `log.exception()`.

brkt_cli/gce/__init__.py
```diff
@@ -351,23 +469,51 @@ def run(self, values):
         return run_launch(values, self.config)
 
 
+class ShareLogsGCESubcommand(Subcommand):
+
+    def name(self):
+        return 'share-logs-gce'
```
bburtin: Should this be `share-logs`?

brkt_cli/gce/__init__.py
```diff
@@ -351,23 +469,51 @@ def run(self, values):
...
+            'share-logs',
+            formatter_class=brkt_cli.SortingHelpFormatter,
+            description='Share Logs',
+            help='Share logs'
+        )
+        share_logs_gce_args.setup_share_logs_gce_args(share_logs_parser)
+        setup_instance_config_args(share_logs_parser, parsed_config)
+
+    def run(self, values):
+        log.warn(
```
bburtin: I don't think you need this, since it's a new command.  We don't need a top-level command, just `brkt gce share-logs`.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+
+def setup_share_logs_gce_args(parser):
+    parser.add_argument(
+        '--instance',
+        metavar='ID',
```
bburtin: Does GCE have instance IDs?  I thought instances only have names.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+        dest='instance',
+        help='The instance with Bracket system logs to be shared',
+        required=True
+    )
+    parser.add_argument(
+        '--zone',
+        metavar='NAME',
+        help='GCE zone (e.g. us-west-2)',
```
bburtin: Does GCE also have a `us-west-2`?  Or is this copy/pasta from AWS?

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+        dest='zone',
+        required=True
+    )
+    parser.add_argument(
+        '--account',
+        metavar='NAME',
+        help='account to share logs with (e.g. email address)',
```
bburtin: This is vague.  Do we support specifying accounts by something other than email address?  If not, we should just say email address.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_argument(
+        '--bucket',
+        metavar='NAME',
+        help='bucket/file name to store logs',
+        dest='bucket',
+        required=True
+    )
+    parser.add_argument(
+        '--file',
+        metavar='NAME',
```
bburtin: I'm guessing you mean `PATH`, unless it really does have to be a file name without the path.

markvoll: Yes, this is just the name of the file. The path isn't necessary.

bburtin:
One more nit: the descriptions for existing options start with a capital letter.  The new options should follow convention.

brkt_cli/gce/share_logs_gce_args.py
markvoll: Changed to email.
screen shot 2017-01-24 at 4 52 48 pm

bburtin: Would it be more intuitive to use `ADDRESS` instead of `NAME`?

brkt_cli/gce/share_logs_gce_args.py
markvoll: Good catch. Changed to 'NAME'
screen shot 2017-01-24 at 4 57 50 pm

brkt_cli/gce/share_logs_gce_args.py
markvoll: screen shot 2017-01-24 at 5 01 02 pm

brkt_cli/gce/share_logs_gce_args.py
markvoll: done

brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+    disk_name = 'sdb-' + gce_svc.get_session_id()
+    log.info('Sharing logs')
+
+    try:
+        image_project = 'ubuntu-os-cloud'
+        ubuntu_image = 'ubuntu-1404-trusty-v20160222'
+
+        # Check to see if the bucket-name is valid
+        if gce_svc.check_bucket_name(values.bucket):
+            log.info('Bucket-name is not valid')
```
markvoll: Updated...
screen shot 2017-01-24 at 5 52 09 pm

brkt_cli/gce/__init__.py
markvoll: Done

brkt_cli/gce/__init__.py
markvoll: Yes, changed

brkt_cli/gce/__init__.py
markvoll: Removed log.warn statement

brkt_cli/gce/gce_service.py
markvoll: Working on this now

brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+                return 1
+        else:
+            log.info("Can't upload log files")
+            return 1
+
+        log.info(
+            'Logs available at https://storage.cloud.google.com/%s/%s'
+            % (values.bucket, values.file))
+
+    except Exception as e:
```
bburtin: Do we need this `except`?  Typed exceptions should be handled at the callsite.  If an unexpected exception is thrown, we want to see a traceback in the output, so that we have more context.

brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+        log.exception('share_logs error: %s ' % e)
+
+    # Delete the instance, disks, and snapshot
+    finally:
+        if snap:
```
bburtin: If you get rid of the `except` above, you should put the `delete_snapshot()` call in its own `try/except` block.  That way, an exception raised during `delete_snapshot()` won't cause the initial exception to get dropped.  If this doesn't make sense, stop by so we could talk in more detail.

brkt_cli/gce/gce_service.py
```diff
@@ -202,10 +203,72 @@ class GCEService(BaseGCEService):
...
+
+
+    def check_bucket_file(self, bucket, file):
+        try:
+            existing_file = self.storage.objects().get(
+                    bucket=bucket,
+                    object=file).execute()
+            if existing_file:
+                return True
+        except:
```
bburtin: Do we really want to catch `Exception` here?  I'm concerned that an unexpected exception (e.g. `IOException`)  will result in confusing behavior.

brkt_cli/gce/gce_service.py
```diff
@@ -202,10 +203,72 @@ class GCEService(BaseGCEService):
...
+        return False
+
+    def wait_bucket_file(self, bucket, file):
+        for i in range(18):
+            try:
+                self.storage.objects().get(
+                   bucket=bucket,
+                   object=file).execute()
+                return True
+            except:
```
bburtin: Can this exception be typed?  Also, you may want to log the exception at debug level so we can see what's going on when running in verbose mode.

brkt_cli/gce/gce_service.py
```diff
@@ -202,10 +203,72 @@ class GCEService(BaseGCEService):
...
+
+    def get_public_image(self):
+        try:
+            # guest_os is a family
+            self.log.info("Trying to get from family...")
+            request = self.compute.images().getFromFamily(
+                project='ubuntu-os-cloud',
+                family='ubuntu-1404-lts')
+            public_image = request.execute(num_retries=5)
+        except:
```
bburtin: Another untyped exception.

brkt_cli/gce/__init__.py
```diff
@@ -145,6 +146,112 @@ def run_launch(values, config):
...
+    log.info('Sharing logs')
+
+    try:
+        image_project = 'ubuntu-os-cloud'
+
+        # Retrieve latest image from family
+        ubuntu_image = gce_svc.get_public_image()
+        # Check to see if the bucket-name is valid
+        if gce_svc.check_bucket_name(values.bucket):
+        	raise ValidationError('Bucket-name is not valid')
```
markvoll: Updated check_bucket_name() to raise separate error when bucket is not available vs invalid.

brkt_cli/gce/__init__.py
```diff
@@ -351,23 +469,47 @@ def run(self, values):
...
+            LaunchGCEImageSubcommand(),
+            ShareLogsGCESubcommand()]
 
 
 def check_args(values, gce_svc, cli_config):
     if values.encryptor_image:
         if values.bucket != 'prod':
-            raise ValidationError("Please provided either an encryptor image or an image bucket")
+            raise ValidationError(
+                "Please provided either an encryptor image or an image bucket")
```
markvoll: 👍

brkt_cli/gce/__init__.py
```diff
@@ -153,103 +154,104 @@ def run_sharelogs(values, config):
...
+		log.info(
+		    'Logs available at https://storage.cloud.google.com/%s/%s'
+		    % (values.bucket, values.path))
+	finally:
+		try:
+			if snap:
+				gce_svc.delete_snapshot(snapshot_name)
+			gce_svc.cleanup(values.zone, None)
+		except Exception as e:
+			raise util.BracketError("Failed during cleanup: %s", e)
```
bburtin: Would it make sense to log instead of raising an exception?  `BracketError` means that something failed.  What's worse is that if an exception was thrown in the `try` block, it would get dropped on the floor if cleanup fails.

brkt_cli/gce/gce_service.py
```diff
@@ -214,20 +214,54 @@ def check_bucket_name(self, bucket):
...
         except errors.HttpError as e:
             code = json.loads(e.content)['error']['code']
-            if code == 403:
-                return True
+            if code == 404:
+                # bucket doesn't exist. Check syntax
+                return self.validate_bucket_name(bucket)
+            elif code == 403:
+                # bucket does exist. Permisions not valid
+                raise ValidationError("Bucket exists: Permission denied."
```
bburtin: "Bucket exists" is gratuitous if you say that permission is denied.  I think it's sufficient to say "Permission denied for bucket X" or something similar.   The "update bucket permissions" part is a bit presumptive.  It's good to report what failed, but telling the user how to fix the problem may be a slippery slope.

brkt_cli/gce/gce_service.py
```diff
@@ -214,20 +214,54 @@ def check_bucket_name(self, bucket):
...
             else:
-                return False
+                # unexpected Http error
+                raise ValidationError(e)
 
     # Check if the bucket name uses valid syntax
-    #def validate_bucket_name(self,bucket):
+    def validate_bucket_name(self,bucket):
+        for s in bucket:
+            if s.isupper():
```
bburtin: Is this right?  You want to return `True` if the first character is upper case?  Also, please add a unit test for this function.

brkt_cli/gce/gce_service.py
```diff
@@ -214,20 +214,54 @@ def check_bucket_name(self, bucket):
...
+            return True
+        # Bucket name is valid
+        return False
 
 
     # Check if the file name uses valid syntax
-    #def validate_file_name(self,):
+    def validate_file_name(self, file):
+        if "[" in file or "]" in file or "*" in file or "?" in file:
+            raise ValidationError('File name is invalid')
```
bburtin: The message should tell the user what the requirements are for a valid filename.  Otherwise they won't know how to fix the problem.  Also, please add a unit test.

brkt_cli/gce/gce_service.py
```diff
@@ -238,36 +272,33 @@ def check_bucket_file(self, bucket, file):
...
-        except:
-            # guest_os is an image already
-            self.log.info("Trying to get specific image...")
-            request = self.compute.images().get(
-                project='ubuntu-os-cloud',
-                image='ubuntu-1404-lts')
-            public_image = request.execute(num_retries=5)
-            logger.info("Public image: %s", public_image['name'])
+        # guest_os is a family
+        self.log.info("Trying to get from family...")
```
bburtin: Is this message helpful to the end user?  If not, it should either be debug or not there at all.

brkt_cli/gce/gce_service.py
markvoll: updated to "Permission denied for bucket <bucket>"

brkt_cli/gce/__init__.py
markvoll: Yes, makes sense. Changed to log.error

brkt_cli/gce/gce_service.py
markvoll: Changed to catch a typed Exception.

brkt_cli/gce/__init__.py
```diff
@@ -265,7 +265,7 @@ def share_logs(values, gce_svc):
 				gce_svc.delete_snapshot(snapshot_name)
 			gce_svc.cleanup(values.zone, None)
 		except Exception as e:
-			raise util.BracketError("Failed during cleanup: %s", e)
+			log.error("Failed during cleanup: %s", e)
```
bburtin: This should be a warning.  An error is a state from which you can't recover.

brkt_cli/gce/gce_service.py
```diff
@@ -233,8 +233,7 @@ def check_bucket_name(self, bucket):
                 return self.validate_bucket_name(bucket)
             elif code == 403:
                 # bucket does exist. Permisions not valid
-                raise ValidationError("Bucket exists: Permission denied."
-                    " Update bucket permissions or create new bucket")
+                raise ValidationError("Permission denied for bucket %s", (bucket))
```
bburtin: Parens around `bucket` are not necessary, and I believe you want `%`, not `,`.

brkt_cli/gce/gce_service.py
```diff
@@ -281,9 +280,12 @@ def check_bucket_file(self, bucket, file):
                     object=file).execute()
             if existing_file:
                 return True
-        except:
-            pass
-        return self.validate_file_name(file)
+        except errors.HttpError as e:
+            code = json.loads(e.content)['error']['code']
```
bburtin: Should this be `e.resp.status`?  Seems weird to be parsing the HTTP status out of the payload.

brkt_cli/gce/__init__.py
```diff
@@ -225,6 +334,16 @@ def register(self, subparsers, parsed_config):
         setup_instance_config_args(launch_gce_image_parser, parsed_config,
                                    mode=INSTANCE_METAVISOR_MODE)
 
+        share_logs_parser = gce_subparsers.add_parser(
+            'share-logs',
+            description='Share Logs',
+            help='Share logs',
```
bburtin: `help` should describe what the command does in 10 words or less.  If this is not viable, I'd say get rid of it instead of just copying the name of the command.

brkt_cli/gce/__init__.py
```diff
@@ -225,6 +334,16 @@ def register(self, subparsers, parsed_config):
         setup_instance_config_args(launch_gce_image_parser, parsed_config,
                                    mode=INSTANCE_METAVISOR_MODE)
 
+        share_logs_parser = gce_subparsers.add_parser(
+            'share-logs',
+            description='Share Logs',
```
bburtin: `description` should briefly describe the process of what the command does.  This can be longer than `help`.

brkt_cli/gce/__init__.py
```diff
@@ -366,23 +486,46 @@ def run(self, values):
         return run_launch(values, self.config)
 
 
+class ShareLogsGCESubcommand(Subcommand):
```
bburtin: I don't think we want a top-level command.  The new convention is to put GCE commands under `gce`.

brkt_cli/gce/gce_service.py
```diff
@@ -212,10 +213,105 @@ class GCEService(BaseGCEService):
...
+            code = e.resp.status
+            if code == 404:
+                # bucket doesn't exist. Check syntax
+                return self.validate_bucket_name(bucket)
+            elif code == 403:
+                # bucket does exist. Permisions not valid
+                raise ValidationError("Permission denied for bucket %s" % bucket)
+            else:
+                # unexpected Http error
+                raise ValidationError(e)
```
bburtin: If this is an unexpected error, it should be `raise`, and not a `ValidationError`.

brkt_cli/gce/gce_service.py
```diff
@@ -212,10 +213,105 @@ class GCEService(BaseGCEService):
...
+
+    # Check if the bucket name uses valid syntax
+    def validate_bucket_name(self,bucket):
```
bburtin: This should return `True` when the name is valid.  Also, please add a unit test that calls this function directly and tests all the edge cases.

brkt_cli/gce/__init__.py
```diff
@@ -160,6 +162,113 @@ def run_launch(values, config):
...
+	log.info('Sharing logs')
+	snap = None	
+
+	try:
+		image_project = 'ubuntu-os-cloud'
+		# Retrieve latest image from family
+		ubuntu_image = gce_svc.get_public_image()
+		# Check to see if the bucket is available
+		if gce_svc.check_bucket_name(values.bucket):
+			raise ValidationError('Bucket-name is not valid')
```
bburtin: Do you mean "bucket name" or `--bucket-name`?  Also, when the user sees this, he won't know what the validation rules are.  This will create confusion.

brkt_cli/gce/gce_service.py
```diff
@@ -212,10 +213,105 @@ class GCEService(BaseGCEService):
...
+        if bucket.startswith('-') or bucket.endswith('-'):
+            return True
+        # Bucket name is valid
+        return False
+
+
+    # Check if the file name uses valid syntax
+    def validate_file_name(self, file):
+        if "[" in file or "]" in file or "*" in file or "?" in file:
+            raise ValidationError('File name is invalid')
```
bburtin: The user won't know why the file name is invalid.

brkt_cli/gce/gce_service.py
```diff
@@ -212,10 +213,105 @@ class GCEService(BaseGCEService):
...
+                   bucket=bucket,
+                   object=file).execute()
+                return True
+            except errors.HttpError as e:
+                self.log.debug(e)
+                code = e.resp.status
+                if code == 404:
+                    time.sleep(5)
+                else:
+                    self.log.debug("Failed uploading tar file: %s", e)
```
bburtin: Do we want the user to know about this failure?

brkt_cli/gce/gce_service.py
```diff
@@ -212,10 +213,105 @@ class GCEService(BaseGCEService):
...
+                    return False
+        return False
+
+    def get_public_image(self):
+        # guest_os is a family
+        self.log.info("Trying to get from family...")
```
bburtin: Will the user know what "get from family" means?  If not, get rid of it.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    parser.add_argument(
+        '--bucket',
+        metavar='NAME',
+        help='Bucket name to store logs',
+        dest='bucket',
+        required=True
+    )
+    parser.add_argument(
+        '--path',
```
bburtin: From the option name, it's not clear what you're storing in the path.  Is it the logs?  Should this be `--log-path`?  Also, it's not clear whether this is a local path or a bucket path.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -0,0 +1,58 @@
...
+        metavar='NAME',
```
bburtin: A name and a path are different things.

brkt_cli/gce/gce_service.py
markvoll: Yes, this won't be run by a customer

brkt_cli/gce/gce_service.py
```diff
@@ -225,52 +225,46 @@ def check_bucket_name(self, bucket):
...
+            raise ValidationError(
+                "Bucket name must be lower case letters numbers and hyphens")
 
         if (3 > len(bucket) or len(bucket) > 63):
-            return True
+            raise ValidationError('Bucket name must be 3-63 characters')
 
         if 'google' in bucket:
-            return True
+            raise ValidationError('Bucket name cant contain "google" ')
```
bburtin: Typo.  These errors are end-user visible, so we can't be sloppy with spelling.

brkt_cli/gce/gce_service.py
```diff
@@ -225,52 +225,46 @@ def check_bucket_name(self, bucket):
...
 
         if bucket.startswith('-') or bucket.endswith('-'):
-            return True
+            raise ValidationError('Bucket name cant start or end with "-"')
```
bburtin: Typo.

brkt_cli/gce/share_logs_gce_args.py
```diff
@@ -44,8 +44,8 @@ def setup_share_logs_gce_args(parser):
     )
     parser.add_argument(
         '--path',
-        metavar='NAME',
-        help='Path in bucket to store file (e.g. dir/filename)',
+        metavar='Path',
```
bburtin: The convention is `PATH`, not `Path`.

test_gce.py
```diff
@@ -360,3 +395,131 @@ def test_cleanup(self):
...
+    def test_bucket_name_exists(self):
+        gce_svc = GCEService1()
+        with self.assertRaises(ValidationError):
+            share_logs(self.values, gce_svc)
+
+    # uppercase in bucket name
+    def test_bucket_name_invalid1(self):
+        session_id = util.make_nonce()
+        gce_svc = gce_service.GCEService(self.values.project, session_id, log)
+        bucket = 'Test-bucket'
```
bburtin: It would be more concise to just write a single test.  Create a list like `['Test-bucket', 'Test.bucket', 'ab', ...]`.  Then loop through the list and make sure that each validation raises `ValidationError`.  Write less code when you have the option.

PR #540, opened 2017-01-18
Add XTS support for ESX encryption

krishnanrs:
- Add a hidden {--crypto-policy} optional argument to specify the
crypto policy for the guest root volume
 - Default the crypto policy for the guest root volume to XTS to
bring it on par with AWS and GCE
 - Updated tests


apoorva-jindal:
LGTM

PR #543, opened 2017-01-21
Enable SRIOV supprot by default

krishnanrs:
Enable SRIOV support by default
 - Irrespective of whether the guest image supports SRIOV or not
   Bracket supports it and hence should be enabled on all images
   encrypted using the new NetBSD Bracket encryptor

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -950,10 +950,10 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
             aws_svc, guest_instance, image_id
         )
 
+        net_sriov_attr = aws_svc.get_instance_attribute(guest_instance.id,
+                                                        "sriovNetSupport")
         if (guest_image.virtualization_type == 'hvm' and
             'brkt-avatar-freebsd' not in mv_image.name):
-            net_sriov_attr = aws_svc.get_instance_attribute(guest_instance.id,
-                                                            "sriovNetSupport")
             if net_sriov_attr.get("sriovNetSupport") == "simple":
                 log.warn("Guest Operating System license information will not "
                          "be preserved because guest has sriovNetSupport "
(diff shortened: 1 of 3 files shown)


jimvoll:
The sriov changes look good to me

krishnanrs:
I've tested this and put some checks in place to enable SRIOV only if the guest image does not have it enabled already (otherwise it becomes a no-op).

krishnanrs:
Can someone take another look? I tested this out and seems to be working correctly. If there are any nits or errors, let me know.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -988,6 +988,23 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
                 encryptor_instance, mv_image, image_id=image_id,
                 vol_type=vol_type, iops=iops, legacy=legacy,
                 save_encryptor_logs=save_encryptor_logs, status_port=status_port)
+
+        if 'brkt-avatar-freebsd' in mv_image.name and \
```
bburtin: Now that we're doing this check in multiple places, can we break it out into a utility function?  It's still a temporary hack, but at least it would be contained.

brkt_cli/aws/encrypt_ami.py
```diff
@@ -988,6 +988,23 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
...
+            log.info('Enabling sriovNetSupport for guest instance %s',
+                      guest_instance.id)
+            try:
+                ret = aws_svc.modify_instance_attribute(guest_instance.id,
+                                                        "sriovNetSupport",
+                                                        "simple")
+                if ret:
+                    log.info('sriovNetSupport enabled successfully')
+                else:
+                    log.info('Failed to enable sriovNetSupport')
```
bburtin: Should this be a warning or failure?

brkt_cli/aws/encrypt_ami.py
```diff
@@ -988,6 +988,23 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
...
+            except EC2ResponseError as e:
+                log.warn('Unable to enable sriovNetSupport for guest '
```
bburtin: Is it safe to continue here?

brkt_cli/aws/encrypt_ami.py
krishnanrs: Warning would be better.

brkt_cli/aws/encrypt_ami.py
krishnanrs: Yes.

brkt_cli/aws/encrypt_ami.py
jimvoll: warning - just means the instance will fall back to slower xn interface

brkt_cli/aws/encrypt_ami.py
jimvoll: yes it's safe to continue

PR #544, opened 2017-01-21
Fix VMDK encryption failure

krishnanrs:
Assign the crypto policy value to the correct instance_config attribute


apoorva-jindal:
LGTM

PR #550, opened 2017-01-29
Userdata changes for unencrypted guests

bonzotronix:
* These changes enable running MV instances (launched from the usual 'creator' images/AMIs with unencrypted guest roots attached. Such instances must have 'solo_mode: metavisor' in the userdata; they may also need Yeti endpoints and Yeti CA certs included as well.
* Adds (hidden) --unencrypted-guest arg to make-user-data to allow MV instances to be launched with unencrypted guest root disks.
* Allow brkt env and CA certs to be included in userdata for instances in 'metavisor' mode.
* Ensure the solo_mode is always included in userdata (only new for 'metavisor' mode.
* Adapted unit tests to these new behaviors.

Testing:
* Ran updated unit tests.
* Encrypted instance in GCE.
* Generated userdata manually for various CLI inputs.

### DIFF
--- a/brkt_cli/instance_config.py
+++ b/brkt_cli/instance_config.py
@@ -35,9 +35,9 @@ GUEST_FILES_CONTENT_TYPE = 'text/brkt-guest-files'
 
 # Some instance config args are only supported when the Metavisor instance
 # is running in 'creator' mode.
-INSTANCE_METAVISOR_MODE = 1
-INSTANCE_CREATOR_MODE   = 2
-INSTANCE_UPDATER_MODE   = 3
+INSTANCE_METAVISOR_MODE = 'metavisor'
+INSTANCE_CREATOR_MODE   = 'creator'
+INSTANCE_UPDATER_MODE   = 'updater'
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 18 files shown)


brkt_cli/instance_config.py
```diff
@@ -78,24 +83,31 @@ def add_brkt_file(self, dest_filename, file_contents):
     def add_guest_file(self, guest_file):
         self._guest_files.append(guest_file)
 
+    def ensure_solo_mode_in_config(self):
```
bburtin: Can we default `brkt_config['solo_mode']` in `__init__()`?  Seems like that would be cleaner.

brkt_cli/instance_config_args.py
```diff
@@ -68,35 +68,35 @@ def setup_instance_config_args(parser, parsed_config,
         dest='proxy_config_file'
     )
 
-    if mode in (INSTANCE_CREATOR_MODE, INSTANCE_UPDATER_MODE):
```
bburtin: Please make sure that this change doesn't unintentionally revert behavior from commit 00126f289dc0b44108e35026a0912069f153ffac

bonzotronix: Confirmed.  The help for make-user-data looks right after these changes.

brkt_cli/instance_config.py
bonzotronix: We'd still need to check and make sure solo_mode didn't get omitted when the set_brkt_config() method is called.  So I'm not sure how to improve this.

brkt_cli/instance_config.py
```diff
@@ -39,6 +39,11 @@
 INSTANCE_CREATOR_MODE   = 2
 INSTANCE_UPDATER_MODE   = 3
 
+SOLO_MODE_STR = {
```
bonzotronix: Sounds good.  Done.

bonzotronix:
PTAL

brkt_cli/instance_config.py
bburtin: It's Pythonic!  (*groan*)  You're never 100% sure.  I generally lean toward using sensible defaults and writing fewer lines of code.  It's easier than attempting to write bullet-proof code in a language where all bets are off.

PR #553, opened 2017-01-31
Use boolean for allow_unencrypted_guest

bonzotronix:
* Instead of "true", use a boolean (true) for the flag to convey that it's OK to boot from an unencrypted guest.


bonzotronix:
(did a re-push just to kick travis into realizing that the checks had passed)

PR #558, opened 2017-02-03
Provide ability to tag GCE instances

krishnanrs:
GCE provides ability to tag instances and subsequently these tags
can be used in the network/firewall rules. Hence firewall rules
can be specified based on specific instance tags, which can be
assigned at launch time to control the network communication(s)
required by the specific GCE instance.
Online docs hace also been updated to reflect the new arguments/
usage.

### DIFF
--- a/brkt_cli/gce/encrypt_gce_image_args.py
+++ b/brkt_cli/gce/encrypt_gce_image_args.py
@@ -72,6 +72,16 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         default=parsed_config.get_option('gce.subnetwork', None),
         required=False
     )
+    parser.add_argument(
+        '--gce-tag',
+        metavar='VALUE',
+        dest='gce_tags',
+        action='append',
+        help=(
+              'Set a GCE tag on the encryptor instance. May be specified'
+              ' multiple times.'
+        )
+    )
     # Optional Image Name that's used to launch the encryptor instance. This
     # argument is hidden because it's only used for development.
     parser.add_argument(
--- a/aws.md
+++ b/aws.md
@@ -162,8 +162,8 @@ optional arguments:
                         Any port in range 1-65535 can be used except for port
                         81. (default: 80)
   --subnet ID           Launch instances in this subnet (default: None)
-  --tag KEY=VALUE       Custom tag for resources created during encryption.
-                        May be specified multiple times. (default: None)
+  --tag KEY=VALUE       Set an AWS tag on resources created during encryption.
+                        May be specified multiple times.
   --token TOKEN         Token that the encrypted instance will use to
                         authenticate with the Bracket service. Use the make-
                         token subcommand to generate a token. (default: None)
@@ -218,8 +218,8 @@ optional arguments:
                         Any port in range 1-65535 can be used except for port
                         81. (default: 80)
   --subnet ID           Launch instances in this subnet (default: None)
-  --tag KEY=VALUE       Custom tag for resources created during encryption.
-                        May be specified multiple times. (default: None)
+  --tag KEY=VALUE       Set an AWS tag on resources created during update. May
+                        be specified multiple times.
   --token TOKEN         Token that the encrypted instance will use to
                         authenticate with the Bracket service. Use the make-
                         token subcommand to generate a token. (default: None)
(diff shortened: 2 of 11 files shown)


krishnanrs:
Couple of items of note:
1. The `validate_image_name` is now overloaded since the same validation applies for tags as well as instance names
2. No specific tests have been added

If there are any feedback/suggestions on the above items, let me know.

krishnanrs:
Please take a look again.

brkt_cli/aws/diag_args.py
```diff
@@ -67,7 +67,7 @@ def setup_diag_args(parser):
         help='Launch instances in this subnet'
     )
     parser.add_argument(
-        '--tag',
```
bburtin: We should hide the original option and log deprecation warnings when they're used.  Maybe do this in a separate commit?

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -72,6 +72,16 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         default=parsed_config.get_option('gce.subnetwork', None),
         required=False
     )
+    parser.add_argument(
+        '--gce-tag',
+        metavar='VALUE',
+        dest='tags',
```
bburtin: Can we call the arg field `gce_tags`, so that there isn't a discrepancy between the option name and variable name?

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -72,6 +72,16 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
...
+        action='append',
+        help=(
+              'Set a tag on the GCE encryptor instance. May be specified'
```
bburtin: The wording doesn't distinguish Bracket tags vs. CSP tags.  And this PR reignited the terminology debate.  Let's wait until the terminology is finalized before merging.

brkt_cli/gce/encrypt_gce_image_args.py
```diff
@@ -110,6 +120,6 @@ def setup_encrypt_gce_image_args(parser, parsed_config):
         dest='crypto',
         metavar='NAME',
         choices=[CRYPTO_GCM, CRYPTO_XTS],
-        help=argparse.SUPPRESS,
-        default=None
+        help='Specify the name of the generated encrypted image',
```
bburtin: Can we put this into a separate commit?  Seems like it's not related to tags.

krishnanrs: I don't know why/how I got this change in. Will remove this from the commit.

brkt_cli/aws/diag_args.py
krishnanrs: If a user decides to use both the old and new option (just for kicks), should we print the warning and merge the two list of argument values?

krishnanrs:
Please take a look again, pending agreement of the naming of tags vs claims or anything else.

krishnanrs:
@bburtin can you take one final look?

brkt_cli/aws/__init__.py
```diff
@@ -130,7 +130,17 @@ def run_diag(values):
...
     default_tags = {}
-    default_tags.update(brkt_cli.parse_tags(values.tags))
+    aws_tags = None
+    if values.tags:
+        log.warn(
+            'The "--tag" argument has been deprecated. Please use '
+            '"--aws-tag" instead.'
+        )
+        if values.aws_tags:
+            aws_tags = values.tags + values.aws_tags
```
bburtin: Since `--aws-tags` is replacing `--tags`, I think it's a bit magical for us to be appending the two lists.

brkt_cli/aws/__init__.py
```diff
@@ -311,7 +331,17 @@ def run_update(values, config, verbose=False):
     encrypted_image = _validate_ami(aws_svc, values.ami)
     encryptor_ami = values.encryptor_ami or _get_encryptor_ami(values.region)
     default_tags = encrypt_ami.get_default_tags(nonce, encryptor_ami)
-    default_tags.update(brkt_cli.parse_tags(values.tags))
+    aws_tags = None
```
bburtin: Can we turn this into a single `get_aws_tags()` function instead of duplicating the code?

brkt_cli/aws/diag_args.py
bburtin: Sorry, I missed your original question about merging.

krishnanrs:
Removed all AWS changes, except for minor doc updates. Will submit the AWS changes in a separate commit. Please review just for the GCE specific changes.

PR #559, opened 2017-02-06
Normalize image name for S3 downloads

krishnanrs:
The Metavisor images are stored in S3 buckets in the format
{{<image_name>/<image_name>.ovf}}. The {{--metavisor-ovf-image-name}}
argument expects the bucket key, instead of the OVF file name. This
change allows the user to specify either the bucket key or the full
OVF image name and normalizes them accordingly to ensure ease of use.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -957,6 +957,9 @@ def download_ovf_from_s3(bucket_name, image_name=None, proxy=None):
         raise Exception("Invalid bucket-name")
     ovf_name = None
     download_file_list = []
+    # Normalize image_name for S3 downloads
+    if image_name and image_name.endswith('.ovf'):
+        image_name = image_name[:image_name.find('.ovf')]
     try:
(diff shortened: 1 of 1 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -957,6 +957,9 @@ def download_ovf_from_s3(bucket_name, image_name=None, proxy=None):
         raise Exception("Invalid bucket-name")
     ovf_name = None
     download_file_list = []
+    # Normalize image_name for S3 downloads
+    if image_name and image_name.endswith('.ovf'):
```
bburtin: For what it's worth, this magic will only work if `.ovf` is in lower case.

krishnanrs: Agreed, but I don't think that is much of a problem because:
1. The filename in the S3 bucket ends with ".ovf" (lowercase)
2. This is an option primarily used by dev/test engineers
3. There are other places within the esx service where we expect the file extension to be lowercase
However I will wait for @apoorva-jindal to confirm.

apoorva-jindal:
LGTM. 
I agree with Boris, but since the file dumped by vcenter has ".ovf" in lowercase, we can assume it to be lowercase for now.

PR #561, opened 2017-02-08
Deprecate -v at the subcommand level

bburtin:
Suppress the -v option from subcommand usage output and log a warning
when the option is used.  Also update sample usage in AWS docs.

Back story: some of our subcommands support the -v (--verbose) option
and others don't.  It's more consistent to just have it at the brkt
command level than to make sure that every subcommand implements it.

### DIFF
--- a/brkt_cli/get_public_key/__init__.py
+++ b/brkt_cli/get_public_key/__init__.py
@@ -11,6 +11,7 @@
 # CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and
 # limitations under the License.
+import argparse
 import logging
 
 import brkt_cli.crypto
@@ -54,7 +55,7 @@ class GetPublicKeySubcommand(Subcommand):
             '--verbose',
             dest='make_private_key_verbose',
             action='store_true',
-            help='Print status information to the console'
+            help=argparse.SUPPRESS
         )
 
     def verbose(self, values):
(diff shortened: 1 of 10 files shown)


brkt_cli/get_public_key/__init__.py
```diff
@@ -54,7 +55,7 @@ def register(self, subparsers, parsed_config):
             '--verbose',
             dest='make_private_key_verbose',
             action='store_true',
-            help='Print status information to the console'
+            help=argparse.SUPPRESS
```
krishnanrs: No deprecation warnings here?

brkt_cli/make_key/__init__.py
```diff
@@ -71,7 +72,7 @@ def register(self, subparsers, parsed_config):
             '--verbose',
             dest='make_private_key_verbose',
             action='store_true',
-            help='Print status information to the console'
+            help=argparse.SUPPRESS
```
krishnanrs: Same as above: no deprecation warning?

brkt_cli/get_public_key/__init__.py
bburtin: I implemented the deprecation warning so that it's printed at runtime.  Not sure what's better: suppress `-v` from the usage output, or keep it and have the help say that it's deprecated?

krishnanrs: Ignore this. I didn't realize that you already added it in brkt_cli/__init__.py.

PR #562, opened 2017-02-09
Read config variables for GCE Launch command

krishnanrs:
The `brkt gce launch` command was not reading the brkt config
settings for some of the attributes. This update ensures that
the launch command is on par with the encrypt and update GCE
commands for reading the config settings.


xcllnt:
Yay!

PR #564, opened 2017-02-09
Remove unused ca_cert.pem

bburtin:
We no longer need a certificate to make API calls to the Bracket
service.  Remove the file from the repo and put the content into the
instance config unit test.


brkt_cli/test_instance_config.py
```diff
@@ -270,19 +282,23 @@ def test_ca_cert(self):
                 ic = instance_config_from_values(values)
 
         # Now use endpoint args and a valid cert, with all three modes
-        cli_args = endpoint_args + ' --ca-cert %s' % _get_ca_cert_filename()
```
bonzotronix: We can also get rid of the _get_ca_cert_filename func.

bburtin: Done.  Good catch!

bonzotronix:
LGTM

PR #566, opened 2017-02-10
Provide ability to add ssh public key to instance user data

krishnanrs:
When launching unencrypted guests on ESX, currently there are no
means to pass the SSH public key to the instance. As a result, if
a (dev) user tries to SSH to the metavisor, it would fail.

This change allows creating a user-data blob with the ssh public
key to be used by the metavisor instance. The user-data is then
passed to the instance at launch time which allows it to download
the public key and allow anyone having the associated private key
to be able to SSH in to the metavisor.

The option is kept hidden since the primary audience for the use is
Bracket (internal) engineering who would want access to the Metavisor.
A debug Metavisor version would still be required for the SSH to work.


apoorva-jindal:
LGTM

PR #567, opened 2017-02-11
Add a "/" to the metavisor directory in the ovf bucket

atonderski:
Nondebug freebsd esx encryption was completely broken.

Since we're naming our artifacts "version" and "version-debug" we
were getting a conflict when just calling bucket.list("name").
In the case of debug it would work fine, but for release both debug
and nondebug artifacts were downloaded.

This would have worked fine if it wasn't for the fact that one of the
files in both buckets is called disk-0.vmdk.


atonderski:
Agreed, let's not rush it. 
However I do think we should put in this change even if we change the naming as you suggested. It doesn't hurt anything and this way we are safe from similar collisions happening in the future.

PR #569, opened 2017-02-13
Add auth command

bburtin:
Add a top-level auth command that authenticates with the Bracket service
and prints an API auth token.  This command can be used by developers
and scripts that need to make calls to the Bracket REST API.  Update the
existing token validation code to use the same yeti module.

The next step will be to add login/logout commands (probably under
config), which use the yeti module to authenticate and stores the API
token in the user's config file.

---------------
```
$ ./brkt auth -h
usage: brkt auth [-h] [--email ADDRESS] [--password PASSWORD] [--root-url URL]
                 [--out PATH]

Authenticate with the Bracket service. On success, print the JSON Web Token
that can be used to make calls to Bracket REST API endpoints.

optional arguments:
  --email ADDRESS      If not specified, show a prompt.
  --out PATH           Write to a file instead of stdout. This can be used to
                       avoid character encoding issues when redirecting output
                       on Windows.
  --password PASSWORD  If not specified, show a prompt.
  --root-url URL       Bracket service root URL (default:
                       https://api.mgmt.brkt.com)
  -h, --help           show this help message and exit
```

### DIFF
--- a/brkt_cli/auth.py
+++ b/brkt_cli/auth.py
@@ -1,0 +1,77 @@
+# Copyright 2017 Bracket Computing, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# https://github.com/brkt/brkt-cli/blob/master/LICENSE
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+import getpass
+import logging
+
+import brkt_cli
+from brkt_cli import ValidationError
+from brkt_cli import argutil
+from brkt_cli import util
+from brkt_cli.subcommand import Subcommand
+from brkt_cli.yeti import YetiService, YetiError
+
+log = logging.getLogger(__name__)
+
+
+SUBCOMMAND_NAME = 'auth'
+
+
+class AuthSubcommand(Subcommand):
+
+    def name(self):
+        return SUBCOMMAND_NAME
+
+    def register(self, subparsers, parsed_config):
+        parser = subparsers.add_parser(
+            self.name(),
+            description=(
+                'Authenticate with the Bracket service. On success, print '
+                'the JSON Web Token that can be used to make calls '
+                'to Bracket REST API endpoints.'
+            ),
+            help='Authenticate with the Bracket service',
+            formatter_class=brkt_cli.SortingHelpFormatter
+        )
+        parser.add_argument(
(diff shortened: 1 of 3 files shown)


brkt_cli/auth.py
```diff
@@ -0,0 +1,77 @@
...
+            '--password',
+            help='If not specified, show a prompt.'
+        )
+        parser.add_argument(
+            '--root-url',
+            metavar='URL',
+            default='https://api.mgmt.brkt.com',
+            help='Bracket service root URL'
+        )
+        argutil.add_out(parser)
```
bonzotronix: The helptext added by add_out() isn't quite right, as we're dealing with token data rather than private key data.  Suggest either calling add_argument() here or passing add_out some obj_name='token data' arg to override the 'private key' text.

brkt_cli/yeti.py
```diff
@@ -0,0 +1,138 @@
...
+    if 'error_description' in d:
+        reason = d['error_description']
+    return reason
+
+
+def get_json(url, token=None, timeout=10.0):
+    """ Send an HTTP GET to the given endpoint and return the response
+    payload as a dictionary.
+
+    :param token the auth token (JWT)
```
bonzotronix: Nit: "the API token"

brkt_cli/yeti.py
```diff
@@ -0,0 +1,138 @@
...
+        raise YetiError(http_status=r.status_code, message=msg)
+
+    return r.json()
+
+
+def post_json(url, token=None, json=None, timeout=10.0):
+    """ Send an HTTP POST to the given endpoint and return the response
+    payload as a dictionary.
+
+    :param token the auth token (JWT)
```
bonzotronix: "the API token"

brkt_cli/yeti.py
```diff
@@ -0,0 +1,138 @@
...
+class YetiService(object):
+
+    def __init__(self, root_url, token=None):
+        self.root_url = root_url
+        self.token = token
+
+    def auth(self, email, password):
+        """ Authenticate with the Yeti service.
+
+        :return the auth token (JWT)
```
bonzotronix: "the API token"

bonzotronix:
Looks nice and clean; just some wording nits.
I was hoping the 'brkt auth' command would also be the one that would add the token to the config, but we can discuss that separately.

brkt_cli/auth.py
bburtin: Good catch.  I filed a separate PR to clean this up: https://github.com/brkt/brkt-cli/pull/570

bburtin:
Thanks Adam.  I was originally going to bundle auth and login, but couldn't think of a clean way to do it.  You'd end up with a single command that has options like `--update-config` or `--no-update-config`.  I'm currently leaning toward 3 simpler commands: `brkt auth`, `brkt config login`, and `brkt config logout`.

PR #572, opened 2017-02-15
Add login, logout, whoami commands

bburtin:
Add new commmands for logging in and out, and showing the current user.
The login command stores the user's API token in the config file and
logout unsets it.  logout is effectively an alias for "brkt config unset
api-token".

-------------

```
$ ./brkt config login --email me@example.com --password xxxxx
Logged into stage as me@example.com
$ ./brkt config list
api-token=eyJ0eXAi...
$ ./brkt config whoami
me@example.com
$ ./brkt config whoami --json
{
    "email": "me@example.com",
    "eula_accepted": true,
    "eula_accepted_ts": 18446744011573954816,
    "name": "Me Too",
    "uuid": "ac5be98d-021a-4c31-8aee-0d90825102aa"
}
$ ./brkt config logout
$ ./brkt config whoami
Not logged in.  Please run brkt config login.
```
----------------
```
$ ./brkt config login -h
usage: brkt config login [-h] [--email ADDRESS] [--password PASSWORD]

Authenticate with the Bracket service and store the API token in config.

optional arguments:
-h, --help           show this help message and exit
--email ADDRESS      If not specified, show a prompt
--password PASSWORD  If not specified, show a prompt
```
```
$ ./brkt config logout -h
usage: brkt config logout [-h]

Delete the API token stored in config.

optional arguments:
-h, --help  show this help message and exit
```
```
$ ./brkt config whoami -h
usage: brkt config whoami [-h] [--json]

Print the email address of the user who was logged in with the brkt
config
login command.

optional arguments:
-h, --help  show this help message and exit
--json      Print all user properties in JSON format
```

### DIFF
--- a/brkt_cli/config/__init__.py
+++ b/brkt_cli/config/__init__.py
@@ -14,19 +14,22 @@
 import argparse
 import collections
 import errno
+import getpass
 import logging
 import os
 import os.path
 import shutil
 import sys
 import tempfile
+
 import yaml
 
 import brkt_cli
-
+from brkt_cli import util
 from brkt_cli.subcommand import Subcommand
 from brkt_cli.util import parse_endpoint, render_table_rows
 from brkt_cli.validation import ValidationError
+from brkt_cli.yeti import YetiError, YetiService
 
 log = logging.getLogger(__name__)
 
@@ -124,7 +127,7 @@ def _unlink_noraise(path):
 
 
 class CLIConfig(object):
-    """CLIConfig exposes an interface that subcommands can use to retrive
+    """CLIConfig exposes an interface that subcommands can use to retrieve
     persistent configuration options.
     """
 
@@ -489,6 +492,47 @@ The leading `*' indicates that the `stage' environment is currently active.
             'env_name',
             help='The environment name')
 
+        # Log in and out.
+        login_parser = config_subparsers.add_parser(
+            'login',
+            help='Log into the Bracket service',
+            description=(
+                'Authenticate with the Bracket service and store the API '
+                'token in config.'
+            )
+        )
+        login_parser.add_argument(
+            '--email',
+            metavar='ADDRESS',
+            help='If not specified, show a prompt'
+        )
+        login_parser.add_argument(
+            '--password',
+            help='If not specified, show a prompt'
+        )
+
+        config_subparsers.add_parser(
+            'logout',
+            help='Log out of the Bracket service',
+            description=(
+                'Delete the API token stored in config.'
+            )
+        )
+
+        whoami_parser = config_subparsers.add_parser(
+            'whoami',
+            help='Show the current user',
+            description=(
+                'Print the email address of the user who was logged in '
+                'with the brkt config login command.'
+            )
+        )
+        whoami_parser.add_argument(
+            '--json',
+            action='store_true',
+            help='Print all user properties in JSON format'
+        )
+
     def _write_config(self):
         """Create ~/.brkt if it doesn't exist and safely write out a
         new config.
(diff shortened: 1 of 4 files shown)


brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
...
+        print token
+
+    def _logout(self):
+        self.parsed_config.unset_option('token')
+        self._write_config()
+
+    def _whoami(self):
+        y = self._get_yeti_service()
+        if not y.token:
+            raise ValidationError('API token is not set in config')
```
bonzotronix: Maybe add some text like "use the 'login' command to save an API token to the config'

brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
             raise ValidationError('Error: unknown environment ' + values.env_name)
         self._write_config()
 
+    def _get_yeti_service(self):
+        """Return the YetiService object for the current environment."""
+        _, env = self.parsed_config.get_current_env()
+        root_url = 'https://%s:%d' % (
+            env.public_api_host, env.public_api_port)
+        token = self.parsed_config.get_option('token')
```
bonzotronix: Maybe "api-token" for the field name?

bonzotronix:
Basically LGTM.  I'd like to do a last round of getting consensus on the CLI syntax, as we haven't had a review of the latest proposal (and it has deviated a bit from the proposal discussed on Feb 3).  I've asked for feedback on #ux-improvements.

brkt_cli/config/__init__.py
bburtin: There's barely any logic here, so I wasn't sure what to unit test.  Let me know if you have suggestions.

brkt_cli/config/__init__.py
bburtin: Yeah, that would be better than reusing the existing `token` key.  The way it is now, if you log in, you won't be able to encrypt because the API token doesn't have a `kid` claim.

brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
...
+        password = values.password or getpass.getpass('Password: ')
+        y = self._get_yeti_service()
+
+        try:
+            token = y.auth(email, password)
+        except YetiError as e:
+            raise ValidationError(e.message)
+
+        self.parsed_config.set_option('token', token)
+        self._write_config()
```
bburtin: I verified that the config file permissions are 0600.

brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
...
+        print token
```
bonzotronix: I'd suggest that we not print the token for 'brkt config login'.  Maybe a success message instead ("Successfully logged into env %ENV. API token written to config.")

brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
...
+        email = values.email or raw_input('Email: ')
+        password = values.password or getpass.getpass('Password: ')
+        y = self._get_yeti_service()
+
+        try:
+            token = y.auth(email, password)
+        except YetiError as e:
+            raise ValidationError(e.message)
+
+        self.parsed_config.set_option('token', token)
```
bonzotronix: As discussed in #ux-improvements, I think we should make these API tokens be saved on a per-env basis.  Otherwise, it'll be too easy to use an API token obtained by logging in with env A against a Yeti API endpoint for env B.

brkt_cli/config/__init__.py
```diff
@@ -641,6 +679,47 @@ def _unset_env(self, values):
...
+            raise ValidationError('API token is not set in config')
+        try:
+            cust = y.get_customer()
+        except YetiError as e:
+            if e.http_status == 401:
+                raise ValidationError(
+                    'Config API token is not authorized to access %s' %
+                    y.root_url)
+            raise ValidationError(e.message)
+        print '%s %s' % (cust.email, cust.name)
```
krishnanrs: In addition to the email and name, can we print the URL/host? This will especially become important in on-prem world and we are exploring possibilities of having multiple on-prem Yeti's.

brkt_cli/config/__init__.py
bburtin: I was printing the token, so that `login` behaves the same way as `auth`, but I'm fine either way.  Why do you think that printing the token is a bad idea?

bonzotronix: I think it's OK to have `login` output be different from that of `auth`.  Just thinking as a user, I'd rather just see a success (or error) message with `login` and not see the JWT ugliness.  If I'm using `auth` then I care about the JWT data, so it's right to output the token with that command.

bburtin: Ok, makes sense.

krishnanrs:
Are we going to update the README?

bburtin:
Yeah, I'll do that once this stuff is ready for prime time.

PR #574, opened 2017-02-15
Validate vCenter/ESX credentials before proceeding

krishnanrs:
This change attempts to connect and validate the vCenter/ESX credentials
before attempting to download the Metavisor image. This ensures that
in the event of invalid credentials, the error is reported back sooner
rather than after some time (which is takes to download the Metavisor
image from S3).


brkt_cli/esx/__init__.py
```diff
@@ -144,6 +130,22 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
         )
     except Exception as e:
         raise ValidationError("Failed to connect to vCenter: ", e)
+
+    # Download images from S3
```
apoorva-jindal: The issue with doing a download after connecting to vcenter is if it takes a long time to download the MV (which it can depending on the network), the vc_swc connection becomes stale, and any subsequent requests fail on it with BadStatus exception. There are two ways around it. (i) Connect, then disconnect, then download and finally connect again. However, doing too many connects too fast results in increased vcenter timeouts increasing the encryption time. (ii) Download, ask for password and then immediately connect. The downside here is that the password is asked and checked for after download.

krishnanrs: The entire issue here was that if wrong credentials were entered, then we download the MV OVF, which can take a minute or two before telling the user that auth with vCenter/ESX host failed. We were already doing (ii) today (but not disconnecting). Hence either we go the path of (i) or abandon this change.

PR #575, opened 2017-02-15
Use the template-vm-name argument when launch VM instance

krishnanrs:
When encrypting an image based off a local Metavisor/encryptor image,
the template-vm-name argument was being ignored and hence the VM
instance was always getting launched with the name as
'Encryptor-VM-<Date>'. This specifically causes problems when creating
encrypted ESX instances as it would be harder to identify the encrypted
instance when it does not use the user specified name.


apoorva-jindal:
LGTM

PR #577, opened 2017-02-16
Fix error recovery during encryption on ESX.

apoorva-jindal:
1. Teardown the instance (unless --no-teardown is used) on encryption failure.
2. Improve help messages.

### DIFF
--- a/brkt_cli/esx/encrypt_vmdk.py
+++ b/brkt_cli/esx/encrypt_vmdk.py
@@ -30,7 +30,6 @@ import logging
 from brkt_cli.encryptor_service import (
     wait_for_encryptor_up,
     wait_for_encryption,
-    EncryptionError,
     ENCRYPTOR_STATUS_PORT
 )
 from brkt_cli.util import Deadline
@@ -55,7 +54,7 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
         # Reconfigure VM with more CPUs and memory
         vc_swc.reconfigure_vm_cpu_ram(vm)
         # Reconfigure VM with serial port
-        if serial_port_file_name is not None:
+        if serial_port_file_name:
             vc_swc.add_serial_port_to_file(vm, serial_port_file_name)
         # Add datastore path to the guest vmdk
         guest_vmdk_path = vc_swc.get_datastore_path(guest_vmdk)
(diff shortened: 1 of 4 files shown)


brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -115,21 +115,20 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
...
                 print(vc_swc.get_vm_name(template_vm))
-    except EncryptionError as e:
+    except Exception as e:
         log.exception("Failed to encrypt the image with error %s", e)
         try:
             vc_swc.connect()
             vm = vc_swc.find_vm(mv_vm_name)
             vc_swc.power_off(vm)
             vc_swc.detach_disk(vm, unit_number=2)
+            if serial_port_file_name is not None:
```
krishnanrs: Won't this defeat the purpose of saving the logs when we especially want it when encryption fails?

apoorva-jindal: This will detach the serial port from the VM, not delete the console file.

brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -115,21 +114,20 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
                 log.info("Creating the template VM")
                 template_vm = vc_swc.clone_vm(vm, vm_name=vm_name, template=True)
                 print(vc_swc.get_vm_name(template_vm))
-    except EncryptionError as e:
+    except Exception as e:
         log.exception("Failed to encrypt the image with error %s", e)
```
bburtin: `log.exception()` automatically (magically) logs the exception that was thrown.  You don't need to log it.  Should be safe to get rid of `e`.

brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -115,21 +114,20 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
...
         try:
             vc_swc.connect()
             vm = vc_swc.find_vm(mv_vm_name)
             vc_swc.power_off(vm)
             vc_swc.detach_disk(vm, unit_number=2)
+            if serial_port_file_name is not None:
```
bburtin: `if serial_port_file_name:`

PR #578, opened 2017-02-17
Fix the OVF image downloads

krishnanrs:
Do not add the "/" suffix to the bucket/image name unnecessarily.

If a metavisor image name is specified, then it is OK to add a "/"
suffix to it to ensure that we are not overlapping with another
bucket with a similar prefix. However if the metavisor image name
is not specified and the CLI tries to sort thru' the buckets to
identify the latest one (used for Prod and Stage published images)
then the bucket/image name already has a "/" suffixed to it and that
causes the image download to fail as we are adding an additional "/"
to it.


apoorva-jindal:
LGTM

PR #579, opened 2017-02-17
Fix connection issues in Windows.

apoorva-jindal:
1. Report logging for sslerror so that the cli does not seem to be stuck.
2. Retry on ConnectionTimeouts in Windows.


krishnanrs:
Changes LGTM. I will let @bburtin confirm if he has concerns with the aesthetics.

bburtin:
Icky but fine.  :-)

PR #581, opened 2017-02-17
Download artifacts from S3 only if required

krishnanrs:
Previously the brkt-cli would always try to download the Metavisor
artifacts from S3 everytime an encrypt/update command is run, even
if the command fails subsequently do to issues with the vCenter/ESX
server and/or the specified parameters. With this change, we:
 - Check if artifacts were already downloaded
 - Check if the VMDK hash matches the manifest file
Only when one or more of the files are not found or the manifest hash
does not match the VMDK, do we try to download the artifacts from S3.

Additionally a hidden --no-cleanup option has been provided which can
be used in case the user does not want the downloaded artifacts to be
deleted after the encryption/update completes successfully. This is
especially useful in the case where the same artifact is required to
encrypt/update multiple images and eliminates the need to download
the artifacts from S3 for each of the encrypt/update operations.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -225,6 +225,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
                 user_data_str=user_data_str,
                 serial_port_file_name=values.serial_port_file_name,
                 status_port=values.status_port,
+                cleanup=values.cleanup,
             )
         return 0
     except Exception as e:
@@ -371,6 +372,7 @@ def run_update(values, parsed_config, log, use_esx=False):
                 download_file_list=download_file_list,
                 user_data_str=user_data_str,
                 status_port=values.status_port,
+                cleanup=values.cleanup,
             )
         return 0
     except:
--- a/brkt_cli/esx/encrypt_vmdk_args.py
+++ b/brkt_cli/esx/encrypt_vmdk_args.py
@@ -238,3 +238,13 @@ def setup_encrypt_vmdk_args(parser):
         help=argparse.SUPPRESS,
         default=None
     )
+    # Optional argument to keep the downloaded artifacts. Can we used in
+    # cases where the same (downloaded) OVF is used for multiple
+    # encryption/update jobs
+    parser.add_argument(
+        '--no-cleanup',
+        dest='cleanup',
+        default=True,
+        action='store_false',
+        help=argparse.SUPPRESS
+    )
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -951,6 +951,37 @@ def initialize_vcenter(host, user, password, port,
     return vc_swc
 
 
+def need_to_download_from_s3(file_list):
+    """ Checks if the necessary files are already downloaded
+       
+        Returns: True or False to indicate whether the download
+        is required or not and an OVF image name when download
+        is not required
+    """
+    ovf_image = None
+    for item in file_list:
+        file_name = item[item.rfind('/')+1:]
+        if not os.path.exists(file_name):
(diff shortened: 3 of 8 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -225,6 +225,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
                 user_data_str=user_data_str,
                 serial_port_file_name=values.serial_port_file_name,
                 status_port=values.status_port,
+                no_cleanup=values.no_cleanup,
```
bburtin: Can we change this to `cleanup` instead of `no_cleanup`?  Negative variable names can affect code readability (e.g. `if not no_cleanup`).

brkt_cli/esx/encrypt_vmdk_args.py
```diff
@@ -238,3 +238,13 @@ def setup_encrypt_vmdk_args(parser):
         help=argparse.SUPPRESS,
         default=None
     )
+    # Optional argument to keep the downloaded artifacts. Can we used in
+    # cases where the same (downloaded) OVF is used for multiple
+    # encryption/update jobs
+    parser.add_argument(
+        '--no-cleanup',
+        dest='no_cleanup',
```
bburtin: You can keep the option as `--no-cleanup`, but do `dest=cleanup, default=True, action=store_false`

brkt_cli/esx/esx_service.py
```diff
@@ -951,6 +951,34 @@ def initialize_vcenter(host, user, password, port,
...
+def need_to_download_from_s3(file_list):
+    """ Checks if the necessary files are already downloaded
+       
+        Returns: True or False to indicate whether the download
+        is required or not and an OVF image name when download
+        is not required
+    """
+    ovf_image = None
+    for file in file_list:
+        file_name = file[file.rfind('/')+1:]
```
bburtin: I believe that the use of `file` as a variable name is discouraged, because it's a builtin: https://docs.python.org/2/library/functions.html#file `file[file.rfind()]` looks kind of crazy.  Maybe just rename these to f1 and f2?

krishnanrs:
Please take  a look again

brkt_cli/esx/esx_service.py
```diff
@@ -990,15 +1018,19 @@ def download_ovf_from_s3(bucket_name, image_name=None, proxy=None):
...
-            certificate = Key(bucket)
-            certificate.key = file
-            certificate.get_contents_to_filename(target_file)
-            if (".ovf" in file_name):
-                ovf_name = target_file
-            download_file_list.append(target_file)
+        ovf_name, download_required = need_to_download_from_s3(file_list)
+        if download_required:
+            for file in file_list:
+                file_name = file[file.rfind('/')+1:]
```
bburtin: One last nit about `file[file.rfind()]`.

krishnanrs: That wasn't a line I changed, but since I am in the act of cleaning up...

brkt_cli/esx/esx_service.py
```diff
@@ -951,6 +951,34 @@ def initialize_vcenter(host, user, password, port,
...
+    ovf_image = None
+    for item in file_list:
+        file_name = item[item.rfind('/')+1:]
+        if not os.path.exists(file_name):
+            return ovf_image, True
+        if '.mf' in file_name:
+            with open(file_name) as manifest_data:
+                manifest = json.load(manifest_data)
+                manifest_vmdk_sha = manifest['disk-0.vmdk']
+        if '.vmdk' in file_name:
```
apoorva-jindal: Can you also validate the ovf file sha as well?

apoorva-jindal:
LGTM

PR #582, opened 2017-02-21
Add ability to provide host instead of cluster.

apoorva-jindal:
1. If there are no clusters in a datacenter, provide the ability to let
customers provide a host name to --vcenter-cluster.
2. Change the encryptor VM name to template name only if
the encryption is happening on esx host.
3. Fix error in detach disk.


brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -199,9 +199,12 @@ def encrypt_from_local_ovf(vc_swc, enc_svc_cls, guest_vmdk, crypto_policy,
         log.info("Launching VM from local OVF")
         ovf_image_name = ovf_image_name + ".ovf"
         validate_local_mv_ovf(source_image_path, ovf_image_name)
+        mv_vm_name = None
+        if vc_swc.is_esx_host() is True:
```
krishnanrs: Nit. Can simply use if vc_swc.is_ext_host()

PR #584, opened 2017-02-22
MV OVF manifest file name change

apoorva-jindal:


apoorva-jindal:
Fixed so that change is backward compatible.

PR #585, opened 2017-02-22
Pass an empty vm_name for VMDK update operation

krishnanrs:
The launch_from_s3 method is now expecting a vm_name to be always
provided. For VMDK updates, this can be set to None.


apoorva-jindal:
LGTM

PR #586, opened 2017-02-24
Print an update every minute when we are waiting for vcenter task

atonderski:
We are very rarely seeing the encrypt-with-vcenter command freeze indefinitely. This is an attempt to narrow down the cause

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -23,6 +23,7 @@ import signal
 import hashlib
 import requests
 import boto.s3
+import itertools
 
 from functools import wraps
 from threading import Thread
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -373,12 +374,15 @@ def __get_obj(self, content, vimtype, name):
...
-        while True:
+        for i in itertools.count():
             if task.info.state == 'success':
                 return task.info.result
             if task.info.state == 'error':
                 raise Exception('Task failed to finish with error %s' %
                                 task.info.error)
+            time.sleep(1)
+            if i % 60 == 0:
+                log.info("Still waiting for task %s" % task)
```
apoorva-jindal: Instead of printing task, can you print task.info.state, task.info.result, task.info.progress, task.info.name, task.info.descriptionId. Also increase the time from 60 to 180.

bburtin:
Closing old pull request.

PR #587, opened 2017-02-27
Migrate to boto3 for retrieving S3 objects for esx_service

dasein:
There is inconsistent behavior with boto when downloading objects
from S3 where the bucket name contains dots:

https://github.com/boto/boto/issues/2836

To address this issue, VMware operations for fetching images from S3
are now using boto3. In addition to the new dependency on boto3, the
following issues have been addressed.

- The hidden --bucket-name argument now supports s3 buckets with dots
  in the name

- The --metavisor-ovf-image argument can return the latest
  metavisor image with exact or partial matches (e.g. 0-0-964 or
  metavisor-0-0-964-g0cf0e0e62.ovf)

- If the --metavisor-ovf-image argument is _not_ provided, we
  default to retrieving the latest modified image with a *latest* or
  *release-* prefix

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -22,13 +22,16 @@ import os
 import signal
 import hashlib
 import requests
-import boto.s3
-
 from functools import wraps
+from operator import attrgetter
 from threading import Thread
+
+import boto3
+from botocore.handlers import disable_signing
 from pyVim import connect
 from pyVmomi import vmodl
 from pyVmomi import vim
+
 from brkt_cli.util import (
     retry,
     RetryExceptionChecker
@@ -36,10 +39,12 @@ from brkt_cli.util import (
 from brkt_cli import validate_ssh_pub_key
 from brkt_cli.instance_config import INSTANCE_UPDATER_MODE
 from brkt_cli.validation import ValidationError
-from boto.s3.key import Key
 
 
 log = logging.getLogger(__name__)
+logging.getLogger('boto3').setLevel(logging.FATAL)
+logging.getLogger('botocore').setLevel(logging.FATAL)
+logging.getLogger('s3transfer').setLevel(logging.FATAL)
 
 
 class TimeoutError(Exception):
@@ -963,16 +968,15 @@ def initialize_vcenter(host, user, password, port,
 
 def need_to_download_from_s3(file_list):
     """ Checks if the necessary files are already downloaded
-       
+
         Returns: True or False to indicate whether the download
-        is required or not and an OVF image name when download
-        is not required
+        is required or not
     """
-    ovf_image = None
-    for item in file_list:
-        file_name = item[item.rfind('/')+1:]
+    fetch_s3_objects = True
+    vmdk_sha = ovf_sha = manifest = False
+    for file_name in file_list:
         if not os.path.exists(file_name):
-            return ovf_image, True
+            break
         if '.mf' in file_name:
             with open(file_name) as manifest_data:
                 manifest = json.load(manifest_data)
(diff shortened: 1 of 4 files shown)


dasein:
I tempered my eagerness to migrate _everything_ to boto3. According to AWS docs we boto and boto3 can lively safely side by side. Hence, immediate fixes with boto3 first...

brkt_cli/esx/esx_service.py
```diff
@@ -982,73 +986,80 @@ def need_to_download_from_s3(file_list):
...
+
     # Normalize image_name for S3 downloads
     if image_name and image_name.endswith('.ovf'):
         image_name = image_name[:image_name.find('.ovf')]
+
     try:
-        anon = not (set(['AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY']) <= set(os.environ))
+        _environ = dict(os.environ)
+
         if proxy:
```
krishnanrs: It would be better that we check for both the OS env as well as the proxy argument. The OS env, if specified, should override the proxy argument (if that is also specified).

brkt_cli/esx/esx_service.py
```diff
@@ -982,73 +986,80 @@ def need_to_download_from_s3(file_list):
...
+            #TODO(workaround) https://github.com/boto/boto3/issues/338
+            os.environ["HTTP_PROXY"] = "http://%s:%d" % (proxy.host, proxy.port)
+            os.environ["HTTPS_PROXY"] = "https://%s:%d" % (proxy.host, proxy.port)
+
+        s3 = boto3.resource('s3')
+
+        if not (set(['AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY']) <= set(os.environ)):
+            s3.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
+
+        bucket = s3.Bucket(bucket_name)
```
krishnanrs: How are we handling backwards compatibility with the NetBSD images which would be in a different bucket?

brkt_cli/esx/esx_service.py
```diff
@@ -982,73 +986,80 @@ def need_to_download_from_s3(file_list):
...
+        blist = list(bucket.objects.filter(Prefix="metavisor/"))
+        if not blist:
+            blist = list(bucket.objects.all())
+
+        # Get latest, exact, or partial match of ovfs
+        # No debug for partial match
+        if image_name is None:
+            ovfs = [ o for o in blist if o.key.endswith('.ovf')
```
krishnanrs: This check fails for existing FreeBSD and NetBSD images in the old bucket(s)

brkt_cli/esx/esx_service.py
dasein: The behavior mimicks the current behavior with specifying a specific HTTP_PROXY for VMware metavisor downloads (vs proxy settings for yeti API calls). I do not want to change that behavior in this PR.

brkt_cli/esx/esx_service.py
dasein: It does not matter what bucket they are in. We get whatever "latest" (or release) image that was promoted to that bucket. We use the job promotions to determine what is a latest release. If someone wants a specific image, they will provide the --metavisor-ovf-image-name argument.

brkt_cli/esx/esx_service.py
dasein: Not sure what you mean here. Can you specify the failure. The changes made should be compatible with both NetBSD and FreeBSD based ovf artifacts.

brkt_cli/esx/esx_service.py
krishnanrs: If I recall correctly, in the current format, there is no "latest" image/bucket. We simply sort all the buckets based on the "release-xx-xx" prefix.

apoorva-jindal:
LGTM

brkt_cli/esx/esx_service.py
dasein: Just to follow up and confirm what we talked about. We retrieve the images sorted by last_modified so that we can support both NetBSD images (published under release) and FreeBSD images (published under latest) in the same bucket. However, we _cannot_ expect to serve an 'official' release for _both_ products from the same bucket. This means we would only promote prod and stage metavisor releases to their respective buckets for the images we expect to make current.

PR #588, opened 2017-02-28
Handle vCenter connection expiry

apoorva-jindal:
If a command takes a lot of time to execute, the vCenter connection
can go stale and expire. Handle this by checking the validity of
the connection before each function call.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -25,6 +25,7 @@ import requests
 from functools import wraps
 from operator import attrgetter
 from threading import Thread
+from httplib import BadStatusLine
 
 import boto3
 from botocore.handlers import disable_signing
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -354,6 +359,23 @@ def connected(self):
...
         return True
 
+    def validate_connection(self):
+        try:
+            content = self.si.RetrieveContent()
+            self.__get_obj(content, [vim.VirtualMachine], None)
+        except BadStatusLine:
+            self.si = None
+            try:
+                log.info("vCenter connection expired, reconnecting")
```
bburtin: Should this be logged at debug level?  Seems like we should only log status or things that users would want to react to.

apoorva-jindal: The CLI may seem hung while the vCenter re-connection is happening and suddenly connection related messages may start coming in case any connection error occurs. This may help user understand what is going on. I can make it debug as well, let me know what you recommend.

bburtin: Sure, that's fine.  If the user sees unexpected hangs, go ahead and leave it at info.

krishnanrs:
LGTM

PR #591, opened 2017-03-01
Share-logs GCE update

markvoll:
User is now notified if the selected bucket
is not in the correct project or can not be created.


brkt_cli/gce/gce_service.py
```diff
@@ -219,52 +219,32 @@ def __init__(self, project, session_id, logger):
...
         except errors.HttpError as e:
-            code = e.resp.status
-            if code == 404:
-                # bucket doesn't exist. Check syntax
-                self.validate_bucket_name(bucket)
-            elif code == 403:
-                # bucket does exist. Permisions not valid
-                raise ValidationError("Permission denied for bucket %s" % bucket)
+            message = json.loads(e.content)["error"]["message"]
+            if 'You already own this bucket' in message:
```
bburtin: Does GCE give you an error code?  It's a bit brittle to make assumptions about the message text.

brkt_cli/gce/gce_service.py
```diff
@@ -219,52 +219,32 @@ def __init__(self, project, session_id, logger):
...
+                return
             else:
-                # unexpected Http error
+                self.log.warn("Bucket must be in project: %s", project)
```
bburtin: Should this be `error`?  `warn` means that you're able to proceed, despite the fact that something unexpected happened.

brkt_cli/gce/gce_service.py
```diff
@@ -219,52 +219,32 @@ def __init__(self, project, session_id, logger):
...
     def validate_file_name(self, file):
-        m = re.match(r'[*?#]', file)
+        m = re.search(r'[*?#]', file)
         if m:
             raise ValidationError("File name can't contain '*','?'' or '#'")
-        if "[" or "]" in file:
-            raise ValidationError("File name cant contain '['' or ']'")
-        return 
+        if "]" in file or "[" in file:
+            raise ValidationError("File name cant contain '[' or ']'")
```
bburtin: Typo.  Also, what about putting `[` and `]` into the previous check?  Seems like there's no reason to special-case them.

markvoll: Consolidated into one check

brkt_cli/gce/gce_service.py
markvoll: Yes, switched to error

brkt_cli/gce/gce_service.py
markvoll: GCE does give an error code (409) but it gives the same error code if you own the bucket or you don't, only the messages are different.

PR #595, opened 2017-03-01
Consolidate the code that saves the config file

bburtin:
Remove ConfigSubcommand._write_config() and move the code into
CLIConfig.save_config().  Before this change, we created ~/.brkt only
when calling ConfigSubcommand.  Callsites that called
CLIConfig.save_config() would fail when ~/.brkt didn't exist.

### DIFF
--- a/brkt_cli/config/__init__.py
+++ b/brkt_cli/config/__init__.py
@@ -325,9 +325,18 @@ class CLIConfig(object):
     def save_config(self):
         """Save the current config to disk.
         """
+        try:
+            os.mkdir(CONFIG_DIR, 0755)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
         f = tempfile.NamedTemporaryFile(delete=False, prefix='brkt_cli')
-        yaml.dump(self._config, f)
-        f.close()
+        try:
+            self.write(f)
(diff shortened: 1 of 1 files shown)


brkt_cli/config/__init__.py
```diff
@@ -325,9 +325,18 @@ def write(self, f):
     def save_config(self):
         """Save the current config to disk.
         """
+        try:
+            os.mkdir(CONFIG_DIR, 0755)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
         f = tempfile.NamedTemporaryFile(delete=False, prefix='brkt_cli')
-        yaml.dump(self._config, f)
```
bonzotronix: Where do we now dump the self._config YAML to f?

bburtin:
I decided not to write unit tests, because `~/.brkt/config` is hardcoded.  I guess I could generalize the code, but I'm not sure that it's worth the added complexity.

brkt_cli/config/__init__.py
bburtin: `self.write(f)` below

bonzotronix: Ahh, thanks.

PR #598, opened 2017-03-06
Support new encryptor image name format

krishnanrs:
The encryptor image name is used to determine valid images for AWS.
The image name is also used to determine the encryptor type (NetBSD
vs. FreeBSD) to make certain decisions (for cryptor policy and
sriovNetSupport). The image name format for FreeBSD encryptor
image was changed from brkt-avatar-freebsd-<date> to
metavisor-<branch>-<hash>. This change supports the new image name
format, but ignores the ealier image name format (brkt-avatar-freebsd).

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -234,7 +234,7 @@ def run_encrypt(values, config, verbose=False):
 
     mv_image = aws_svc.get_image(encryptor_ami)
     if values.crypto is None:
-        if mv_image.name.startswith('brkt-avatar-freebsd'):
+        if mv_image.name.startswith('metavisor'):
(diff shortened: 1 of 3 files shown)


xcllnt:
It's probably best to match both "brkt-avatar-freebsd-* and "metavisor-*". This to remain compatibility with previous image names.

krishnanrs:
I don't see any reason to match "brkt-avatar-freebsd-*" and have to remove it later on, especially given the fact that we haven't even release FreeBSD MV for Beta yet. Most folks anyways use the latest encryptor and the only downside of not supporting the old image name format is that the encrypted images will not have sriovNet support enabled (if the source image had it enabled).

PR #603, opened 2017-03-09
Get API token from an environment variable

bburtin:
Update config.get_yeti_service() to get the API token from the
BRKT_API_TOKEN environment variable instead of the user's config.
Turned out that storing the API token in the config file isn't
convenient for scripting.  With an environment variable, the workflow is
the same for end users and scripts.

Update brkt auth to read the default API URL from config instead of
hardcoding to the production service.

### DIFF
--- a/brkt_cli/config/__init__.py
+++ b/brkt_cli/config/__init__.py
@@ -348,7 +348,10 @@ def _get_yeti_service(parsed_config):
     _, env = parsed_config.get_current_env()
     root_url = 'https://%s:%d' % (
         env.public_api_host, env.public_api_port)
-    token = parsed_config.get_option('api-token')
+    token = (
+        os.environ.get('BRKT_API_TOKEN') or
+        parsed_config.get_option('api-token')
+    )
     return YetiService(root_url, token=token)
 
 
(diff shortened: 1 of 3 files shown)


bburtin:
```
$ ./brkt make-token
$BRKT_API_TOKEN is not set. Run brkt auth to get an API token.

$ export BRKT_API_TOKEN=$(./brkt auth --email test@example.com --password test123)

$ ./brkt make-token
eyJhbGciOiJFUzM4NCIsImtpZCI6Ijg0NGQyYzEwZDk0ODQ0ZmMxNWIwNGU5YTQxZjgxYWMyMTc1YWM3MGI5MjZhNjc5MDliNDNhMmViOWJmYWIyMmYiLCJ0eXAiOiJKV1QifQ.eyJleHAiOjE1MjA2NDI0MzQsImlhdCI6MTQ4OTEwNjQzNCwiaXNzIjoiQnJhY2tldCBDb21wdXRpbmcgSW5jLiIsImp0aSI6IjI3NDM5MDIxNzc3MjQwOTQwNTEifQ.lbswLw9Cq_Ky4cga2iW5E2BCNDUOGDVDsjZbV1jDx0QcW9mnE-bhWzwpuhpQwQfnK64OhvIwbuzk9Jmbzy7WWWk-963ZHA8UkheBeoNbE_4U7WL9WHoTNU9ZfgKYBz1G
```

brkt_cli/config/__init__.py
```diff
@@ -361,14 +364,14 @@ def get_yeti_service(parsed_config):
...
-            'Not logged in.  Please run brkt config login.')
+            '$BRKT_API_TOKEN is not set. Run brkt auth to get an API token.')
 
     try:
         y.get_customer()
     except YetiError as e:
         if e.http_status == 401:
             raise ValidationError(
-                'Config API token is not authorized to access %s' %
+                '$BRKT_API_TOKEN is not authorized to access %s' %
```
krishnanrs: Since the token can be passed as an OS env or an argument, would it be better to make this error message more generic?

bburtin: The current plan is to remove the API token from config, so that there's one way of specifying it.  I'll do this in a separate commit.

PR #604, opened 2017-03-10
Handle long clone VM times

apoorva-jindal:
If a VM takes too much time to clone, the find vm object
operation may fail. Handle the exceptions and retry.


brkt_cli/esx/esx_service.py
```diff
@@ -730,6 +730,13 @@ def clone_vm(self, vm, powerOn=False, vm_name=None, template=False):
             vm_name = "template-vm-" + timestamp
         task = vm.Clone(folder=destfolder, name=vm_name, spec=clonespec)
         self.__wait_for_task(task)
+        try:
+            vm = self.__get_obj(content, [vim.VirtualMachine], vm_name)
+            return vm
+        except BadStatusLine, vmodl.fault.ManagedObjectNotFound:
+            pass
```
bburtin: Would it be helpful to log a debug message here?

PR #605, opened 2017-03-13
Don't validate the launch token

bburtin:
Remove the code that validates a launch token by using it to
authenticate with the Bracket service.  This validation will break once
we decouple API tokens and launch tokens.


bburtin:
In the common case, launch tokens will be generated by Yeti, so they won't need to be validated.  If we need to validate user-generated launch tokens in the future, we'll need to add a separate Yeti endpoint and update the CLI to call it.

PR #606, opened 2017-03-13
Get launch token from Bracket service during encryption

bburtin:
Update the encrypt and update commands to get the Metavisor launch token
from the Bracket service if it's not specified on the command line.

With this change, the CLI will always send a launch token to Metavisor
when encrypting or updating.  If the launch token is not specified with
the --token option, we will now require that the $BRKT_API_TOKEN
environment variable be set.  This allows the CLI to authenticate and
get a launch token from the Bracket service.

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -21,6 +21,7 @@ from cryptography.hazmat.backends import default_backend
 import brkt_cli
 from brkt_cli import (
     add_brkt_env_to_brkt_config,
+    config,
     encryptor_service,
     get_proxy_config
 )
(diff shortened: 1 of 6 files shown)


brkt_cli/instance_config_args.py
```diff
@@ -135,13 +136,15 @@ def setup_instance_config_args(parser, parsed_config,
...
-                                cli_config=None):
+                                cli_config=None, launch_token=None):
     """ Return an InstanceConfig object, based on options specified on
     the command line and Metavisor mode.
 
     :param values an argparse.Namespace object
     :param mode the mode in which Metavisor is running
     :param cli_config an brkt_cli.config.CLIConfig instance
+    :param launch_token the token that Metavisor will use to authenticate
+    with Yeti.  If not specified, use values.token.
```
bonzotronix: Would it be preferable to have the new arg be a boolean `require_token` and have `instance_config_from_values` call `_get_launch_token()` if the token is not found in `values`?

bburtin: Yeah, that might be an improvement.  Let me try it and see what the code looks like.

bonzotronix: Per our discussion, it's fine to leave the API defined as is.

PR #608, opened 2017-03-14
Add --brkt-tags to encrypt and update commands

bburtin:
Add a --brkt-tags option to the following commands:

* aws encrypt
* aws update
* gcp encrypt
* gcp update
* gcp launch
* vmware encrypt-with-vcenter
* vmware encrypt-with-esx-host
* vmware update-with-vcenter
* vmware update-with-esx-host

When one or more tags is specified with the --brkt-tag option, we pass
the tags to the Bracket service at the time we generate the launch
token.  Yeti embeds the tags into the generated JWT as claims.

### DIFF
--- a/brkt_cli/argutil.py
+++ b/brkt_cli/argutil.py
@@ -25,3 +25,16 @@ def add_out(parser):
             'character encoding issues when redirecting output on Windows.'
         )
     )
+
+
+def add_brkt_tag(parser):
+    parser.add_argument(
+        '--brkt-tag',
+        metavar='NAME=VALUE',
+        dest='brkt_tags',
+        help=(
+            'Bracket tag which will be embedded in the JWT as a claim.  All '
+            'characters must be alphanumeric or [-_.].  The tag name cannot '
+            'be a JWT registered claim name (see RFC 7519).'),
+        action='append'
+    )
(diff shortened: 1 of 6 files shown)


brkt_cli/argutil.py
```diff
@@ -25,3 +25,16 @@ def add_out(parser):
             'character encoding issues when redirecting output on Windows.'
         )
     )
+
+
+def add_brkt_tag(parser):
```
bonzotronix: Could we move this to `setup_instance_config_args()` and put it in a mutually_exclusive_group with `--token`?

bburtin: I like the idea of the mutually exclusive group.  Let me see if I can make that work.

I can't move the call into `setup_instance_config_args()` because `make-user-data` doesn't need it.  I'd rather not add another switch to `setup_instance_config_args()`, since it's already pretty complicated.

brkt_cli/make_user_data/__init__.py
```diff
@@ -125,7 +125,11 @@ def register(self, subparsers, parsed_config):
         )
 
         setup_instance_config_args(
-            parser, parsed_config, mode=INSTANCE_METAVISOR_MODE)
+            parser,
+            parsed_config,
+            mode=INSTANCE_METAVISOR_MODE,
+            brkt_tag=False
```
bonzotronix: These latest changes LGTM.
Can I ask that we add a comment near here, documenting the decision that `make-user-data` should only take a `--token input` (optionally) and never a `--brkt-tag` input because we think the customer running `make-user-data` will never want it to automatically get a token from Yeti?

bburtin: Will do.

PR #609, opened 2017-03-14
Set no-teardown only on error

apoorva-jindal:


bburtin:
Looks good to me, but someone who uses this should approve.  Or maybe you're the target user?  :-)

Garpetun:
LGTM!

PR #610, opened 2017-03-15
Provide ability to launch Bracket "wrapped" instance

krishnanrs:
A Bracket wrapped instance consists of a Metavisor which wraps the
guest root volume without encrypting it. This greatly speeds up the
process of deploying Bracket instances, providing all the benefits
of the Bracket Metavisor, without an encrypted guest root volume.

A new `wrap-guest-image` command has been added to all the three supported
CSPs which given a guest "image", would wrap the guest image with
the Bracket metavisor image and launch an instance using the same.
The launched instance should show up on your Bracket Management
Console (UI) within a few minutes. All Bracket supported features
including disk encryption (of attached disks), network policy and
network segmentation would be supported by such an instance.

The recommendation is once the instance is launched, the user create
an image based off the running instance for subsequent use as part
of their Bracket infrastructure.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -29,6 +29,8 @@ from brkt_cli.aws import (
     diag_args,
     encrypt_ami,
     encrypt_ami_args,
+    wrap_image,
+    wrap_image_args,
     share_logs,
     share_logs_args,
     update_encrypted_ami_args
@@ -41,6 +43,7 @@ from brkt_cli.aws.update_ami import update_ami
 from brkt_cli.instance_config import (
     INSTANCE_CREATOR_MODE,
     INSTANCE_UPDATER_MODE,
+    INSTANCE_METAVISOR_MODE,
 )
 from brkt_cli.instance_config_args import (
     instance_config_from_values,
@@ -189,6 +192,59 @@ def run_share_logs(values):
         snapshot_id=values.snapshot_id,
         bracket_aws_account=values.bracket_aws_account
     )
+    return 0
+
+
+@_handle_aws_errors
+def run_wrap_image(values, config, verbose=False):
+    nonce = util.make_nonce()
+
+    aws_svc = aws_service.AWSService(
+        nonce,
+        retry_timeout=values.retry_timeout,
+        retry_initial_sleep_seconds=values.retry_initial_sleep_seconds
+    )
+    log.debug(
+        'Retry timeout=%.02f, initial sleep seconds=%.02f',
+        aws_svc.retry_timeout, aws_svc.retry_initial_sleep_seconds)
+
+    aws_svc.connect(values.region, key_name=values.key_name)
+
+    if values.validate:
+        guest_image = _validate_guest_ami(aws_svc, values.ami)
+    else:
+        guest_image = aws_svc.get_image(values.ami)
+
+    metavisor_ami = values.encryptor_ami or _get_encryptor_ami(values.region)
+    if values.validate:
+        values.encrypted_ami_name = None
+        _validate(aws_svc, values, metavisor_ami)
+        brkt_cli.validate_ntp_servers(values.ntp_servers)
+
+    mv_image = aws_svc.get_image(metavisor_ami)
+    # Raise error if it is not a FreeBSD Metavisor
+    if 'metavisor-' not in mv_image.name:
+        raise ValidationError(
+            'Unsupported Bracket image for wrapped guest: %s' %
+            mv_image.name
+        )
+
+    instance_config = instance_config_from_values(
+        values, mode=INSTANCE_METAVISOR_MODE, cli_config=config)
+
+    instance_id = wrap_image.launch_wrapped_image(
+        aws_svc=aws_svc,
+        image_id = guest_image.id,
+        metavisor_ami=metavisor_ami,
+        wrapped_instance_name=values.wrapped_instance_name,
+        subnet_id=values.subnet_id,
+        security_group_ids=values.security_group_ids,
+        instance_type=values.instance_type,
+        instance_config=instance_config
+    )
+    # Print the Instance ID to stdout, in case the caller wants to process
+    # the output. Log messages go to stderr
+    print(instance_id)
     return 0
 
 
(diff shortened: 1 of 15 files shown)


krishnanrs:
Will update docs later. For now have added atleast some unit tests for the changes. Will review the naming convention for the commands with PM.

brkt_cli/esx/wrap_image.py
```diff
@@ -0,0 +1,119 @@
...
+#
+# https://github.com/brkt/brkt-cli/blob/master/LICENSE
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+"""
+Create an encrypted VMDK based on an existing unencrypted VMDK.
```
apoorva-jindal: Fix the description of the file.

brkt_cli/esx/wrap_image.py
```diff
@@ -0,0 +1,119 @@
...
+            raise Exception("Invalid MV OVF")
+        vm = launch_mv_vm_from_s3(vc_swc, ovf_name,
+                                  download_file_list, vm_name,
+                                  cleanup)
+        guest_vmdk_path = vc_swc.get_datastore_path(guest_vmdk)
+        vc_swc.add_disk(vm, filename=guest_vmdk_path, unit_number=1)
+        if user_data_str:
+            vc_swc.send_userdata(vm, user_data_str)
+        vc_swc.power_on(vm)
+        ip_addr = vc_swc.get_ip_address(vm)
```
krishnanrs: Want to keep it consistent across all clouds and do not want to create a template. I would prefer that we leave it up to the user.

apoorva-jindal: Then I will recommend to change "--template-vm-name" argument in the input to just "--output-vm-name" or something similar.

krishnanrs: Please review the updated changes

brkt_cli/aws/wrap_image.py
```diff
@@ -0,0 +1,166 @@
...
+def get_name_from_image(image):
+    name = append_suffix(
+        image.name,
+        get_wrapped_suffix(),
+        max_length=INSTANCE_NAME_MAX_LENGTH
+    )
+    return name
+
+
+def create_instance_security_group(aws_svc, vpc_id=None):
```
bburtin: Can you add a short comment that describes the reason for the security group?

brkt_cli/aws/wrap_image.py
```diff
@@ -0,0 +1,166 @@
...
+    log.info('Created security group with id %s', sg.id)
+    try:
+        aws_svc.add_security_group_rule(
+            sg.id, ip_protocol='tcp',
+            from_port=22,
+            to_port=22,
+            cidr_ip='0.0.0.0/0')
+    except Exception as e:
+        log.error('Failed adding security group rule to %s: %s', sg.id, e)
+        try:
```
bburtin: I think you can make this a one-liner by calling `encrypt_ami.clean_up()` (should actually be in the `aws` module).

brkt_cli/aws/wrap_image.py
```diff
@@ -0,0 +1,166 @@
...
+                         instance_config=None):
+    temp_sg_id = None
+    guest_image = aws_svc.get_image(image_id)
+    guest_root_device = guest_image.root_device_name
+    guest_snapshot = guest_image.block_device_mapping[guest_root_device].snapshot_id
+    mv_image = aws_svc.get_image(metavisor_ami)
+
+    # Raise error if it is not a FreeBSD Metavisor
+    if 'metavisor-' not in mv_image.name:
+        raise BracketError(
```
bburtin: This looks like validation code.  Would it make sense to move it up a level and raise `ValidationError`?

brkt_cli/esx/wrap_with_esx_host_args.py
```diff
@@ -0,0 +1,181 @@
...
+        default="8",
+        required=False)
+    parser.add_argument(
+        "--memory",
+        help="Memory to assign to Encryptor VM",
+        metavar='GB',
+        dest="memory_gb",
+        default="32",
+        required=False)
+    parser.add_argument(
```
apoorva-jindal: Do you want to call it output-vm-name or vm-name? I will leave it to your judgement. The ESX changes look good to me.

brkt_cli/aws/wrap_image.py
krishnanrs: I've made all the recommended changes. Will make further changes once your merge your AWS args changes.

brkt_cli/aws/__init__.py
```diff
@@ -193,6 +196,59 @@ def run_share_logs(values):
...
+
+    metavisor_ami = values.encryptor_ami or _get_encryptor_ami(values.region)
+    if values.validate:
+        values.encrypted_ami_name = None
+        _validate(aws_svc, values, metavisor_ami)
+        brkt_cli.validate_ntp_servers(values.ntp_servers)
+
+    mv_image = aws_svc.get_image(metavisor_ami)
+    # Raise error if it is not a FreeBSD Metavisor
+    if 'metavisor-' not in mv_image.name:
```
bburtin: Should this be in the `if values.validate` block?

brkt_cli/aws/wrap_image_args.py
```diff
@@ -0,0 +1,75 @@
...
+    parser.add_argument(
+        '-v',
+        '--verbose',
+        dest='aws_verbose',
+        action='store_true',
+        help=argparse.SUPPRESS
+    )
+    # Hide deprecated --tag argument
+    parser.add_argument(
+        '--tag',
```
bburtin: Do we need `-v` and `--tag` in the new command?  If we need them in order for shared argument parsing code to work, go ahead and leave them in.

brkt_cli/aws/__init__.py
krishnanrs: I want it outside the block. The validate option checks whether all the arguments you entered was correct, whereas we do not want to attempt this operation on non-FreeBSD MV since otherwise the instance will get launched, but would be unusable. In other words, don't want a case where we are launching resources which end up being unusable.

brkt_cli/aws/wrap_image_args.py
krishnanrs: I left it there to be on par with the existing AWS commands. However if you think they should be removed, let me know.

bburtin: Yeah, let's take them out.  The others are only there to maintain backward compatibility.  Thanks!

brkt_cli/aws/__init__.py
bburtin: Makes sense.

PR #612, opened 2017-03-16
Add --base64 arg to make-user-data

bonzotronix:
* In the case of MV instances on ESX, the userdata blob must be base64-encoded.  Add an optional argument to make-user-data to encode the output.


brkt_cli/make_user_data/__init__.py
```diff
@@ -135,6 +138,13 @@ def register(self, subparsers, parsed_config):
             help=argparse.SUPPRESS
         )
         parser.add_argument(
+            '-b',
```
bburtin: Can we skip the `-b`?  We haven't used it anywhere else, and base64 is an edge case.

brkt_cli/make_user_data/test_make_user_data.py
```diff
@@ -35,6 +36,17 @@ def setUp(self):
         self.testdata_dir = os.path.join(my_dir, 'testdata')
         self.maxDiff = None # show full diff with knowngood multi-line strings
 
+    def check_output_vs_knowngood(self, output, knowngood_file):
+        if not os.path.exists(knowngood_file):
```
bburtin: Should this be `self.assertTrue(os.path.exists(known_good_file, msg=...)`?

bonzotronix: Sorry I missed this earlier.  Fixed now in latest commit.

PR #615, opened 2017-03-17
Hide the encryptor image option for GCP

krishnanrs:
The encryptor_image is an optional argument for GCP commands which
is primarily used for internal testing or when a user has a local
encryptor image in their GCP account. As a result, it should be
hidden so as to not confuse users and also make it consistent
with similar commands for other CSPs.

Also consolidated the common GCP arguments into a single file.


bburtin:
Do you want to use this opportunity to start a `gcp_args` module which adds arguments in a consistent way?

krishnanrs:
Done. Take a look again.

PR #616, opened 2017-03-20
Increase retry timeout for GCE instances

krishnanrs:
Increase the retry timeout for GCE instance to avoid occasional
failures.


robertabbott:
lgtm

PR #619, opened 2017-03-21
Use the correct brkt-env

krishnanrs:
Use the user passed brkt-env if available, before defaulting to the
default environment (or the environment set in `brkt config`)


bburtin:
I don't think it's a good idea to pass the `values` object down into utility functions, since it represents the command line and is loosely typed.  Seems like it would be cleaner to take the opposite approach: pass the `CLIConfig` object into `brkt_cli.brkt_env_from_values()`.

PR #620, opened 2017-03-22
Prefer user-passed brkt-env, if available

krishnanrs:
With the recent changes to support brkt-tags, we always defauled
to the environment specified in the `brkt config` settings, ignoring
the user passed environment. This change picks the user specified
environment (if available) before defaulting the the `brkt config`
environment settings.

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -226,7 +226,7 @@ def instance_config_from_values(values=None, mode=INSTANCE_CREATOR_MODE,
     return ic
 
 
-def get_launch_token(values, cli_config):
+def get_launch_token(values, cli_config, brkt_env=None):
     """ Return the launch token either from values.token or from Yeti, in that
     order.  Assume that the values.token and values.brkt_tags fields exist.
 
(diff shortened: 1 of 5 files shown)


brkt_cli/instance_config_args.py
```diff
@@ -226,7 +226,7 @@ def instance_config_from_values(values=None, mode=INSTANCE_CREATOR_MODE,
     return ic
 
 
-def get_launch_token(values, cli_config):
+def get_launch_token(values, cli_config, brkt_env=None):
```
bburtin: Seems like it would be cleaner to just pass the `BracketEnvironment` and let the caller deal with figuring out where it comes from?

krishnanrs: The config also needs to be passed since the API token can be specified via a config argument. We will have to move that logic further up otherwise.
`    token = (
        os.environ.get('BRKT_API_TOKEN') or
        parsed_config.get_option('api-token')
    )
`

bburtin: Right.  The caller would be responsible for getting the `BracketEnvironment` from the command line or config.  Then `get_launch_token()` uses the given environment to get the launch token.

krishnanrs: Maybe I am not understanding your comment. The `BracketEnvironment` object is being passed. I just making it an optional argument so that I do not break other parts which rely on the same underlying code.

PR #623, opened 2017-03-24
Remove dependency on the cryptography library

bburtin:
Remove the dependency on the cryptography library, to make installation
easier.  Most users can either specify Bracket tags during encryption
time or use make-token to get a launch token from the Bracket service.
Users who need to generate tokens from private keys will need to run
"pip install cryptography".

Add brkt_cli.crypto.cryptography_library_available, to allow, callsites
to know whether to attempt operations that require cryptography.  Update
callsites and unit tests to check this variable.

Update the cert validation code to shell out to openssl when the
cryptography library isn't available.

Don't initialize the make-key and get-public-key commands when the
cryptography library isn't available.

Update the Travis configuration to run the test suite with and without
the cryptography library.


brkt_cli/crypto/__init__.py
```diff
@@ -138,7 +152,33 @@ def validate_cert(cert_data):
...
+    log.debug('Using openssl to validate cert')
+    key_file = tempfile.NamedTemporaryFile(delete=False)
+    key_file.write(cert_data)
+    key_file.close()
+
     try:
-        x509.load_pem_x509_certificate(cert_data, default_backend())
-    except Exception as e:
-        raise ValidationError('Error validating CA cert: %s' % e)
+        subprocess.check_output(['openssl', 'x509', '-in', key_file.name])
```
bonzotronix: Do we have to redirect stderr here?

brkt_cli/brkt_jwt/__init__.py
```diff
@@ -123,6 +123,12 @@ def run(self, values):
         signing_key = values.signing_key or values.signing_key_option
```
bonzotronix: Should we add a check for cryptography_library_available in setup_make_jwt_args(), so it doesn't add the arg if cryptography is not installed?  Or add a note to that option's help text a la "requires cryptography package"?

bonzotronix:
FTP.  Looks good.  Just a couple questions.

brkt_cli/brkt_jwt/__init__.py
bburtin: I like that.  Let me see if I can make it work.

brkt_cli/crypto/__init__.py
bburtin: Good catch.  `openssl` writes to `stderr` when validation fails.  I'll fix it.

bburtin:
FTP?  I didn't realize you're an NWA fan.

brkt_cli/crypto/__init__.py
```diff
@@ -132,13 +144,45 @@ def read_private_key(pem_path):
...
+    The output contains both stdout and stderr.
+    """
+    p = subprocess.Popen(
+        args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    if input_content:
+        p.stdin.write(input_content)
+        p.stdin.close()
+    p.wait()
+    return p.returncode, p.stdout.read()
+
```
bonzotronix: At some point we should verify that this works (or at least, fails gracefully) when running brkt-cli on Windows.

Also, we might want to move _run_cmd to a public function in util.py, but that can happen later.

PR #626, opened 2017-03-28
Check for user provided and auth tokens

krishnanrs:
ESX and GCP operations require a valid token. However the token
could either be user provided as a `--token` argument or a Bracket
generated one by setting the `BRKT_API_TOKEN` argument. Hence we
should validate atleast one of them is set for these operations.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -108,8 +108,8 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
     brkt_env = brkt_cli.brkt_env_from_values(values)
     if brkt_env is None:
         _, brkt_env = parsed_config.get_current_env()
-    if not values.token:
-        raise ValidationError('Must provide a token')
+    # Verify we have a valid launch token
+    instance_config_args.get_launch_token(values, parsed_config)
 
     proxy = None
     if values.http_proxy:
@@ -295,8 +295,8 @@ def run_update(values, parsed_config, log, use_esx=False):
     brkt_env = brkt_cli.brkt_env_from_values(values)
     if brkt_env is None:
         _, brkt_env = parsed_config.get_current_env()
-    if not values.token:
-        raise ValidationError('Must provide a token')
+    # Verify we have a valid launch token
+    instance_config_args.get_launch_token(values, parsed_config)
 
     proxy = None
     if values.http_proxy:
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -108,7 +108,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
     brkt_env = brkt_cli.brkt_env_from_values(values)
     if brkt_env is None:
         _, brkt_env = parsed_config.get_current_env()
-    if not values.token:
+    if not values.token and not os.environ.get('BRKT_API_TOKEN'):
```
bburtin: I think you want `instance_config_args.get_launch_token(values, config)`?  That takes care of handling the token when it comes from either the command line or from Yeti.

krishnanrs: Can you take a look again?

brkt_cli/esx/__init__.py
```diff
@@ -108,7 +108,8 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
     brkt_env = brkt_cli.brkt_env_from_values(values)
     if brkt_env is None:
         _, brkt_env = parsed_config.get_current_env()
-    if not values.token:
+    token = instance_config_args.get_launch_token(values, parsed_config)
+    if not token:
```
bburtin: This check isn't necessary.  `get_launch_token()` will always return a token.

krishnanrs: The get_launch_token will return a token only if either values.token or values.brkt_tags are set. I believe we have a fundamental problem here that we should be able to launch an instance if either a JWT token or API token is specified (so long as either of them is valid) and neither of them need to be associated with any brkt claims.

brkt_cli/gcp/__init__.py
```diff
@@ -454,8 +454,7 @@ def check_args(values, gcp_svc, cli_config):
         if values.bucket != 'prod':
             raise ValidationError(
                 "Please provide either an encryptor image or an image bucket")
-    if not values.token:
-        raise ValidationError('Must provide a token')
+    instance_config_args.get_launch_token(values, cli_config)
```
bburtin: It looks like you're getting a token and dropping it on the floor.  If this is this a validation step, please add a comment so that it's more obvious why you're calling the function.

krishnanrs: I added the comment to the esx changes, but this one got skipped.

PR #628, opened 2017-03-30
Allow configuration of static IP on the encryptor VM

apoorva-jindal:
Use vCenter customization APIs to configure static IP address,
default router, subnet mask, DNS domain and DNS servers.

### DIFF
--- a/brkt_cli/esx/encrypt_vmdk_args.py
+++ b/brkt_cli/esx/encrypt_vmdk_args.py
@@ -94,6 +94,42 @@ def setup_encrypt_vmdk_args(parser):
         required=False
     )
     parser.add_argument(
+        '--static-ip-address',
+        metavar='IP',
+        dest='static_ip',
+        help='Specify the static IP address of the encryptor VM',
+        required=False
+    )
+    parser.add_argument(
+        '--static-subnet-mask',
+        metavar='IP',
+        dest='static_mask',
+        help='Specify the static subnet mask of the encryptor VM',
+        required=False
+    )
+    parser.add_argument(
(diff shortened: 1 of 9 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -79,6 +81,43 @@ def compute_sha1_of_file(filename):
...
+                                  self.mask)
+        if not validate_ip_address(self.gw):
+            raise ValidationError("Default router %s is not in IP address "
+                                  "format", self.gw)
+        for server in self.dns:
+            if not validate_ip_address(server):
+                raise ValidationError("DNS server %s is not in IP address "
+                                      "format", server)
+        if not validate_dns_name_ip_address(self.dns_domain):
+            raise ValidationError("DMNS domain %s is invalid ",
```
bburtin: typo?

brkt_cli/esx/encrypt_vmdk_args.py
```diff
@@ -94,6 +94,42 @@ def setup_encrypt_vmdk_args(parser):
         required=False
     )
     parser.add_argument(
+        '--static-ip-address',
```
bburtin: Krishnan and I have started putting common args for AWS and GCP into the `aws_args` and `gcp_args` modules.  This ensures that an argument that appears in multiple commands is added in a uniform way.  Would it make sense to do something similar for VMware?

apoorva-jindal: That makes sense to me. I think we should start doing something similar for VMware too.

PR #631, opened 2017-03-30
Create guest image snapshot when lacking permission

krishnanrs:
The wrap-image command (for AWS) made an assumption that the user
will have permission to launch the guest image snapshot. This is
true if the image is owned by the account where command is being
run. However in some cases, especially with the marketplace images,
the user might not have "create volume" permissions on the guest
root image snapshot. In such cases, the wrap command should:
	* Launch the guest image
	* Stop the guest instace
	* Snapshot the guest root volume
	* Terminate guest instance
	* Launch the created snapshot wrapped with the Metavisor
          image

### DIFF
--- a/brkt_cli/aws/test_aws_service.py
+++ b/brkt_cli/aws/test_aws_service.py
@@ -381,6 +381,9 @@ def build_aws_service():
     # Guest image
     bdm = BlockDeviceMapping()
     bdm['/dev/sda1'] = BlockDeviceType(snapshot_id='snap-12345678')
+    snapshot = Snapshot()
+    snapshot.id = 'snap-12345678'
+    aws_svc.snapshots[snapshot.id] = snapshot
     id = aws_svc.register_image(name='Guest image', block_device_map=bdm)
     guest_image = aws_svc.get_image(id)
 
--- a/brkt_cli/aws/wrap_image.py
+++ b/brkt_cli/aws/wrap_image.py
@@ -18,6 +18,8 @@ Create an Bracket wrapped based on an existing unencrypted AMI.
 Overview of the process:
     * Obtain the Bracket (metavisor) image to be used
     * Obtain the root volume snapshot of the guest image
+    * When lacking "create volume" permissions on the guest image
+        create a local snapshot of the guest image
     * Configure the Bracket image to be launched with the guest
         root volume attached at /dev/sdf
     * Pass appropriate user-data to the Bracket image to indicate
@@ -31,6 +33,7 @@ running the AWS command line utility.
 
 import logging
 
+from boto.exception import EC2ResponseError
 from boto.ec2.blockdevicemapping import (
     BlockDeviceMapping,
     EBSBlockDeviceType,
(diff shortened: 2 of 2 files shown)


brkt_cli/aws/wrap_image.py
```diff
@@ -114,6 +122,24 @@ def launch_wrapped_image(aws_svc, image_id, metavisor_ami,
     else:
         instance_name = get_name_from_image(guest_image)
 
+    try:
+        aws_svc.get_snapshot(guest_snapshot)
+    except EC2ResponseError as e:
+        if e.error_code == 'InvalidSnapshot.NotFound':
```
bburtin: Would this lead to a misleading error message if the snapshot ID is invalid?

brkt_cli/aws/wrap_image.py
```diff
@@ -114,6 +122,24 @@ def launch_wrapped_image(aws_svc, image_id, metavisor_ami,
...
+            log.error("Insufficient permission to launch guest image. " +  e.error_message)
+            try:
+                guest_instance = run_guest_instance(aws_svc, image_id,
+                    subnet_id=subnet_id, instance_type=instance_type)
+                wait_for_instance(aws_svc, guest_instance.id)
+                guest_snapshot, _, _, _, _ = _snapshot_root_volume(aws_svc,
+                    guest_instance, image_id)
+            finally:
+                if guest_instance:
```
bburtin: I think you need to initialize `guest_instance` to `None` outside the `try`.

brkt_cli/aws/wrap_image.py
```diff
@@ -114,6 +122,24 @@ def launch_wrapped_image(aws_svc, image_id, metavisor_ami,
     else:
         instance_name = get_name_from_image(guest_image)
 
+    try:
+        aws_svc.get_snapshot(guest_snapshot)
+    except EC2ResponseError as e:
+        if e.error_code == 'InvalidSnapshot.NotFound':
+            log.error("Insufficient permission to launch guest image. " +  e.error_message)
```
bburtin: If this is an expected state, it should be `info` instead of `error`.

brkt_cli/aws/test_aws_service.py
```diff
@@ -205,6 +205,12 @@ def get_snapshots(self, *snapshot_ids):
         return [self.get_snapshot(id) for id in snapshot_ids]
 
     def get_snapshot(self, snapshot_id):
+        if snapshot_id == 'snap-12345678':
```
bburtin: Can we have the test put the `Snapshot` object into `self.snapshots`, so that the `get_snapshot()` call succeeds as expected?

brkt_cli/aws/wrap_image.py
krishnanrs: Will do

brkt_cli/aws/wrap_image.py
krishnanrs: Will do

brkt_cli/aws/wrap_image.py
krishnanrs: Snapshot ID cannot be invalid as it is obtained from the guest AMI. If the guest AMI is invalid, then there are checks to handle that.

brkt_cli/aws/test_aws_service.py
krishnanrs: The option was to either put it in this one place or initialize this at multiple places with each of the tests where the aws_svc is being initialized (I'm counting atleast 8). If you are OK with that, will submit a change for that.

bburtin: Oh, I misunderstood.  I thought this was a hack for a single test.  What about putting the `Snapshot` into `self.snapshots` in `test_aws_service.build_aws_service()`?

krishnanrs: Let me know if the new change does this as expected.

PR #635, opened 2017-03-31
Normalize user specified OVF

krishnanrs:
When a user specifies a Metavisor OVF, do not automatically assume
that it does not have the `.ovf` extension. Add the extension only
when it is not present.


apoorva-jindal:
LGTM

PR #636, opened 2017-04-03
Back up the old template till the new template is created

apoorva-jindal:
During update, the old template gets replaced with the new one.
During the transition, the old one gets deleted and and the new
one is added. If something goes wrong during the addition of the
new template, the old template is lost while the new one is not
created either. So, backup the old template, then create a new one,
and finally delete the old template.

### DIFF
--- a/brkt_cli/esx/update_vmdk.py
+++ b/brkt_cli/esx/update_vmdk.py
@@ -90,15 +90,24 @@ def update_ovf_image_mv_vm(vc_swc, enc_svc_cls, guest_vm, mv_vm,
                 print(ovf)
         else:
             # delete the old vm template
-            log.info("Deleting the old template")
             template_vm = vc_swc.find_vm(template_vm_name)
-            if (template_vm):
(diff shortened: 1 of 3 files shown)


apoorva-jindal:
Changed to using the nonce (stored as session-id) to create an unique backup template name.

brkt_cli/esx/update_vmdk.py
```diff
@@ -90,15 +90,24 @@ def update_ovf_image_mv_vm(vc_swc, enc_svc_cls, guest_vm, mv_vm,
...
         else:
             # delete the old vm template
-            log.info("Deleting the old template")
             template_vm = vc_swc.find_vm(template_vm_name)
-            if (template_vm):
-                vc_swc.destroy_vm(template_vm)
+            old_template_vm_name = None
+            if template_vm:
+                old_template_vm_name = template_vm_name + "-" + \
+                                       str(vc_swc.session_id)
```
krishnanrs: No need to cast this to a string since the nonce/sessionID is already a string.

PR #638, opened 2017-04-04
Simplify the logic that handles BracketEnvironment

bburtin:
Move the logic that handles BracketEnvironment from
instance_config_from_values() to the callsites.  We want to default the
environment to prod when encrypting and updating.  We don't want to
default it when launching with an encrypted disk, but we do when
wrapping or launching with an unencrypted disk.  And the environment can
come from either the command line or config.  It was getting hard enough
to keep track of all of these combinations that I decided to move these
decisions up to the command-line parsing code.

instance_config_from_values() now accepts a BracketEnvironment object
instead of CLIConfig.  It's the callsite's responsibility to figure out
where the environment data comes from, and whether it should be
defaulted.

Move get_yeti_service() to instance_config_args, since that's the only
callsite.

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -14,17 +14,16 @@
 
 import argparse
 import logging
+import os
 
 import brkt_cli
+import brkt_cli.crypto
 from brkt_cli import (
-    add_brkt_env_to_brkt_config,
-    config,
     encryptor_service,
-    get_proxy_config
-)
+    get_proxy_config,
+    yeti)
 from brkt_cli import argutil
 from brkt_cli import brkt_jwt
-import brkt_cli.crypto
 from brkt_cli.config import CLIConfig
 from brkt_cli.instance_config import (
     InstanceConfig,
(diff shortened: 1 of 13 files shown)


brkt_cli/instance_config_args.py
```diff
@@ -139,49 +139,33 @@ def setup_instance_config_args(parser, parsed_config,
...
+                                brkt_env=None, launch_token=None):
     """ Return an InstanceConfig object, based on options specified on
     the command line and Metavisor mode.
 
     :param values an argparse.Namespace object
     :param mode the mode in which Metavisor is running
-    :param cli_config an brkt_cli.config.CLIConfig instance
+    :param brkt_env a BracketEnvironment object
     :param launch_token the token that Metavisor will use to authenticate
     with Yeti.  If not specified, use values.token.
```
bonzotronix: That last part is now incorrect.  We allow launch_token=None now.

bburtin: Thanks for catching this.  I'll fix the docstring.

brkt_cli/test_instance_config.py
```diff
@@ -193,23 +198,17 @@ def test_default_brkt_env(self):
...
-            self.assertEqual(
-                '%s:%d' % (prod_env.hsmproxy_host, prod_env.hsmproxy_port),
-                brkt_config.get('hsmproxy_host')
-            )
+        # Verify that we require BracketEnvironment for creator and
+        # updater mode.
+        for _ in (INSTANCE_CREATOR_MODE, INSTANCE_UPDATER_MODE):
+            with self.assertRaises(Exception):
+                values = instance_config_args_to_values('')
+                instance_config_from_values(values)
```
bonzotronix: We should include the mode arg here, right?

bonzotronix:
Looks like MakeTokenSubcommand needs to have its get_yeti_service() call adjusted to the new arg and location.

bonzotronix:
BTW, do the travis tests run unittests both with and without the cryptography pkg installed?

bonzotronix:
FTP.  Code does look cleaner (tests too).  Just a couple questions (in comments).

bburtin:
Thanks for catching the `get_yeti_service()` call thing.  I updated the unit tests to run with and without cryptography in a previous commit.

brkt_cli/test_instance_config.py
bonzotronix: Still thinking we should use the 'mode' here.

bburtin: Sorry, I missed this the first time around.

brkt_cli/test_instance_config.py
```diff
@@ -193,23 +198,17 @@ def test_default_brkt_env(self):
...
-            self.assertEqual(
-                '%s:%d' % (prod_env.api_host, prod_env.api_port),
-                brkt_config.get('api_host')
-            )
-            self.assertEqual(
-                '%s:%d' % (prod_env.hsmproxy_host, prod_env.hsmproxy_port),
-                brkt_config.get('hsmproxy_host')
-            )
+            with self.assertRaises(Exception, msg=mode):
+                values = instance_config_args_to_values('')
```
bonzotronix: I was thinking the mode arg would be given to `instance_config_args_to_values()`.  AFACT, the 'msg' arg is ignored by `assertRaises`.

bburtin: Sorry, I'm dense.  Should be fixed now.

PR #641, opened 2017-04-05
Refactor the encrypt_ami module

bburtin:
Move utility functions that are called from multiple places out of
encrypt_ami and into a new aws_util module.  Move constants into a new
aws_constants module.  Remove the old backward compatibility test.  It's
no longer necessary, now that we have a single CLI codebase.

### DIFF
--- a/brkt_cli/aws/update_ami.py
+++ b/brkt_cli/aws/update_ami.py
@@ -25,8 +25,12 @@ import json
 import logging
 import os
 
-import encrypt_ami
 from brkt_cli import encryptor_service
+from brkt_cli.aws.aws_service import (
+    wait_for_instance, stop_and_wait,
+    wait_for_image, create_encryptor_security_group, clean_up,
+    log_exception_console, wait_for_volume_attached
+)
 from brkt_cli.encryptor_service import (
     wait_for_encryptor_up,
     wait_for_encryption,
@@ -37,18 +41,10 @@ from brkt_cli.instance_config import (
 )
 from brkt_cli.user_data import gzip_user_data
 from brkt_cli.util import Deadline
-from encrypt_ami import (
-    clean_up,
-    create_encryptor_security_group,
-    log_exception_console,
-    wait_for_instance,
-    wait_for_image,
-    DESCRIPTION_GUEST_CREATOR,
-    DESCRIPTION_METAVISOR_UPDATER,
-    NAME_GUEST_CREATOR,
-    NAME_METAVISOR_UPDATER,
-    NAME_ENCRYPTED_ROOT_SNAPSHOT,
-    NAME_METAVISOR_ROOT_SNAPSHOT,
+from brkt_cli.aws.aws_constants import (
+    NAME_GUEST_CREATOR, DESCRIPTION_GUEST_CREATOR, NAME_METAVISOR_UPDATER,
+    DESCRIPTION_METAVISOR_UPDATER, NAME_ENCRYPTED_ROOT_SNAPSHOT,
+    NAME_METAVISOR_ROOT_SNAPSHOT
 )
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


brkt_cli/aws/update_ami.py
```diff
@@ -211,7 +210,7 @@ def update_ami(aws_svc, encrypted_ami, updater_ami, encrypted_ami_name,
             (mv_root_id, encrypted_guest.id)
         )
         aws_svc.attach_volume(mv_root_id, encrypted_guest.id, root_device_name)
-        encrypted_guest = encrypt_ami.wait_for_volume_attached(
+        encrypted_guest = brkt_cli.aws.aws_util.wait_for_volume_attached(
```
bonzotronix: No biggie, but my stylistic preference would be toward adding a `from brkt_cli.aws.aws_util import stop_and_wait, wait_for_volume_attached` and then referencing these two functions directly.

bonzotronix: Or rather, adding those two functions to the import list in the new lines 30-36.

bburtin: Sure, I can fix that.  Most of the code style decisions were made by PyCharm.  😄

brkt_cli/aws/aws_util.py
```diff
@@ -0,0 +1,386 @@
...
+        # Log progress if necessary.
+        now = time.time()
+        if now - last_progress_log > 60:
+            log.info(_get_snapshot_progress_text(snapshots))
+            last_progress_log = now
+
+        sleep(5)
+
+
+def _get_snapshot_progress_text(snapshots):
```
bonzotronix: This is now public, so we should remove the leading underscore.

bburtin: I only see it called from `aws_util` and `test_aws_util`.  Is there a callsite that I'm missing?

brkt_cli/aws/aws_util.py
```diff
@@ -0,0 +1,386 @@
...
+
+class SnapshotError(BracketError):
+    pass
+
+
+class InstanceError(BracketError):
+    pass
+
+
+def wait_for_instance(
```
bonzotronix: I could easily imagine all of the aws_util functions residing in aws_service.py (especially since they do operate on AWSService objects).  Did we not want to move them to aws_service.py because that would make the file too bloated?  If so, would we want to move the global functions (like wait_for_volume) from aws_service.py to the new aws_util.py?

brkt_cli/aws/test_aws_util.py
```diff
@@ -0,0 +1,30 @@
...
+class TestSnapshotProgress(unittest.TestCase):
+
+    def test_snapshot_progress_text(self):
+        # One snapshot.
+        s1 = Snapshot()
+        s1.id = '1'
+        s1.progress = u'25%'
+        self.assertEqual(
+            '1: 25%',
+            aws_util._get_snapshot_progress_text([s1])
```
bonzotronix: Here.

brkt_cli/aws/test_aws_util.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+
+        # Two snapshots.
+        s2 = Snapshot()
+        s2.id = '2'
+        s2.progress = u'50%'
+
+        self.assertEqual(
+            '1: 25%, 2: 50%',
+            aws_util._get_snapshot_progress_text([s1, s2])
```
bonzotronix: and here

brkt_cli/aws/aws_util.py
bburtin: Yeah, that's reasonable.  I'll move the functions there instead of creating a new module.

brkt_cli/aws/test_aws_util.py
bburtin: I think it's reasonable to call a private function from unit test code.  We can expose the function later if we need to call it from application code outside `aws_util`/`aws_service`.

bonzotronix: OK, that's fine.

PR #642, opened 2017-04-06
Share-logs AWS Rewrite

markvoll:
This program creates a snapshot of the metavisors root disk,
attaches it to a volume, and launches a new instance
with the volume attached. The logs are then compressed
into a tar file and uploaded to a s3 bucket. This replaces
the existing "diag" command.

Usage: brkt aws share-logs --instance <instance_id>
--snapshot <snapshot_id> --region <region>
--bucket <bucket_name> --log-path <file_path>


brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,199 @@
...
+        # wait for instance to launch
+        logs_svc.wait_instance(new_instance)
+
+        # wait for file to upload
+        logs_svc.wait_bucket_file(bucket, path, region, s3)
+        log.info('Deleting new snapshot and instance')
+
+    finally:
+        try:
+            aws_svc.delete_snapshot(snapshot.id)
```
bburtin: Looks like `snapshot` may be `None` if an exception is thrown before it's initialized.  Also, I'd recommend using `AWSService.clean_up()`, to avoid the possibility of exceptions inside your `finally` block.

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,199 @@
...
+            aws_svc.terminate_instance(new_instance.id)
```
bburtin: You should also be able to pass the instance ID to `clean_up()`.

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,199 @@
...
+        :raises ValidationError if the name is invalid
+        """
+        m = re.match(r'[A-Za-z0-9()\!._/\-\'\*]+$', path)
+        if not m:
+            raise ValidationError(
+                "path may only contain letters, numbers, "
+                "and the following characters: !-_.*'()"
+            )
 
+    def wait_snapshot(self, snapshot):
```
bburtin: Does `aws_service.wait_for_snapshots()` do what you want?

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,199 @@
...
+        try:
+            aws_svc.delete_snapshot(snapshot.id)
+            aws_svc.terminate_instance(new_instance.id)
+        except Exception as e:
+            log.warn("Failed during cleanup: %s", e)
+
+
+class SharelogsService():
+
+    def wait_instance(self, instance):
```
bburtin: Does `aws_service.wait_for_instance()` do what you want?

brkt_cli/aws/share_logs.py
markvoll: Changed code to used clean_up method.

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,199 @@
...
+
+    finally:
+        try:
+            aws_svc.delete_snapshot(snapshot.id)
+            aws_svc.terminate_instance(new_instance.id)
+        except Exception as e:
+            log.warn("Failed during cleanup: %s", e)
+
+
+class SharelogsService():
```
markvoll: Fixed

brkt_cli/aws/share_logs.py
markvoll: Yes. Switched to this method.

brkt_cli/aws/share_logs.py
markvoll: Yes. Switched to this method.

brkt_cli/aws/test_share_logs.py
```diff
@@ -0,0 +1,96 @@
...
+# License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+from brkt_cli.aws import test_aws_service
+from brkt_cli.validation import ValidationError
+from brkt_cli.aws import share_logs
+
+
+class AWSService1(test_aws_service.DummyAWSService):
```
markvoll: Still looking into this

brkt_cli/aws/__init__.py
```diff
@@ -549,8 +487,8 @@ def run(self, values):
         if values.aws_subcommand == 'update':
             verbose = brkt_cli.is_verbose(values, self)
             return run_update(values, self.config, verbose=verbose)
-        if values.aws_subcommand == 'diag':
-            return run_diag(values)
+        # if values.aws_subcommand == 'diag':
```
bburtin: Cruft?

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,183 @@
...
+
+    log.info('Sharing logs')
+    snapshot = None
+
+    try:
+        s3 = logs_svc.s3_connect()
+        # Check bucket for file and bucket permissions
+        bucket_exists = logs_svc.check_bucket_file(bucket, path, region, s3)
+
+        if bucket_exists is False:
```
bburtin: `if not bucket_exists`

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,183 @@
...
+            log.info('Creating new bucket')
+            # If bucket isn't already owned create new one
+            new_bucket = logs_svc.make_bucket(bucket, region, s3)
+            # Reconnect with updated bucket list
+            s3 = logs_svc.s3_connect()
+            # Allow public write access to new bucket
+            new_bucket.set_acl('public-read-write')
+
+        if snapshot_id is None:
```
bburtin: `if not snapshot_id`

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,183 @@
...
+            # Wait for snapshot to post
+            log.info('Waiting for snapshot...')
+            aws_service.wait_for_snapshots(aws_svc, snapshot.id)
+
+        else:  # Taking logs from a snapshot
+            snapshot = aws_svc.get_snapshot(snapshot_id)
+
+        # Split path name into path and file
+        os.path.split(path)
+        file = os.path.basename(path)
```
bburtin: `file` is a reserved word.  Your code will work, but it's considered bad form to give your variables the same name as a reserved word.

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,183 @@
...
+
+        # wait for file to upload
+        log.info('Waiting for file to upload')
+        logs_svc.wait_bucket_file(bucket, path, region, s3)
+        log.info('Deleting new snapshot and instance')
+
+    finally:
+        try:
+            if snapshot:
+                aws_service.clean_up(aws_svc, instance_ids=[new_instance.id],
```
bburtin: Looks like it's possible for `new_instance` to be `None`.

brkt_cli/aws/share_logs.py
```diff
@@ -13,24 +13,183 @@
...
+        log.info('Waiting for instance...')
+        aws_service.wait_for_instance(aws_svc, new_instance.id)
+
+        # wait for file to upload
+        log.info('Waiting for file to upload')
+        logs_svc.wait_bucket_file(bucket, path, region, s3)
+        log.info('Deleting new snapshot and instance')
+
+    finally:
+        try:
```
bburtin: I don't think you need the `try/except`.  `clean_up()` should handle cleanup failures internally.

brkt_cli/aws/share_logs_args.py
```diff
@@ -29,23 +29,34 @@ def setup_share_logs_args(parser, parsed_config):
...
+        metavar='NAME',
+        dest='bucket',
+        help='Bucket that logs file is uploaded to',
+        required=True
+    )
+    parser.add_argument(
+        '--log-path',
+        metavar='PATH',
+        dest='path',
+        help='PATH in bucket to store logs file (e.g. dir/logs.tar.gz)',
```
bburtin: The `e.g.` may be unnecessary.  I believe that the default will automatically be printed in the usage output.

brkt_cli/aws/share_logs_args.py
```diff
@@ -29,23 +29,34 @@ def setup_share_logs_args(parser, parsed_config):
         dest='instance_id',
         help='The instance with Bracket system logs to be shared'
     )
+    parser.add_argument(
+        '--region',
+        metavar='NAME',
+        dest='region',
+        help='The region of the instance(e.g. us-west-2',
```
bburtin: Typo: ` (e.g. us-west-2)`

brkt_cli/aws/test_share_logs.py
```diff
@@ -0,0 +1,176 @@
...
+from brkt_cli.aws import share_logs
+import boto
+
+
+# This class is used for testing ShareLogs
+class AWSService1(share_logs.ShareLogsService):
+
+    created = False
+
+    class Snapshot():
```
bburtin: Can you use the boto classes here, instead of creating your own?  If you have to have them, it's better to declare them at the module level since they're used outside AWSService1.

brkt_cli/aws/test_share_logs.py
```diff
@@ -0,0 +1,176 @@
...
+# limitations under the License.
+
+import unittest
+from brkt_cli.validation import ValidationError
+from brkt_cli.aws import share_logs
+import boto
+
+
+# This class is used for testing ShareLogs
+class AWSService1(share_logs.ShareLogsService):
```
bburtin: Why is this called `AWSService1` if it's derived from `ShareLogsService`?  And do you really need inheritance here?  I would expect your dummy class would have the same signature as `ShareLogsService`, since it's a replacement and doesn't call out to AWS.

markvoll: Changed the name to make more sense, this class inherits from ShareLogsService because my test routine now iterates through the methods found in ShareLogsService instead of dummy methods. Are you suggesting making direct calls to ShareLogsService over using inheritance?

PR #643, opened 2017-04-06
Split the make-token command out of brkt_jwt

bburtin:
Split the code that handles the make-token subcommand out of the
brkt_jwt module.  This divides responsibility more clearly, and avoids
circular dependencies in JWT library callsites (e.g.
callsite->brkt_jwt->brkt_cli->brkt_jwt).

Also move the timestamp parsing functions into the util module, since
they're called from multiple places.

### DIFF
--- a/brkt_cli/make_token.py
+++ b/brkt_cli/make_token.py
@@ -1,0 +1,197 @@
+# Copyright 2017 Bracket Computing, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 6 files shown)


brkt_cli/make_token.py
```diff
@@ -0,0 +1,196 @@
...
+
+import argparse
+import json
+import logging
+import uuid
+
+import brkt_cli
+import brkt_cli.crypto
+from brkt_cli import brkt_jwt, ValidationError, config, util, argutil
+from brkt_cli.brkt_jwt import SUBCOMMAND_NAME
```
bonzotronix: Why not move SUBCOMMAND_NAME here?

bonzotronix:
FTP.  Basically LGTM.  Just one nit/question.

PR #647, opened 2017-04-07
Expose SSH Key pair for wrap command

krishnanrs:
For the `wrap-guest-image` command, exposed the SSH Key argument.
In the absence of this argument, users may be unable to connect
to the launched instance.

### DIFF
--- a/brkt_cli/aws/wrap_image_args.py
+++ b/brkt_cli/aws/wrap_image_args.py
@@ -41,7 +41,7 @@ def setup_wrap_image_args(parser, parsed_config):
     aws_args.add_security_group(parser)
     aws_args.add_subnet(parser)
     aws_args.add_aws_tag(parser)
-    aws_args.add_key(parser)
+    aws_args.add_key(parser, help='SSH key pair name')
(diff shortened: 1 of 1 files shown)


brkt_cli/aws/wrap_image_args.py
```diff
@@ -41,7 +41,7 @@ def setup_wrap_image_args(parser, parsed_config):
     aws_args.add_security_group(parser)
     aws_args.add_subnet(parser)
     aws_args.add_aws_tag(parser)
-    aws_args.add_key(parser)
+    aws_args.add_key(parser, help='SSH key pair name')
```
bburtin: Would it be better to change the function signature to `add_key(parser, exposed=False)`?  That way, the help text will be consistent if multiple commands want to expose the option.

krishnanrs: I'd rather prefer if instead there was a "required" flag to indicate whether the parameter is required or not. For now, I do not see this argument required to be exposed for any other command in brkt-cli and hence would stick with this simpler change. If we find a need down the line, then we can make this change.

PR #648, opened 2017-04-07
Remove sshpubkeys dependency

bburtin:
Remove the dependency on sshpubkeys, to simplify setup and avoid issues
with compiling binaries.  We were only using this library to
sanity-check the public key that is passed via the internal
--ssh-public-key option.

Replace the call to sshpubkeys with a simple string comparison that
checks the key content.

### DIFF
--- a/brkt_cli/crypto/__init__.py
+++ b/brkt_cli/crypto/__init__.py
@@ -110,6 +110,16 @@ def is_private_key(pem):
     return 'PRIVATE KEY' in pem
 
 
+def is_public_key(key):
+    """ Return True if the given key contents resemble an ssh public
+    key.  This a sanity check for validation of command line arguments, as
+    opposed to a rigorous test that parses the key.
+
+    :param key the public key contents as a string
+    """
+    return 'ssh-' in key or 'ecdsa-' in key
+
+
 def read_private_key(pem_path):
     """ Read a private key from a PEM file.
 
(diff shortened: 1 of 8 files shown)


brkt_cli/crypto/__init__.py
```diff
@@ -110,6 +112,24 @@ def is_private_key(pem):
     return 'PRIVATE KEY' in pem
 
 
+def is_public_key(key):
+    """ Return True if the given key contents resemble an ssh private
```
bobveznat: s/private/public/ ?

bburtin: Oops, fixed.

brkt_cli/crypto/test_crypto.py
```diff
@@ -211,3 +212,15 @@ def test_validate_cert(self):
         brkt_cli.crypto.validate_cert(TEST_CERT)
         with self.assertRaises(ValidationError):
             brkt_cli.crypto.validate_cert('foobar')
+
+
+class TestPublicKey(unittest.TestCase):
+
+    def test_is_public_key(self):
```
bobveznat: Do we pass this public key in via userdata to the instance? If we do we may as well support certificates. A cert authority line looks like this:

cert-authority ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAYQC6Shl5kUuTGqkSc8D2vP2kls2GoB/eGlgIb0BnM/zsIsbw5cWsPournZN2IwnwMhCFLT/56CzT9ZzVfn26hxn86KMpg76NcfP5Gnd66dsXHhiMXnBeS9r6KPQeqzVInwE=

bburtin: Done.

PR #653, opened 2017-04-12
Update default AWS instance types

bburtin:
Use modern AWS instance types that are supported in all regions.
Default the guest instance to m4.large and use c4.xlarge for the
encryptor/updater.


bonzotronix:
I would defer to krishnan or maybe jvoll for reviewing this.

bburtin:
Bob approved using m4.large and c4.xlarge.

PR #655, opened 2017-04-13
Optional IAM role for wrapped instance

krishnanrs:
Added an optional `--iam` argument which can be specified when
launching a wrapped instance. The specified IAM role is validated
before instance launch. This is useful when launching wrapped
instances to be used with Bracket Object store or other services
that depend on an instance IAM role.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -25,6 +25,7 @@ from datetime import datetime
 import boto
 import boto.sts
 import boto.vpc
+import boto.iam
 from boto.exception import EC2ResponseError, BotoServerError
 
 from brkt_cli import util, encryptor_service
@@ -102,6 +103,10 @@ class BaseAWSService(object):
 
     @abc.abstractmethod
     def get_volume(self, volume_id):
+        pass
+
+    @abc.abstractmethod
+    def iam_role_exists(self, role):
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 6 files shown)


krishnanrs:
My bad. I did not run the tests correctly in the first place. Added a test as part of fixing the failures as well.

brkt_cli/aws/aws_service.py
```diff
@@ -387,6 +392,14 @@ def get_volumes(self, tag_key=None, tag_value=None):
         get_all_volumes = self.retry(self.conn.get_all_volumes)
         return get_all_volumes(filters=filters)
 
+    def iam_role_exists(self, role):
+        conn = boto.iam.connect_to_region(self.region)
+        try:
+            conn.get_instance_profile(role)
+        except:
```
bburtin: Do we really want to silently fail here?

krishnanrs: The idea was to simply check if the IAM role exists or not so that we can do the error handling. Is there anything more you think we should do? The expectation is that this is an optional parameter and the user knows what IAM role they want. We just want to prevent the dumb typos.

bburtin: Can you do this by handling `EC2ResponseError` and checking the error code?  That way if something unexpected happens, we'll know about it.

krishnanrs: Done.

PR #656, opened 2017-04-13
Documentation Updates

krishnanrs:
Updated documentation to reflect the new CLI usage by using the
`--brkt-tag` argument with the various commands. The `--brkt-tag`
can be used once the user can run `brkt auth`, generated an API
token and set the `$BRKT_API_TOKEN` environment variable with the
corresponding output.

### DIFF
--- a/aws.md
+++ b/aws.md
@@ -54,7 +54,7 @@ Run **brkt aws encrypt** to create a new encrypted AMI based on an existing
 image:
 
 ```
-$ brkt aws encrypt --region us-east-1 --token <token> ami-76e27e1e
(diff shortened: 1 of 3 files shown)


aws.md
```diff
@@ -65,6 +65,11 @@ $ brkt aws encrypt --region us-east-1 --token <token> ami-76e27e1e
 ami-07c2a262
 ```
 
+**Note:** If you have already run the `brkt auth` command and set your
```
bburtin: I would omit the "if" part.  If auth isn't a requirement at this point, it will be soon.

PR #659, opened 2017-04-16
Consolidate common ESX command line arguments

krishnanrs:
Consolidated common ESX command line arguments across the various
ESX commands. Help texts were also modified to reflect the actual
usage/purpose of the argument.


apoorva-jindal:
LGTM. Thanks for fixing.

PR #661, opened 2017-04-19
Fix configuration of static IP for encryption

apoorva-jindal:
1. Configuration of static ip requires first power on
and then power off before it can be done.
2. Add console file after static ip configuration.
3. Set a static host name.

Updater does not work yet with static IP.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -581,7 +581,7 @@ class VCenterService(BaseVCenterService):
                                    memoryMB=(memoryGB*1024),
                                    numCPUs=numCPUs,
                                    files=vmx_file,
-                                   guestId='otherGuest64',
+                                   guestId='otherLinuxGuest64',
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -613,7 +614,7 @@ def configure_static_ip(self, vm, static_ip):
         adaptermap.adapter.dnsDomain = static_ip.dns_domain
         ident = vim.vm.customization.LinuxPrep(
             domain=static_ip.dns_domain,
-            hostName=vim.vm.customization.FixedName(name=vm.config.name))
```
krishnanrs: The configure_static IP is invoked by the update and wrap commands. Wouldn't that cause the instance to be always launched with the name as brkt-encryptor? Do we really want that?

apoorva-jindal:
Fixed the VM name already. I use session id now.

PR #662, opened 2017-04-20
Allow make-token times specified in days, months, years

krishnanrs:
Allow the expiry timestamp and not valid before timestamp to be
specified in user friendly manner of relative hour, day, week,
month or year.

### DIFF
--- a/brkt_cli/test_util.py
+++ b/brkt_cli/test_util.py
@@ -163,5 +163,13 @@ class TestTimestamp(unittest.TestCase):
         self.assertEqual(dt, util.parse_timestamp(str(ts)))
         self.assertEqual(dt, util.parse_timestamp(dt.isoformat()))
 
+        ts = int(time.time()) + 18000 # 5 hours from now
+        dt1 = datetime.fromtimestamp(ts, tz=iso8601.UTC)
+        dur = util.parse_timestamp('5h')
+        ts = int(time.time()) + 18000 # 5 hours from now
+        dt2 = datetime.fromtimestamp(ts, tz=iso8601.UTC)
+        self.assertLessEqual(dt1, dur)
+        self.assertGreaterEqual(dt2, dur)
+
         with self.assertRaises(ValidationError):
             util.parse_timestamp('abc')
--- a/brkt_cli/util.py
+++ b/brkt_cli/util.py
@@ -18,7 +18,7 @@ import logging
 import re
 import time
 import uuid
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import iso8601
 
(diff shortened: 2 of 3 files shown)


krishnanrs:
Let me know if this looks OK or if the help messages need updates. I also added a check to ensure that the not-valid before timestamp is not later than the expiry time.

brkt_cli/util.py
```diff
@@ -310,6 +310,28 @@ def parse_timestamp(ts_string):
     """
     now = int(time.time())
 
+    # Parse relative timestamp
+    m = re.match('\d+(m|h|d|w|M|Y)$', ts_string)
```
bburtin: I would keep it simple, and only support minutes, hours, and days.  We can always add other time intervals later, if we really need them.  Also, can you break this out into a duration parsing function in `util` and add a unit test?

brkt_cli/util.py
```diff
@@ -326,8 +348,9 @@ def parse_timestamp(ts_string):
         dt = iso8601.parse_date(ts_string)
     except iso8601.ParseError:
         raise ValidationError(
-            'Timestamp "%s" must either be a Unix timestamp or in iso8601 '
-            'format (2016-05-10T19:15:36Z).' % ts_string
+            'Timestamp "%s" must either be a Unix timestamp, in iso8601 '
+            'format (2016-05-10T19:15:36Z) or range offset '
```
bburtin: "or duration"

brkt_cli/util.py
bburtin: Minor correction: probably no need to move it to `util` until we have a second callsite.

krishnanrs: By default we support tokens that have no expiry. The feedback we have received is for expiry to be in multiple years so that user does not have to re-generate tokens often. For that reason, I would prefer to keep the month and year option.

brkt_cli/util.py
```diff
@@ -326,8 +348,9 @@ def parse_timestamp(ts_string):
         dt = iso8601.parse_date(ts_string)
     except iso8601.ParseError:
         raise ValidationError(
-            'Timestamp "%s" must either be a Unix timestamp or in iso8601 '
-            'format (2016-05-10T19:15:36Z).' % ts_string
+            'Timestamp "%s" must either be a Unix timestamp, in iso8601 '
+            'format (2016-05-10T19:15:36Z) or duration '
+            '(8760h, 365d, 52w, 12M, 1Y etc.).' % ts_string
```
bburtin: Since we decided to support minutes, hours, days, and years, let's use more realistic numbers in the example.  Maybe "(e.g. 30m, 6h, 5d, 2y)"?

brkt_cli/util.py
bburtin: Ok, makes sense.  I think we can handle all cases with m, h, d, and y.  I don't think supporting weeks adds much value.  Months add complexity, since calendar months have different lengths and we end up with different durations for "m" and "M".

krishnanrs: Done. Please take a look again.

krishnanrs:
Hopefully testing for one range is sufficient. If you want me to add tests for all supported ranges, let me know.

brkt_cli/test_util.py
```diff
@@ -163,5 +163,9 @@ def test_parse_timestamp(self):
         self.assertEqual(dt, util.parse_timestamp(str(ts)))
         self.assertEqual(dt, util.parse_timestamp(dt.isoformat()))
 
+        ts = int(time.time()) + 18000 # 5 hours from now
+        dt = datetime.fromtimestamp(ts, tz=iso8601.UTC)
+        self.assertEquals(dt, util.parse_timestamp('5h'))
```
bburtin: This looks like a race condition.  I think you'll need to check the the timestamp is within 5 seconds of the expected value.

krishnanrs: Check again.

PR #663, opened 2017-04-20
Enable sriovNetSupport for AWS wrapped instances

krishnanrs:
If the AWS wrapped instance does not have sriovNetSupport enabled,
then stop the instance (which is required) and try to enable the
same. Report an error if sriovNetSupport cannot be enabled, but
leave the instance running as the only drawback is that the
network performance might not be optimal.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -94,6 +94,10 @@ class BaseAWSService(object):
         pass
 
     @abc.abstractmethod
+    def start_instance(self, instance_id):
+        pass
+
+    @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


krishnanrs:
@jimvoll can you review this along with Andy? I am not convinced that this is the right change. I'd rather prefer that the MV AMI has sriovNetSupport enabled. That way when you launch the wrapped instance, either using the CLI or directly from the AWS console (or other orchestration tools), user does not have to manually enable sriovNetSupport for the wrapped instance.

krishnanrs:
Closing this pull request as Andy agreed that the best way forward is to enable sriovNetSupport on the base AMIs themselves.

PR #665, opened 2017-04-21
Add `brkt-tag` argument to `make-user-data` command

krishnanrs:
This makes brkt-cli on par with the UI so that users can use either
the CLI or the UI to generate the userdata when launching instances.
The CLI command is especially more powerful as it allows users to
insert their own cloud-config and scripts, which is currently not
possible via the UI.

### DIFF
--- a/brkt_cli/make_user_data/__init__.py
+++ b/brkt_cli/make_user_data/__init__.py
@@ -24,7 +24,8 @@ from brkt_cli.instance_config import GuestFile
 from brkt_cli.instance_config import INSTANCE_METAVISOR_MODE
 from brkt_cli.instance_config_args import (
     instance_config_from_values,
-    setup_instance_config_args
+    setup_instance_config_args,
+    get_launch_token
 )
 from brkt_cli.subcommand import Subcommand
 from brkt_cli.validation import ValidationError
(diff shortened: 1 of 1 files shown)


krishnanrs:
Can someone clarify which this option was removed as part of https://github.com/brkt/brkt-cli/pull/608? If there is a very good reason to not add it back, let me know.

bonzotronix:
I'm afraid I can't remember what led us to decide that the `brkt-tag` should not be in `make-user-data`.

brkt_cli/make_user_data/__init__.py
```diff
@@ -76,11 +77,15 @@ def make(values, config):
             raise ValidationError(
                 '--brkt-env is only allowed with --unencrypted-guest')
 
+    lt = values.token
+    if not lt and values.brkt_tags:
```
bburtin: Should we raise `ValidationError` if the user specifies both `--token` and `--brkt-tag`?  Seems like `--brkt-tag` would be ignored in this case.

bonzotronix: Yeah, that makes sense.  Sorry I didn't catch that earlier.

bburtin: Oh, those options are already in a mutually exclusive group.  Oops.  :-)

PR #666, opened 2017-04-25
Add metavisor-version argument

dasein:
Add support for determining metavisor version through published
version artifacts in AWS.

- Add mv_version module to brkt_cli

- Update esx provider to fetch OVF objects from S3 using metavisor
  version

- Update esx provider to fetch amis.json using metavisor version

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -129,7 +129,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
             (ovf, file_list) = \
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
-                    image_name=values.image_name,
+                    version=values.metavisor_version,
                     proxy=proxy
                 )
             if ovf is None:
@@ -327,7 +327,7 @@ def run_update(values, parsed_config, log, use_esx=False):
             (ovf_name, download_file_list) = \
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
-                    image_name=values.image_name,
+                    version=values.metavisor_version,
                     proxy=proxy
                 )
             if ovf_name is None:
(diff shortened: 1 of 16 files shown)


dasein:
Tests failed because we cannot reach packages.int.brkt.net from outside. Trying to think of a better workaround to make sure internal dev still works.

brkt_cli/esx/esx_args.py
```diff
@@ -282,6 +282,17 @@ def add_metavisor_ovf_image_name(parser):
     )
 
 
+def add_metavisor_version(parser):
+    parser.add_argument(
+        '--metavisor-version',
+        metavar='NAME',
+        dest='metavisor_version',
+        help='Metavisor version [e.g 1.2.12 ] (default: latest)',
```
krishnanrs: Help says default is 'latest' but it is being defaulted to None

brkt_cli/esx/__init__.py
```diff
@@ -129,7 +129,7 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
             (ovf, file_list) = \
                 esx_service.download_ovf_from_s3(
                     values.bucket_name,
-                    image_name=values.image_name,
+                    version=values.metavisor_version,
```
krishnanrs: Since we have multiple optional arguments (image_name, mv_vesion), we need to:
1. Decide which takes precedence
2. Maybe raise a warning when multiple arguments are use
3. Infer one value from the other. For e.g. if the mv_version is going to be the standard going forward, infer the mv_version from the image_name and then pass down the mv_version to all downstream methods

Basically we should continue supporting the old way of being able to specify a specific MV image (especially required for dev, which I believe we are, looking at this change). Not to mention, this will require doc updates.

dasein: The way this has worked has not changed. If you look farther up in the code, if you specify an image name, it will use that for local encryption. This is just for cases where you are fetching the default metavisor version. The same applies for aws encryption actions...if you do not specify an ami, then we go and fetch amis.json to figure out what default ami is required. The --metavisor-version argument just gives you more control on what version you will fetch by default.

brkt_cli/esx/esx_args.py
dasein: If you do not specify a version (e.g. None) then we figure out the latest published version.

krishnanrs:
lgtm. I will let @apoorva-jindal approve this, pending the tests passing.

PR #667, opened 2017-04-25
Make esx keepalive threads be daemons

atonderski:
We were seeing brkt-cli timeouts after a successful encryption.
Logs were indicating that keepalive-upload thread was preventing exit:
Thread name: MainThread Daemon: 0 IsAlive 1
Thread name: keepalive-upload Daemon: 0 IsAlive 1

Testing: Ran esx encryption successfully


apoorva-jindal:
LGTM

PR #670, opened 2017-05-01
Fix argument name for `wrap-with-esx-host` command

krishnanrs:
All ESX specific VMWare subcommands should refer to `esx-datastore`
and not `esx-datacenter`. A `datacenter` argument only applies for
vCenter specific commands.


apoorva-jindal:
LGTM

PR #671, opened 2017-05-01
Fix static ip configuration for wrap script

apoorva-jindal:
1. Fix static IP configuration for wrap script.
2. Fix VM reconfiguration for wrap script.

### DIFF
--- a/brkt_cli/esx/wrap_image.py
+++ b/brkt_cli/esx/wrap_image.py
@@ -50,7 +50,10 @@ def wrap_from_s3(vc_swc, guest_vmdk, vm_name=None,
         vc_swc.add_disk(vm, filename=guest_vmdk_path, unit_number=1)
         if user_data_str:
             vc_swc.send_userdata(vm, user_data_str)
+        vc_swc.reconfigure_vm_cpu_ram(vm)
         if static_ip:
+            vc_swc.power_on(vm)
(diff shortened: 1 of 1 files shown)


brkt_cli/esx/wrap_image.py
```diff
@@ -50,7 +50,10 @@ def wrap_from_s3(vc_swc, guest_vmdk, vm_name=None,
         vc_swc.add_disk(vm, filename=guest_vmdk_path, unit_number=1)
         if user_data_str:
             vc_swc.send_userdata(vm, user_data_str)
+        vc_swc.reconfigure_vm_cpu_ram(vm)
         if static_ip:
+            vc_swc.power_on(vm)
```
bburtin: Will it be obvious to the next person why you need to power the VM on and off?

apoorva-jindal: Hopefully to the folks using static ip. Things seem to work only after power on and power off in vcenter. No idea why though.

krishnanrs: There is a separate bug filed to ensure that a power off/on sequence to be not required. Until that is fixed, we will have to release note this.

bburtin: Since the behavior is not obvious, it may be worth adding a comment that explains why you're powering the VM on and off.

PR #672, opened 2017-05-02
Try to use GCP private IPs

krishnanrs:
Instances by default have an ephemeral public IP. However in cases
where we have a site-to-site VPN or the ephemeral IP is explicitly
disabled due to policy, the brkt-cli should try to use the
private IP. With this change, brkt-cli checks the encryptor status
on both the public and private IPs during encryption and updates,
much similar to AWS.

### DIFF
--- a/brkt_cli/gcp/gcp_service.py
+++ b/brkt_cli/gcp/gcp_service.py
@@ -113,6 +113,10 @@ class BaseGCPService(object):
         pass
 
     @abc.abstractmethod
+    def get_private_ip(self, name, zone):
+        pass
+
+    @abc.abstractmethod
     def detach_disk(self, zone, instance, diskName):
         pass
 
@@ -406,6 +410,15 @@ class GCPService(BaseGCPService):
             except:
                 pass
         self.log.info("Couldn't find an IP address for this instance.")
+
+    def get_private_ip(self, name, zone):
+        nw = 'networkInterfaces'
+        instance_req = self.compute.instances().get(project=self.project,
+                zone=zone, instance=name)
+        instance = retry(execute_gcp_api_call)(instance_req)
+        if instance[nw][0]['networkIP']:
+            return instance[nw][0]['networkIP']
+        self.log.info("Couldn't find private IP address for this instance.")
 
     def write_serial_console_file(self, zone, instance):
         try:
--- a/brkt_cli/gcp/encrypt_gcp_image.py
+++ b/brkt_cli/gcp/encrypt_gcp_image.py
@@ -84,8 +84,14 @@ def do_encryption(gcp_svc,
                          tags=gcp_tags)
 
     try:
+        host_ips = []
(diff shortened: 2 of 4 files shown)


brkt_cli/gcp/gcp_service.py
```diff
@@ -407,6 +411,20 @@ def get_instance_ip(self, name, zone):
...
+        nw = 'networkInterfaces'
+        for i in range(60):
+            time.sleep(5)
+            instance_req = self.compute.instances().get(project=self.project,
+                    zone=zone, instance=name)
+            try:
+                instance = retry(execute_gcp_api_call)(instance_req)
+                if instance[nw][0]['networkIP']:
+                    return instance[nw][0]['networkIP']
+            except:
```
bburtin: Can this catch a typed exception?  If something unexpected happens, we'll end up with silent failures.

jimvoll: Seems like we won't silently ignore failures. The next line has:
        self.log.info("Couldn't find private IP address for this instance.")

This is the same logic as  get_instance_ip() FWIW

bburtin: The `except` catches all exception types, not just the expected type.  If an unexpected exception is thrown, we won't know what it was.

jimvoll: I understand how exceptions work and don't work in Python.

If you want to print the unexpected exceptions are that are sure to arise you can do
this:
except Exception as e:
    blah blah blah

It doesn't really matter what the exception is, the user is not going to get his/her IP
address and that's all that matters. Printing a stack trace is only helpful for developers,
not the end user. If you can make some meaningful distinction between various
exceptions being caught (different error handling) then it makes sense. Otherwise
it is just fooling ourselves, and torturing our customers with error messages in the form
of stack-traces.

And I'm against that.

bburtin: And I'm against customers waiting for 5 minutes and not knowing why the operation failed.  Pick your poison.

jimvoll: Well I'm against that too! but my reading of the code results in printing
an error message: "Couldn't find private IP address for this instance"
and then failing.

Which is about the best I could hope for. I'm just arguing that checking for a specific
exception or unexpected exception doesn't matter. Print the message and fail.
Which is how get_instance_ip() works. I hope.

jimvoll: Oh, I see my flawed logic. We will retry 60 times, and if it's an expected error
that might make sense, but if it's unexpected, we should just fail right away.

But in either case, we'll print the error message that we couldn't get an IP address.

But failing sooner for unexpected errors makes sense. I just hate the stack trace-back
message associated with them as a user.

krishnanrs:
Can someone look at this again?

jimvoll:
Follow up with Boris, I was ok with the initial code.

krishnanrs:
Can you check again?

PR #673, opened 2017-05-03
Cleanup ESX code

apoorva-jindal:
1. Since power-on and power-off for MV VMs is always
needed to configure static IP, move from the caller to
inside the configuration of static IP function.
Also add a comment to explain.
2. Remove the temporary hack to debug hung threads.
3. During vcenter valdation, remove the redundant
input use_esx.

### DIFF
--- a/brkt_cli/esx/update_vmdk.py
+++ b/brkt_cli/esx/update_vmdk.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import logging
 import os
-import threading
 from brkt_cli.encryptor_service import (
     wait_for_encryptor_up,
     wait_for_encryption,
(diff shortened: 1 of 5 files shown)


brkt_cli/esx/update_vmdk.py
```diff
@@ -116,12 +115,6 @@ def update_ovf_image_mv_vm(vc_swc, enc_svc_cls, guest_vm, mv_vm,
         vc_swc.destroy_vm(guest_vm)
         vc_swc.destroy_vm(mv_vm)
     log.info("Done")
```
krishnanrs: Don't you want to remove a similar temporary hack in encrypt_vmdk?

apoorva-jindal: Yes, I do. Thanks for pointing out. Will do it now.

PR #678, opened 2017-05-17
Fix --metavisor-version to return exact version for OVF images

dasein:
We were not anchoring our bucket filter prefix for OVF images, which
caused us to return unofficial versions (e.g. -debug).

Also tweaked the error handling for users that are specifying both
--metavisor-version and --metavisor-ovf-image-name. The latter now
only refers to local .ovf files.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -97,8 +97,8 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
         if use_esx is False and values.template_vm_name is None:
             raise ValidationError("Missing template-vm-name for the "
                                   "template VM")
-    if (values.source_image_path is not None and values.image_name is None):
-        raise ValidationError("Specify the Metavisor OVF file.")
+    if values.source_image_path is None and values.image_name is not None:
+        raise ValidationError("Specify OVF image location with --ovf-source-directory")
     if use_esx:
         _check_env_vars_set('ESX_USER_NAME')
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -97,8 +97,8 @@ def run_encrypt(values, parsed_config, log, use_esx=False):
         if use_esx is False and values.template_vm_name is None:
             raise ValidationError("Missing template-vm-name for the "
                                   "template VM")
-    if (values.source_image_path is not None and values.image_name is None):
-        raise ValidationError("Specify the Metavisor OVF file.")
+    if (values.source_image_path is None and values.image_name is not None):
```
bburtin: You probably want `if values.source_image_path and values.image_name:`

dasein: Bit confused. The expected behavior is that if someone specifies --metavisor-ovf-image-name, they need to specify the --ovf-source-directory.

krishnanrs: I guess, what @bburtin meant was `if (not values.source_image_path and values.image_name)`

dasein: style!! got it :)

bburtin: Yeah, I'm better with style than with boolean logic.

apoorva-jindal:
LGTM after fixing the style.

PR #679, opened 2017-05-22
Clone the guest disk before encryption and then delete it

apoorva-jindal:
This allow multiple encryption jobs to be run in parallel
on the same guest vmdk.

### DIFF
--- a/brkt_cli/esx/encrypt_vmdk.py
+++ b/brkt_cli/esx/encrypt_vmdk.py
@@ -50,13 +50,17 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
                                 user_data_str=None, serial_port_file_name=None,
                                 status_port=ENCRYPTOR_STATUS_PORT, static_ip=None):
     try:
+        # clone the guest vmdk
+        new_guest_vmdk_name = vc_swc.get_datastore_path(vc_swc.session_id +
+                                                        guest_vmdk)
+        vc_swc.clone_disk(source_disk_name=vc_swc.get_datastore_path(guest_vmdk),
+                          dest_disk_name=new_guest_vmdk_name)
         mv_vm_name = vc_swc.get_vm_name(vm)
         # Reconfigure VM with more CPUs and memory
         vc_swc.reconfigure_vm_cpu_ram(vm)
         # Add datastore path to the guest vmdk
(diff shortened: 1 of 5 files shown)


brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -50,13 +50,17 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
                                 user_data_str=None, serial_port_file_name=None,
                                 status_port=ENCRYPTOR_STATUS_PORT, static_ip=None):
     try:
+        # clone the guest vmdk
+        new_guest_vmdk_name = vc_swc.get_datastore_path(vc_swc.session_id +
+                                                        guest_vmdk)
+        vc_swc.clone_disk(source_disk_name=vc_swc.get_datastore_path(guest_vmdk),
```
krishnanrs: Can't this same method be used for the `wrap` command as well? Can you fix this there as well and get this done in one shot?

brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -50,13 +50,17 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
                                 user_data_str=None, serial_port_file_name=None,
                                 status_port=ENCRYPTOR_STATUS_PORT, static_ip=None):
     try:
+        # clone the guest vmdk
+        new_guest_vmdk_name = vc_swc.get_datastore_path(vc_swc.session_id +
```
krishnanrs: Is there a length limitation on the name (since we are concatenating the string)?

brkt_cli/esx/encrypt_vmdk.py
apoorva-jindal: Fixed. Please review if I missed anything.

brkt_cli/esx/encrypt_vmdk.py
apoorva-jindal: I am not sure what the length limitation is. There a length limitation on the display name, but it does not get rejected even if I use a very long name.

apoorva-jindal: I just checked with length > 64 bytes and it went through. We can fix if we run into an issue here.

PR #681, opened 2017-05-24
Handle error in find-vm

apoorva-jindal:
1. get_obj sometimes gets exceptions when someone in
parallel is deleting other vms. Handle such exceptions.
2. Get vm in encrypt code should not fail. Retry on failure
10 times before giving up.

### DIFF
--- a/brkt_cli/esx/encrypt_vmdk.py
+++ b/brkt_cli/esx/encrypt_vmdk.py
@@ -95,7 +95,12 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
             log.error("Failed to re-connect to vCenter after encryption. "
                       "Please cleanup VM %s manually.", mv_vm_name)
             raise
-        vm = vc_swc.find_vm(mv_vm_name)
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/encrypt_vmdk.py
```diff
@@ -95,7 +95,13 @@ def create_ovf_image_from_mv_vm(vc_swc, enc_svc_cls, vm, guest_vmdk,
             log.error("Failed to re-connect to vCenter after encryption. "
                       "Please cleanup VM %s manually.", mv_vm_name)
             raise
-        vm = vc_swc.find_vm(mv_vm_name)
+        vm = None
+        retry = 0
+        while not vm and retry < 10:
```
krishnanrs: while not vm and retry in range(10):
This way you don't have to maintain another variable. Also, though there be a sleep between the retries? If so, the retry itself could be invoked as a separate method

apoorva-jindal: Range did not work with while. No sleep between retries. VM should always be available, just adding a retry loop as a defensive measure.

PR #684, opened 2017-05-26
Add optional CDROM configuration to encryptor VM

krishnanrs:
Provided an optional (hidden) CDROM flag which can be used to
attach a CDROM device to the encryptor instance which is required
for PCF integration.

### DIFF
--- a/brkt_cli/esx/esx_service.py
+++ b/brkt_cli/esx/esx_service.py
@@ -136,7 +136,7 @@ class BaseVCenterService(object):
     def __init__(self, host, user, password, port,
                  datacenter_name, datastore_name, esx_host,
                  cluster_name, no_of_cpus, memoryGB, session_id,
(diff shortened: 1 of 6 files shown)


brkt_cli/esx/esx_service.py
```diff
@@ -708,6 +709,35 @@ def delete_serial_port_to_file(self, vm, filename):
         log.info("Console message will no longer be dumped to file %s",
                  filename)
 
+    def add_cdrom(self, vm=None):
+        self.validate_connection()
+        # Find the IDE controller
+        ide_ctlr = None
+        if vm:
+            for dev in vm.config.hardware.device:
+                if isinstance(dev, vim.vm.device.VirtualIDEController):
```
apoorva-jindal: We do always have an ide controller in the MV OVF we build?

apoorva-jindal:
LGTM

PR #686, opened 2017-05-26
Log Metavisor version fetch message at debug level

bburtin:
Drop this message down from info to debug.  It's an implementation
detail, and isn't interesting in the normal case.


dasein:
LGTM

PR #688, opened 2017-05-31
AWS Share Logs Update

markvoll:
Fixed large file uploading error. AWS defaults to multipart uploading
for large files which caused uploading errors. The threshold has been
increased to 256MB. Added '--subnet' option to allow for older m4.large
instances to be compatable.


bburtin:
The code change looks fine.  Can you update the commit message and explain the connection between subnet and older instance types?

PR #691, opened 2017-06-12
Allow --service-domain, --ca-cert in make-user-data

bonzotronix:
* To allow users to override proxy settings (and the like), allow the --service-domain and --brkt-env args to make-user-data.


bonzotronix:
Could either Krishnan or Boris take another look at this? (given the change in behavior of always including the Yeti endpoints in the brkt-config)

krishnanrs:
Actually even @jimvoll mentioned the same thing. It should take the environment setting in the {{brkt config}} settings. User can override it using the `--service-domain` flag. I will let others also to chime in with their thoughts.

bburtin:
Yes, good point. It would make sense for `make-user-data` to get the Yeti endpoints from the current config environment, like the other commands do.

bonzotronix:
It does the right thing currently. It gets the Yeti endpoints from the config, unless they are overridden with a `--service-domain` or `--brkt-env` argument. This behavior is exercised in the unit tests (the "foo.com" endpoints come from the config; a couple unit tests override with `--service-domain` of "stage.mgmt.brkt.com").

PR #692, opened 2017-06-13
Make encryption fun again!

atonderski:
Add the BRKT Entertainment System which lets you play games while encrypting.
Everything is ascii based and in the terminal. The last couple of log lines from the actual cli command are displayed in the game windows so you can see what's going on and when the encryption is done.

Only some commands have the --fun flag enabled. These commands are chosen by adding the @gamify decorator to their target functions.

When you choose to play a game the cli creates 2 new processes: one for the command to execute and one for the game. The commands output is piped to a file so that it can be scraped by the game and also restored after the game is done. If you exit the game before the cli command is done, it will just go back to displaying the command output as it comes along.

Added dependencies:
asciimatics (1.8.0)

Testing: Ran a bunch of different commands, both successful and unsuccessful to make sure that
the game wouldn't mess with the commands themselves


brkt_cli/game/gamify_decorator.py
```diff
@@ -0,0 +1,76 @@
...
+            return func(*args, **kwargs)
+
+        try:
+            brkt_env = brkt_env_from_values(args[0], args[1])
+            game.yeti_env = 'https://api.%s' % (
+                '.'.join(brkt_env.public_api_host.split(
+                        '.')[1:]))
+            print game.yeti_env
+            time.sleep(1)
+            token = args[0].token
```
bburtin: `--token` is for launch tokens.  Just check the environment variable to get the API token.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+asciimatics==1.8.0
```
bburtin: Let's remove this, so that people don't complain about having to install 6 extra requirements.  When the user specifies `--fun`, you can dynamically check for the asciimatics library, and tell the user to run `pip install` if it's not there.

brkt_cli/game/gamify_decorator.py
```diff
@@ -4,10 +4,51 @@
 import sys
 import time
 
-import brkt_cli.game as game
-import game_controller
 from brkt_cli import brkt_env_from_values
 
+has_asciimatics = True
+try:
+    import brkt_cli.game as game
```
bburtin: Shouldn't you be importing `asciimatics` here? If importing `brkt_cli.game` fails for another reason, the user will see a misleading error message.

brkt_cli/game/gamify_decorator.py
```diff
@@ -17,17 +58,17 @@ def func_wrapper(*args, **kwargs):
...
-            print game.yeti_env
-            time.sleep(1)
-            token = args[0].token
-            if not token:
-                token = os.getenv('BRKT_API_TOKEN')
-            game.token = token
+                '.'.join(brkt_env.public_api_host.split('.')[1:]))
+            game.token = os.getenv('BRKT_API_TOKEN', None)
         except Exception as e:
             logging.error("You can play but you can't post to yeti :(. "
```
bburtin: End users don't know what Yeti is.  I would say "high scores aren't available."

PR #693, opened 2017-06-19
Fix incorrect thumbprint generation for public keys

aditya-brkt:
Sometimes the EC keypairs generated are a byte shorter.
Either x or y could be 47 bytes instead of 48 which is
mathematically valid. But when computing thumbprint,
these need to be padded with leading zeros.
Added this change.
Generated such a keypair and uploaded it to Yeti.
Made sure the KeyId computed by yeti matches the one
computed by brkt-cli

Jira: YETI-2778 #resolve

### DIFF
--- a/brkt_cli/brkt_jwt/jwk.py
+++ b/brkt_cli/brkt_jwt/jwk.py
@@ -18,12 +18,21 @@ import logging
 
 log = logging.getLogger(__name__)
 
+EC384_KEY_LEN_BYTES = 48
 
-def _long_to_byte_array(long_int):
+# If the length of the bytearray created is less than 48, pad it with leading
+# zeros. EC384 are 384bits i.e. 48byte keys. so both x and y need to be 48bytes
+# with leading zeros wherever required
+
+
+def _long_to_byte_array(long_int, pad_to_len=EC384_KEY_LEN_BYTES):
     bys = bytearray()
     while long_int:
         long_int, r = divmod(long_int, 256)
         bys.insert(0, r)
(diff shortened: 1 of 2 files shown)


brkt_cli/brkt_jwt/jwk.py
```diff
@@ -18,12 +18,24 @@
...
 
+EC384_KEY_LEN_BYTES = 48
 
-def _long_to_byte_array(long_int):
+# If the length of the bytearray created is less than 48, pad it with leading
+# zeros. EC384 are 384bits i.e. 48byte keys. so both x and y need to be 48bytes
+# with leading zeros wherever required
+
+
+def _long_to_byte_array(long_int, expected_len=EC384_KEY_LEN_BYTES):
```
bonzotronix: Sorry to nit-pick, but might "pad_to_len" be a preferable arg name?

brkt_cli/brkt_jwt/test_jwt.py
```diff
@@ -111,7 +111,7 @@ class TestJWK(unittest.TestCase):
 
     def test_long_to_byte_array(self):
         l = long('deadbeef', 16)
-        ba = brkt_jwt.jwk._long_to_byte_array(l)
+        ba = brkt_jwt.jwk._long_to_byte_array(l, expected_len=4)
         self.assertEqual(bytearray.fromhex('deadbeef'), ba)
```
bonzotronix: Might as well add a test of the fancy new padding.  :)

PR #702, opened 2017-07-05
Simplify the update_ami code

bburtin:
Simplify the update_ami code and remove a bunch of (hopefully)
unnecessary logic.

Hardcode the Metavisor and guest root device names, instead of reading
them from the block device mappings.  We don't expect these to change
any time soon.

Consolidate the hacks that preserve volume settings, such as volume type
and IOPS, so that they don't obfuscate the rest of the code.

### DIFF
--- a/brkt_cli/aws/update_ami.py
+++ b/brkt_cli/aws/update_ami.py
@@ -25,29 +25,33 @@ import json
 import logging
 import os
 
-from brkt_cli import encryptor_service
-from brkt_cli.aws.aws_service import (
-    wait_for_instance, stop_and_wait,
-    wait_for_image, create_encryptor_security_group, clean_up,
-    log_exception_console, wait_for_volume_attached
-)
+from brkt_cli.util import Deadline
+
 from brkt_cli.encryptor_service import (
     wait_for_encryptor_up,
-    wait_for_encryption,
+    wait_for_encryption
 )
+
+from brkt_cli import encryptor_service
+from brkt_cli.aws.aws_constants import (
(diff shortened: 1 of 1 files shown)


bburtin:
I'm testing with Ubuntu and RHEL.  Let me know if there are any other AMIs that I should test.

brkt_cli/aws/update_ami.py
```diff
@@ -146,23 +142,6 @@ def update_ami(aws_svc, encrypted_ami, updater_ami, encrypted_ami_name,
             else:
                 os.environ['NO_PROXY'] = updater.private_ip_address
 
-        enc_svc = enc_svc_class(host_ips, port=status_port)
-        log.info('Waiting for updater service on %s (port %s on %s)',
-                 updater.id, enc_svc.port, ', '.join(host_ips))
-        try:
-            wait_for_encryptor_up(enc_svc, Deadline(600))
```
apoorva-jindal: Are you sure this always works? Metavisor in updater mode writes information to the disk like any new JWT provided etc, and when everything is done, it declares encryption complete. If you stop the instance immediately after, not sure how you will ensure that all processing MV needs to do in updater mode is complete?

bburtin: Good call.  I'm in the process of putting this code back.  It worked ok when I tested on Ubuntu, but I hit a race condition with RHEL.

bburtin:
After reinstating the code that waited for the encryption service, I was able to successfully encrypt and update Ubuntu, RHEL, CentOS, and Amazon Linux images.  Can someone approve if the code looks reasonable?

PR #703, opened 2017-07-06
Don't log reason and code when waiting for image

bburtin:
Don't log the reason and code when waiting for an image.  In my testing,
these always come back as None.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -699,19 +699,14 @@ def wait_for_image(aws_svc, image_id):
                 continue
             else:
                 log.warn('Unknown AWS error: %s', e)
-        # These two attributes are optional in the response and only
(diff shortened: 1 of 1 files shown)


brkt_cli/aws/aws_service.py
```diff
@@ -699,19 +699,14 @@ def wait_for_image(aws_svc, image_id):
                 continue
             else:
                 log.warn('Unknown AWS error: %s', e)
-        # These two attributes are optional in the response and only
-        # show up sometimes. So we have to getattr them.
-        reason = repr(getattr(image, 'stateReason', None))
-        code = repr(getattr(image, 'code', None))
```
bonzotronix: Looking at the boto doc, I don't see any evidence that boto.ec2.connection.get_image() returns anything other than the image (on success) or None on failure.  So yeah, not sure why we even look for a reason or code.

PR #704, opened 2017-07-11
Paginate file list in GCP buckets

krishnanrs:
The GCP bucket list command returns the first 1000 items by default.
In the case where a bucket contains more than 1000 items, we do not
paginate thru' the entire list of files and as a result only
operate on the first 1000 entries. This especially causes problems
when you specify an encryptor bucket with more than 1000 files as
it incorrectly picks an older Metavisor image.

### DIFF
--- a/brkt_cli/gcp/gcp_service.py
+++ b/brkt_cli/gcp/gcp_service.py
@@ -591,7 +591,13 @@ class GCPService(BaseGCPService):
             return old
 
     def get_image_file(self, bucket):
-        files = self.storage.objects().list(bucket=bucket).execute()['items']
+        files = []
+        request = self.storage.objects().list(bucket=bucket)
+        # Paginate thru' the list of files in the bucket
+        while request:
+            resp = request.execute()
+            files += resp['items']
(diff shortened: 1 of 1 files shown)


brkt_cli/gcp/gcp_service.py
```diff
@@ -591,7 +591,13 @@ def get_younger(self, new, old):
             return old
 
     def get_image_file(self, bucket):
-        files = self.storage.objects().list(bucket=bucket).execute()['items']
+        files = []
+        request = self.storage.objects().list(bucket=bucket)
```
bburtin: Can this be replaced with a search for the `name` field?  That way, we'd avoid paginating through thousands of results.

krishnanrs: This is an issue only when using the internal `metavisor` bucket. Customers use the prod bucket which currently has only 25 items. Hence this will not be an issue in all customer use case. Further, with the `metavisor` bucket, we do not publish the `latest.image.tar.gz` file and hence we would need the full list to identify the latest image.

PR #706, opened 2017-07-11
Paginate thru' S3 bucket elements

krishnanrs:
The logic to match a MV version to an S3 bucket element was not
paginating thru' the S3 bucket list because of which it would
only attempt to match against the first 1000 entries in the bucket.


krishnanrs:
I want to get this change in quickly to unblock the Yolo failures. An ideal fix would be to check thru' each paged elements, instead of the current logic of getting the entire list of elements from the bucket. But much like the other change, the exposure here is limited as this only happens when using the internal  bucket and does not impact customer use cases where only limited number of versions are published.

PR #712, opened 2017-07-13
Add support for port-groups

apoorva-jindal:
Add support for port-groups and virtual-port-groups.
Expose option nic-type to customers to allow them to choose
the network type they want to configure.

### DIFF
--- a/brkt_cli/esx/esx_args.py
+++ b/brkt_cli/esx/esx_args.py
@@ -309,8 +309,21 @@ def add_disk_type(parser):
         '--disk-type',
         metavar='TYPE',
         dest='disk_type',
-        help='thin/thick-lazy-zeroed/thick-eager-zeroed (default: thin)',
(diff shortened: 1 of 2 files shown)


brkt_cli/esx/esx_args.py
```diff
@@ -309,12 +309,23 @@ def add_disk_type(parser):
...
         metavar='TYPE',
         dest='disk_type',
-        help='thin/thick-lazy-zeroed/thick-eager-zeroed (default: thin)',
+        help='thin/thick-lazy-zeroed/thick-eager-zeroed',
         default='thin',
         required=False
     )
 
 
+def add_nic_type(parser, help=argparse.SUPPRESS):
```
krishnanrs: Maybe make this and the disk-type as choices to prevent any user error and make validation simpler?

apoorva-jindal: Fixed.

apoorva-jindal:
Yes, it does fix the network-name issue when encrypting on the ESX host.

PR #713, opened 2017-07-13
Add interactive mode for aws encrypt

billybobjoeaglt:


brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+
+def run_interactive_encrypt_ami(values, parsed_config):
+    """
+    Runs the interactive encrypt ami mode
+    :param values: the parsed values
+    :param parsed_config: the parsed config
+    :return: new values
+    """
+    if 'BRKT_API_TOKEN' not in os.environ:  # If missing yeti API token, get it
+        email = InteractiveTextField('Enter Email').run()
```
bburtin: Do we really need OOP concepts here?  Seems like it would be more straightforward to use `raw_input()` and `getpass()`.  Or if you really need fancier behavior, I think a function would suffice.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+        root_url = 'https://%s:%d' % (
+            env.public_api_host, env.public_api_port)
+        y = YetiService(root_url)
+
+        token = y.auth(email, password)
+        os.environ['BRKT_API_TOKEN'] = token
+
+    regions = map(lambda x: (x.name, x.name), boto.ec2.regions())  # Ask which region
+    regions.sort()
+    config_region = parsed_config.get_option('aws.region')
```
bburtin: I just realized that I misled you here.  You should be checking `values.region`.  That would handle the user either specifying the region in config or on the command line.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+
+        _, env = parsed_config.get_current_env()
+        root_url = 'https://%s:%d' % (
+            env.public_api_host, env.public_api_port)
+        y = YetiService(root_url)
+
+        token = y.auth(email, password)
+        os.environ['BRKT_API_TOKEN'] = token
+
+    regions = map(lambda x: (x.name, x.name), boto.ec2.regions())  # Ask which region
```
bburtin: Do you have an instance of `AWSService` when you call this code?  We generally put boto calls there, so that we don't have boto calls sprinkled throughout the codebase.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+    regions.sort()
+    config_region = parsed_config.get_option('aws.region')
+    region_ids = [x[0] for x in regions]
+    if config_region is not None and config_region in region_ids:
+        idx = region_ids.index(config_region)
+        regions[idx] = ('%s ** Default' % regions[idx][1], regions[idx][1])
```
bburtin: I think it would be clearer to just have a simple list of regions as opposed to tuples, and a separate `default_region` variable that you compare against.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+
+    region = InteractiveSelectionNameValueMenu('Region', regions).run()
+
+    def run_get_ami_lib():  # Get all AMIs in library
+        conn = boto.ec2.connect_to_region(region)
+        images = conn.get_all_images(owners=['self'])
+        return InteractiveSelectionNameValueMenu('AMI',
```
bburtin: This list has the potential of growing out of hand.  If the user owns more than 50 images, I think we should just prompt for the ID.  Also, please use `AWSService` instead of calling boto directly.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,102 @@
...
+                                               map(lambda x:
+                                                   (x.name if x.name is not None and x.name != '' else x.id, x.id),
```
bburtin: `x.name if x.name`.  Also, consider breaking the `map()` call out to a separate line of code.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,112 @@
...
+    region = InteractiveSelectionMenu('Region', regions).run()
+    if region == default_region_formatted:
+        region = default_region_unformatted
+
+    aws_svc.connect(region)
+
+    def run_get_ami_lib():  # Get all AMIs in library
+        images = aws_svc.get_images(owners=['self'])
+        if len(images) > 50:
+            print format_error('Too many images. Please manually write the AMI.')
```
bburtin: I don't think you need a function that just prepends "Error".  It makes the person reading the code go through an unnecessary level of indirection.  And "Error: " is probably gratuitous anyway, since it's clear from the message that this is an error condition.

Also error messages are usually written to stderr, not stdout.

brkt_cli/aws/encrypt_ami_interactive.py
```diff
@@ -0,0 +1,112 @@
...
+            return None
+        formatted_images = map(lambda x: (x.name if x.name else x.id, x.id), images)
```
bburtin: Am I reading this right?  `formatted_images` is a list that either contains a name or a tuple of two ids?  That is a bit hard to get my head around.

PR #714, opened 2017-07-17
Update AWS code to use boto3

bburtin:
Update AWS code to use boto3.  The bulk of the change is in AWSService,
but higher-level code was also affected for two reasons:

1) A bunch of field names changed
2) Embedded data structures are now dictionaries instead of objects

(2) makes application code a bit icky.  I did my best to code
around this by adding utility functions to a couple of new modules:
boto3_device and boto3_tag.

I was able to encrypt and update Ubuntu, RHEL, CentOS, and Amazon Linux
images with this change.

We still have a little bit of boto2 code in share-logs.  Mark graciously
offered to update and test the share-logs code.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -17,8 +17,9 @@ import re
 import tempfile
 import urllib2
 
-import boto
-from boto.exception import EC2ResponseError, NoAuthHandlerFound
+import botocore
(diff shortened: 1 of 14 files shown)


bburtin:
Thanks Krishnan.  Mark discovered that we fail spectacularly when the user doesn't specify AWS creds.  I'll fix it and update the review.

PR #719, opened 2017-07-20
Disable legacy mode

bburtin:
Disable the legacy code path in encrypt-ami.  This code was put in to
deal with inconsistencies in block device mappings, and problems with
unmounting the root volume from the guest.  I believe that this
workaround is no longer necessary with the latest HVM images.  I tested
the non-legacy code path with Ubuntu 12.04-16.04, CentOS 6 and 7, Amazon
Linux, RHEL 6.9-7.3, and Cloud Foundry BOSH images.  None of them
required the workaround.

This commit disables legacy mode by hard-coding the legacy variable to
False.  We can remove the legacy code altogether once we're convinced
that we didn't break backward compatibility.

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -451,18 +451,32 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
             guest_instance_type='m4.large', instance_config=None,
             save_encryptor_logs=True,
             status_port=encryptor_service.ENCRYPTOR_STATUS_PORT,
-            terminate_encryptor_on_failure=True):
-    log.info('Starting encryptor session %s', aws_svc.session_id)
+            terminate_encryptor_on_failure=True, legacy=False):
+    log.info(
+        'Starting session %s to encrypt %s',
+        aws_svc.session_id,
+        image_id
+    )
+    if legacy:
+        log.warn(
+            'Using legacy encryption mode.  This mode will be removed in '
+            'the next release.'
+        )
 
     encryptor_instance = None
     snapshot_id = None
     guest_instance = None
     temp_sg_id = None
-    guest_image = aws_svc.get_image(image_id)
+
+    # Verify that the guest and encryptor images exist.
+    aws_svc.get_image(image_id)
     aws_svc.get_image(encryptor_ami)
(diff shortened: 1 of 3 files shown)


brkt_cli/aws/encrypt_ami.py
```diff
@@ -451,17 +451,32 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
...
+    # Verify that the guest and encryptor images exist.
+    aws_svc.get_image(image_id)
     aws_svc.get_image(encryptor_ami)
     encrypted_image = None
 
+
+    """
+    # TODO: Remove this code once we get rid of legacy mode.
+
     legacy = False
```
krishnanrs: Should this be removed? Otherwise it will simply override whatever is passed as the legacy mode to this method

brkt_cli/aws/encrypt_ami.py
```diff
@@ -476,6 +491,7 @@ def encrypt(aws_svc, enc_svc_cls, image_id, encryptor_ami, crypto_policy,
                  "preserved because the root disk is attached at %s "
```
krishnanrs: I'd prefer if we display the warning message about the legacy mode here. Again, not set the legacy mode flag as it will now have to be passed by the user as a command line argument.

brkt_cli/aws/encrypt_ami.py
bburtin: Yikes!  Thanks for catching this.

brkt_cli/aws/encrypt_ami.py
bburtin: Not sure what you mean.  Swing by when you get in, so we can discuss.

brkt_cli/aws/encrypt_ami.py
bburtin: Oh... this entire block of code is commented out, so it won't run.

brkt_cli/aws/encrypt_ami.py
bburtin: Krishnan and I agreed that it doesn't make sense to show a warning when the root disk isn't on /dev/sda1.  My testing didn't uncover any issues with licensing.  If we show a warning, it would mislead the user into thinking that something is wrong.

PR #720, opened 2017-07-24
Turn on ENA support if Metavisor supports it

bburtin:
Enable ENA support on the guest instance if it's enabled on the
Metavisor instance.  Shuffle around some of the code in update_ami, to
allow us to do this earlier in the process.  If things go wrong, we'll
know sooner rather than later.  Log the "setting attribute" message at
info level for greater transparency.


bburtin:
- Encrypt and update CentOS images that don't have ENA support enabled
- Verify that the encrypted images have ENA support enabled
- Verify that encrypted instances boot successfully and have ENA support enabled

PR #722, opened 2017-07-25
Get the console output from the proper attribute

krishnanrs:
With the migration to Boto3, read the console output attribute
properly from the response object. Also check if there is a valid
output before returning the same.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -557,7 +557,10 @@ class AWSService(BaseAWSService):
     def get_console_output(self, instance_id):
         response = self.ec2client.get_console_output(
             InstanceId=instance_id)
-        return response['Output']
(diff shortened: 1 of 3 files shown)


brkt_cli/aws/aws_service.py
```diff
@@ -557,7 +557,7 @@ def get_key_pair(self, keyname):
     def get_console_output(self, instance_id):
         response = self.ec2client.get_console_output(
             InstanceId=instance_id)
-        return response['Output']
```
bburtin: Can we leave this as is?  The callsite only needs the content, not the entire dictionary.

krishnanrs: I've pushed a new change with your suggested changes. I liked it this way mainly because it also gives an idea of the actual response structure.

PR #723, opened 2017-07-25
Disable OVF/OVA creation in Windows.

apoorva-jindal:
vCenter messes up the path of the vmdk during OVF
descriptor creation. Till the vCenter bug is fixed,
disable the OVF/OVA creation in Windows.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -78,6 +78,9 @@ def _get_vcenter_password(use_esx):
 def run_encrypt(values, parsed_config, log, use_esx=False):
     session_id = util.make_nonce()
     if values.create_ovf or values.create_ova:
+        # ovf/ova creation on Windows is not supported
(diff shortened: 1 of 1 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -78,6 +78,9 @@ def _get_vcenter_password(use_esx):
 def run_encrypt(values, parsed_config, log, use_esx=False):
     session_id = util.make_nonce()
     if values.create_ovf or values.create_ova:
+        # ovf/ova creation on Windows is not supported
+        if os.name is "nt":
```
bburtin: Should this be `==` instead of `is`?  `is` tests identity, not equality.  It may work in this case, but the person reading the code has to think twice to make sure the code is doing the right thing.

apoorva-jindal: You are correct. Will do another request to convert is to ===.

PR #724, opened 2017-07-25
Check root device when snapshotting the root volume

bburtin:
If the root device name can't be found in the block device mapping,
print an informative message instead of blindly continuing and failing
later.  We've seen AMIs in the field where the root_device_name is not
one of the DeviceNames in block_device_mappings.


bburtin:
Krishnan convinced me to abandon this pull request.  We don't know if this is still an issue.  We'll err toward less code until we know we need it.

PR #725, opened 2017-07-25
Disable credentials for S3 bucket pagination

krishnanrs:
When paginating thru' the S3 bucket objects to identify locate
the required metavisor version, disable credentials request
as these buckets are public. AWS credentials should not be
required when trying to download images especially for ESX.

### DIFF
--- a/brkt_cli/mv_version.py
+++ b/brkt_cli/mv_version.py
@@ -62,8 +62,7 @@ def get_s3_versions(bucket):
 
     s3 = boto3.resource('s3')
     s3.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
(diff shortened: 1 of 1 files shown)


brkt_cli/mv_version.py
```diff
@@ -62,7 +62,7 @@ def get_s3_versions(bucket):
 
     s3 = boto3.resource('s3')
     s3.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
-    client = boto3.client('s3')
+    client = s3.meta.client
```
bburtin: More of a style thing, but you could move this one line up, to avoid doing `s3.meta.client` twice.

PR #727, opened 2017-07-26
Add command to configure static-ip on VM

apoorva-jindal:
Assume that the VM has been created via encryptor which was
assigned a static IP. Thus, it has the old IP retained from
the encryption time.

### DIFF
--- a/brkt_cli/esx/__init__.py
+++ b/brkt_cli/esx/__init__.py
@@ -50,6 +50,7 @@ from brkt_cli.esx import (
     wrap_image,
     wrap_with_vcenter_args,
     wrap_with_esx_host_args,
+    assign_static_ip_args,
 )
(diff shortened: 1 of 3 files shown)


brkt_cli/esx/__init__.py
```diff
@@ -562,6 +563,53 @@ def run_wrap_image(values, parsed_config, log, use_esx=False):
...
+            nic_type=None,
+            verify=values.validate,
+            cdrom=None,
+        )
+    except Exception as e:
+        log.exception(e)
+        raise ValidationError("Failed to connect to vCenter ", e)
+    try:
+        vm = vc_swc.find_vm(values.vm)
+        if vm is None:
```
bburtin: `if not vm`

PR #728, opened 2017-07-27
Add config key for security group

bburtin:
Add an aws.security-group config key, for specifying the default
security group when launching a Metavisor instance during encryption,
update, and wrap-guest-image.


brkt_cli/aws/__init__.py
```diff
@@ -457,6 +457,11 @@ def setup_config(self, config):
             'aws.subnet',
             'The AWS subnet metavisors will be launched into'
         )
+        config.register_option(
```
krishnanrs: Do you want to mention here that only one security group is supported right now? Otherwise the changes lgtm.

PR #730, opened 2017-07-28
Share-logs boto3 upgrade

markvoll:
This change updates share-logs code to use boto3


brkt_cli/aws/test_share_logs.py
```diff
@@ -13,65 +13,86 @@
...
-            return True
-        return False
+    def Object(self, path):
+        return self.objects
 
-    def get_acl(self):
-        return self.acp
 
-class ACP():
+class Buckets():
```
bburtin: Are you sure that you want to go down the road of replicating the boto3 API?  I'm worried that this will turn into a slippery slope.  Would it make sense to abstract out the service like we do with `AWSService`, so that the caller passes parameters as primitive types and gets an object back?

PR #731, opened 2017-07-28
Add OVF properties to configure static IP

apoorva-jindal:
This allows customers to allow addition of OVF properties
to configure static IP in the template VM or the encrypted OVF.
Primarily design for use to encrypt Ops Manager in PCF, but
is a generic implementation and can be used for any customer
which wants to use static IP.


apoorva-jindal:
Update will inherit the properties from encryption. One can directly assign a static ip to the wrap image in brkt-cli and no need for ovf properties to be able to do that.

PR #735, opened 2017-08-03
Add option for specifying the public API root cert

bburtin:
Add a --public-api-root-cert option to all commands that talk to the
Bracket service.  This option is used for specifying the path to the
self-signed root cert when brkt-cli talks to a Customer Managed MCP.

Also add a config environment key for this option.

Add crypto.validate_cert_path() for validating a cert that's stored on
disk.  Call this function when validating the --public-api-root-cert and
--ca-cert values.

### DIFF
--- a/brkt_cli/__init__.py
+++ b/brkt_cli/__init__.py
@@ -24,12 +24,11 @@ import sys
 import tempfile
 from operator import attrgetter
 
-from brkt_cli import brkt_jwt, util, version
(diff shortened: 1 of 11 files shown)


bburtin:
Yep, already there.  😄

krishnanrs:
Must have missed it!

bonzotronix:
Belated LGTM (FWIW)

PR #736, opened 2017-08-08
Documentation updates

krishnanrs:
Added information on encrypting stock images on AWS as well as the
corresponding new argument (`--stock-image-version`)


billybobjoeaglt:
LGTM. Just can you add that we only support centos 6 and 7 and you have to subscribe to the centos images in the marketplace beforehand. Thanks!

PR #739, opened 2017-08-14
Temporary fix when rename VM does not work.

apoorva-jindal:
In case rename VM does not work, e.g. for VSAN, use the not-error-free
solution where the old template vm gets deleted, and then the new one
gets created. Thus, this can cause issues if creation of new template
fails because old one gets deleted. Committing this temporary workaround
till can find a more complete fix without rename.


krishnanrs:
I'm OK with this change as a workaround, but this change creates the risk wherein if after the template deletion, the re-creation of the template fails, then we are left in limbo.

PR #743, opened 2017-08-15
Remove timestamp parsing

bburtin:
Remove the ability to specify the --exp and --nbf token values as
timestamps.  This behavior was deprecated in a previous release because
we want the server time to be the source of truth.  Clients now pass
timestamp values as a duration and the server does the math.


bburtin:
Aditya and I decided to keep timestamp parsing in the CLI.  It's convenient to be able to create a token that expires at a given date and time.

PR #752, opened 2017-08-30
Use the version of Metavisor returned by the Bracket service

bburtin:
Get the Metavisor version from the Bracket service instead of using the
latest Metavisor release.  With a Customer Managed MCP that has not been
upgraded, the latest Metavisor may attempt to call service endpoints
that are not available.

### DIFF
--- a/brkt_cli/aws/__init__.py
+++ b/brkt_cli/aws/__init__.py
@@ -11,6 +11,7 @@
 # CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and
 # limitations under the License.
+import httplib
 import json
 import logging
 import re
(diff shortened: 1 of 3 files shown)


dasein:
LGTM - The only thing I was worried about was changing mv_version.py code and affecting the ESX cloud provider. But it looks like we did not touch any functions that trigger that code path.

bburtin:
Herb gave me the verbal 👍

brkt_cli/aws/__init__.py
```diff
@@ -850,22 +858,44 @@ def _get_encryptor_ami(region_name, version):
...
+    all_buckets = (
+        mv_version.AWS_PROD_BUCKET,
+        mv_version.AWS_STAGE_BUCKET,
+        mv_version.AWS_BUILD_BUCKET
+    )
+
+    for bucket in all_buckets:
+        try:
+            amis_url = mv_version.get_amis_url(version, bucket)
+        except mv_version.MetavisorVersionNotFoundError:
```
krishnanrs: What about the case where a user does not have access to a given bucket, which is especially the case with the AWS_BUILD_BUCKET for users not coming from the corporate network?

bburtin: We either catch `MetavisorVersionNotFoundError` or get a `403` when accessing the build bucket.  That's handled a few lines down.  I tested to make sure the code does what it's supposed to when you're not on the internal network.

PR #753, opened 2017-08-31
Configure volumes for wrapped instance to be deleted on termination

krishnanrs:
Detaching and re-attaching volumes cause them to lose their
DeleteOnTermination property, which needs to be reset on the
running instance.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -697,6 +697,13 @@ class AWSService(BaseAWSService):
                 InstanceId=instance_id,
                 UserData={'Value': value}
             )
+        elif attribute == 'blockDeviceMappings':
+            log.info('Setting blockDeviceMapping for %s to %s',
+                     instance_id, value)
+            modify_instance_attribute(
+                InstanceId=instance_id,
+                BlockDeviceMappings=value
+            )
         else:
             log.info('Setting %s for %s to %s', attribute, instance_id, value)
             modify_instance_attribute(
--- a/brkt_cli/aws/wrap_image.py
+++ b/brkt_cli/aws/wrap_image.py
@@ -231,8 +231,25 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
             aws_svc, instance.id, instance.root_device_name)
         aws_service.wait_for_volume_attached(aws_svc, instance.id, '/dev/sdf')
 
+        # Create guest and metavisor device mappings to be deleted on termination
+        guest_device = boto3_device.make_device(
+            device_name='/dev/sdf',
+            delete_on_termination=True
+        )
+        metavisor_device = boto3_device.make_device(
+            device_name=instance.root_device_name,
+            delete_on_termination=True
+        )
+
         log.info('Starting wrapped instance.')
         instance = aws_svc.start_instance(instance.id)
+        # Re-attached volumes lose their DeleteOnTermination Settings
+        # Modify instance attributes to mark the guest root and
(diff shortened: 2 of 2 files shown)


brkt_cli/aws/aws_service.py
```diff
@@ -697,6 +697,13 @@ def modify_instance_attribute(self, instance_id, attribute,
                 InstanceId=instance_id,
                 UserData={'Value': value}
             )
+        elif attribute == 'blockDeviceMapping':
```
bburtin: Should this be `blockDeviceMappings` to stay consistent with the attribute name?

brkt_cli/aws/wrap_image.py
```diff
@@ -231,8 +231,23 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
...
+        )
+        metavisor_bdm = boto3_device.make_device(
+            device_name=instance.root_device_name,
+            volume_id=mv_root_vol.id,
+            delete_on_termination=True
+        )
+
         log.info('Starting wrapped instance.')
         instance = aws_svc.start_instance(instance.id)
+        aws_svc.modify_instance_attribute(
```
bburtin: When you modify the BDM and pass only the Metavisor and guest devices, does it remove the other device entries?  If not, I think it's still worth adding a comment about why you're making this call.  The first time I read the code, I expected that the other devices would get detached.

brkt_cli/aws/wrap_image.py
```diff
@@ -231,8 +231,23 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
             aws_svc, instance.id, instance.root_device_name)
         aws_service.wait_for_volume_attached(aws_svc, instance.id, '/dev/sdf')
 
+        guest_bdm = boto3_device.make_device(
```
bburtin: `guest_device` would be more consistent with `make_device()`.

brkt_cli/aws/wrap_image.py
```diff
@@ -231,8 +231,23 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
...
+            device_name='/dev/sdf',
+            volume_id=guest_root_vol.id,
+            delete_on_termination=True
+        )
+        metavisor_bdm = boto3_device.make_device(
```
bburtin: `metavisor_device`?

brkt_cli/aws/wrap_image.py
krishnanrs: It only modifies the attributes of the devices that are passed. As is, the wrap image only attaches the MV and guest root volumes. The attributes of the other volumes (if they are attached) would remain unchanged.

brkt_cli/aws/wrap_image.py
```diff
@@ -231,8 +231,27 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
             aws_svc, instance.id, instance.root_device_name)
         aws_service.wait_for_volume_attached(aws_svc, instance.id, '/dev/sdf')
 
+        # Create guest and metavisor device mappings to be deleted on termination
+        guest_device = boto3_device.make_device(
+            device_name='/dev/sdf',
+            volume_id=guest_root_vol.id,
```
bburtin: Is `volume_id` required?  If not, may as well take it out, so that it's clearer what you're trying to accomplish.

krishnanrs: Its not required, but I thought it would just remove all doubts. Anyways, I pushed a new change without the volume_id

PR #754, opened 2017-08-31
Propagate the guest-instance-type for encryptor

bobveznat:
Prior to this change the --guest-instance-type option didn't seem to
really do anything for `brkt aws encrypt`. Now it is propagated through
to the main runinstances call and launches the instance type you want.

If you don't specify the option the application default (currently
c4.2xlarge) gets used.

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -112,8 +112,8 @@ def _get_description_from_image(image):
 
 def _run_encryptor_instance(
         aws_svc, encryptor_image_id, snapshot, root_size, guest_image_id,
-        crypto_policy, security_group_ids=None, subnet_id=None, placement=None,
-        instance_config=None,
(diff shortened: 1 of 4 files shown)


brkt_cli/aws/encrypt_ami.py
```diff
@@ -178,6 +179,9 @@ def _run_encryptor_instance(
 
         bdm = [guest_unencrypted_root, guest_encrypted_root]
         log.info('Launching encryptor instance.')
+        extra_args = {}
+        if instance_type is not None:
```
bburtin: `is not None` is gratuitous

krishnanrs:
Are you sure the guest_instance_type is not being passed correctly? I do know that we do not pass it up for the Metavisor instance and that defaults to c4.xlarge. The guest_instance_type actually defaults to m4.large. In any case, the guest_instance is launched just to get the snapshot of the guest root volume while the actual encryption happens on the Metavisor instance.

PR #756, opened 2017-08-31
Add --encryptor-instance-type to aws encrypt

bobveznat:
We were defaulting to c4.xlarge, a terrific choice, but this instance
requires a VPC. For people with old AWS accounts and EC2 classic this
preculuded them from doing encryption outside of a specific VPC.

The update subcommand already has an --updater-instance-type option.

### DIFF
--- a/aws.md
+++ b/aws.md
@@ -154,7 +154,8 @@ be in the same VPC.
 $ brkt aws encrypt --help
 usage: brkt aws encrypt [-h] [--stock-image-version STOCK_IMAGE_VERSION]
                         [--encrypted-ami-name NAME]
-                        [--guest-instance-type TYPE] [--no-validate]
+                        [--guest-instance-type TYPE]
+                        [--encryptor-instance-type TYPE] [--no-validate]
                         --region NAME [--security-group ID] [--subnet ID]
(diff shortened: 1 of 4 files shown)


bburtin:
If we want to go for extra credit... how hard would it be to automatically choose a compatible instance type when there's no default VPC?  I believe that there's already validation code that looks up the default VPC when a subnet is specified.

bobveznat:
The answer seems to be "probably". You can use this API to figure out if a default VPC exists: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeAccountAttributes.html

bburtin:
Just took a closer look.  There is a `get_default_vpc()` function in `AWSService`.  You should be able to choose a different instance type if that function returns `None`.  If this works, I'd lean toward not adding the encryptor instance type option until we know we needed, so we have fewer knobs to turn.  Go ahead and make the call about whether you think it's worth keeping the option.

dramdass:
One issue I've faced in the past is AWS out-of-capacity of the hard-coded instance type which is why a flag would be nice - or have code automatically detect the out-of-capacity error and try different instance types.

PR #757, opened 2017-09-07
Bump version to 1.0.12

krishnanrs:


bburtin:
I don't think this needs to be a pull request, since we don't need to merge the commit into master.  You can bump the version to 1.0.12 in the release branch and not merge the change.  Then update the code version to 1.0.13pre1 and release versions in README.md to 1.0.12 in a separate commit, which is merged to the master branch.

PR #758, opened 2017-09-08
Validate access to image

krishnanrs:
Boto3 returns an image object, if the image exists, even if the user
does not access to the image. This change ensure that the user also
has access to the image before proceeding. This ensures that there are
no unexpected errors downstream.

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -539,6 +539,11 @@ class AWSService(BaseAWSService):
                 image.load, r'InvalidAMIID\.NotFound')
 
         load()
+        try:
+            image.name
+        except AttributeError:
+            return None
+
         return image
 
     def delete_snapshot(self, snapshot_id):


bburtin:
Looks good, but can you add a comment so that the next person understands what you're trying to do?

brkt_cli/aws/aws_service.py
```diff
@@ -539,6 +539,11 @@ def get_image(self, image_id, retry=False):
                 image.load, r'InvalidAMIID\.NotFound')
 
         load()
+        try:
+            image.name
+        except AttributeError:
+            return None
```
bonzotronix: Looking briefly, I'm not seeing that the get_image() callers are obviously handling a None return value properly (with a useful error message and such).  Do we need to add such checks to the callers?  Or maybe just go ahead and throw an exception here (perhaps after logging a useful error message)?

krishnanrs: The alternative is that I can raise a validation error here itself and not rely on any of the callers.

bonzotronix: I think that would be reasonable (with a nice error message logged somehow, of course).

krishnanrs: Ah, the problem is that get_image call itself is wrapped around a try block where it is expecting a ClientError. Hence I would either have to raise a ClientError, just for this particular case, or let the callers handle it. For now I would prefer to go the latter route and can fix other issues as they arise. I have tested encrypt, update and wrap-image with this change and it appears to work correctly.

PR #759, opened 2017-09-08
Bump version of 1.0.13pre1

krishnanrs:


bburtin:
I believe there's another "1.0.11" in the Docker container example in `README.md`.

krishnanrs:
Updated the docker reference to the old version string.

PR #760, opened 2017-09-10
Enable sriovNet and ENA for wrapped image/instance

krishnanrs:
Enable sriovNetSupport and ENA for a wrapped image or instance, if
necessary

### DIFF
--- a/brkt_cli/aws/aws_service.py
+++ b/brkt_cli/aws/aws_service.py
@@ -1194,13 +1194,29 @@ def get_code_and_message(client_error):
     )
 
 
-def has_ena_support(instance):
+def has_ena_support(resource):
     """ Return True if the given instance has ENA support enabled.  We have
     to do this because the AWS API does not always return the enaSupport
     field.
 
-    :param instance an ec2.Instance object
+    :param instance an ec2.Instance or ec2.Image object
     """
-    if not hasattr(instance, 'ena_support'):
+    if not hasattr(resource, 'ena_support'):
         return False
-    return bool(instance.ena_support)
+    return bool(resource.ena_support)
+
+
+def enable_sriov_net_support(aws_svc, instance):
+    """ Enable sriovNetSupport on instance, if required
+    """
+    if instance.sriov_net_support != "simple":
+        log.info('Enabling sriovNetSupport for %s', instance.id)
+        try:
+            aws_svc.modify_instance_attribute(
+                instance.id,
+                "sriovNetSupport",
+                "simple")
+            log.info('sriovNetSupport enabled successfully')
+        except ClientError as e:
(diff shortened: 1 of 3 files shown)


brkt_cli/aws/aws_service.py
```diff
@@ -1194,13 +1194,29 @@ def get_code_and_message(client_error):
...
+    :param instance an ec2.Instance or ec2.Image object
     """
-    if not hasattr(instance, 'ena_support'):
+    if not hasattr(resource, 'ena_support'):
         return False
-    return bool(instance.ena_support)
+    return bool(resource.ena_support)
+
+
+def enable_sriovNetSupport(aws_svc, instance):
```
bburtin: This should probably be `enable_sriov_net_support`, to follow convention.

brkt_cli/aws/aws_service.py
```diff
@@ -1194,13 +1194,29 @@ def get_code_and_message(client_error):
...
+    if instance.sriov_net_support != "simple":
+        log.info('Enabling sriovNetSupport for %s', instance.id)
+        try:
+            aws_svc.modify_instance_attribute(
+                instance.id,
+                "sriovNetSupport",
+                "simple")
+            log.info('sriovNetSupport enabled successfully')
+        except ClientError as e:
+            log.warn('Unable to enable sriovNetSupport for instance '
```
bburtin: Just to confirm: if modifying the instance attribute fails, the user will still have a working instance?  If this isn't the case, we should probably abort.

brkt_cli/aws/wrap_image.py
```diff
@@ -241,6 +242,25 @@ def wrap_instance(aws_svc, instance_id, metavisor_ami, instance_config=None):
             delete_on_termination=True
         )
 
+        # Enable sriovNetSupport
```
bburtin: Is there enough overlap here that we can turn this into two functions that are called from the encrypt, update, and wrap code?

brkt_cli/aws/aws_service.py
krishnanrs: That is correct. The instance will continue to work, just will not be able to utilize enhanced networking.

brkt_cli/aws/wrap_image.py
krishnanrs: I didn't understand this comment. sriovNetSupport is enabled only in encrypt and wrap-instance/image command. For this I moved it into a common function which is being invoked from these two places. Am I missing something? Please clarify.

bburtin: Oh... ok.  I remembered something about sriovNet from before, and assumed it was in the encrypt path.  Sorry if this is a false alarm.

PR #763, opened 2017-09-18
Updates to AWS reference CloudFormation templates

krishnanrs:
Updated the AWS reference Cloudformation templates to use:
 - NATGateway instead of NAT Instances
 - A Bastion Host for SSH in to private subnet instances
 - Provided a command line flag to the python script to disable the
bastion instances if the user desires

### DIFF
--- a/reference_templates/vpc-dual-az-with-nat.py
+++ b/reference_templates/vpc-dual-az-with-nat.py
@@ -23,6 +23,7 @@ from troposphere.ec2 import (
     EIP,
     Instance,
     InternetGateway,
+    NatGateway,
     NetworkAcl,
     NetworkAclEntry,
     PortRange,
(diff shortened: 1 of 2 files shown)


reference_templates/vpc-dual-az-with-nat.py
```diff
@@ -104,14 +114,23 @@ def add_autorecovery_to_instance(template_obj, instance):
 
 # These are standard freely available Amazon NAT AMIs
 t.add_mapping('RegionMap', {
+    'us-east-1': {
+        'NATAMI': 'ami-4fffc834',
```
bobveznat: Can we rename NATAMI to BASTIONAMI or something? Other than that I'm fine with this.

krishnanrs:
Updated. Please review/approve.

bobveznat:
lgtm

PR #765, opened 2017-10-09
Verify token_type parameter in user specified token

krishnanrs:
This changes verifies that there is a brkt.token_type attribute in
the payload of the user specified token and also ensures that this
attribute's value is set to 'Launch'.

### DIFF
--- a/brkt_cli/instance_config_args.py
+++ b/brkt_cli/instance_config_args.py
@@ -272,6 +272,15 @@ def get_launch_token(values, cli_config):
         tags = brkt_jwt.brkt_tags_from_name_value_list(values.brkt_tags)
         token = y.create_launch_token(tags=tags)
 
+    # If a token_type exists, then make sure that it is a launch token
+    payload = brkt_jwt.get_payload(token)
+    if payload.get('brkt.token_type'):
+        if 'launch' != payload.get('brkt.token_type').lower():
+            raise ValidationError(
+                'Token is not a launch token. Please generate a new launch '
+                'token from the Bracket Management Console'
+            )
+
     return token
 
 


brkt_cli/instance_config_args.py
```diff
@@ -265,6 +265,17 @@ def get_launch_token(values, cli_config):
...
+    if not payload.get('brkt.token_type'):
+        raise ValidationError(
+            'Token payload does not contain a brkt.token_type field. Please '
+            'generate a new launch token from the Bracket Management Console'
+        )
+    if 'launch' != payload.get('brkt.token_type').lower():
+        raise ValidationError(
+            'Token is not a launch token. Please generate a new launch token '
+            'from the Bracket Management Console'
+        )
```
bonzotronix: Wondering if it would be sensible to move this logic to line 274 (after potentially getting the launch token from yeti), as a sanity check for getting launch tokens from Yeti.

krishnanrs: Any suggestions on the error message are also welcome! :)

brkt_cli/instance_config_args.py
```diff
@@ -265,6 +265,17 @@ def get_launch_token(values, cli_config):
...
+    payload = brkt_jwt.get_payload(token)
+    if not payload.get('brkt.token_type'):
+        raise ValidationError(
+            'Token payload does not contain a brkt.token_type field. Please '
+            'generate a new launch token from the Bracket Management Console'
+        )
+    if 'launch' != payload.get('brkt.token_type').lower():
+        raise ValidationError(
+            'Token is not a launch token. Please generate a new launch token '
+            'from the Bracket Management Console'
```
bonzotronix: Error messages look pretty good to me.

brkt_cli/instance_config_args.py
krishnanrs: Are we so unsure about our own token generation mechanism that we would want to validate it post-generation? I can make the change, but my intention was only validate the user specified token.

bonzotronix: OK, I can trust that we won't mess up the launch token generation.

brkt_cli/instance_config_args.py
```diff
@@ -272,6 +272,18 @@ def get_launch_token(values, cli_config):
         tags = brkt_jwt.brkt_tags_from_name_value_list(values.brkt_tags)
         token = y.create_launch_token(tags=tags)
 
+    payload = brkt_jwt.get_payload(token)
+    if not payload.get('brkt.token_type'):
```
aditya-brkt: This will break the existing functionality because locally generated tokens (where user uploads the pub key to yeti) may not have brkt.token_type claim in them.
The check should be:
if token has brkt.token_type claim && brkt.token_type != "Launch":
    don't allow this token to be used as launch token.

bonzotronix: Should we change brkt-cli's "make-token" command so that it includes the "brkt.token_type = Launch" in the tokens it generates with a keypair?

aditya-brkt: No I don't think so. I'd like the "brkt.token_type" to mean that the token was generated on Yeti. Also, the customers may not always use brkt-cli to generate tokens locally.

PR #766, opened 2017-10-11
Update image failing for AMZN Linux

krishnanrs:
AMZN Linux images use /dev/xvda as the default root device whereas
the CLI updater assumes the root device to be always present at
/dev/sda1. As a result, the update image was failing as it was not
reading the volume information from the instance block device mappings
using the correct attach point for AMZN Linux images.

### DIFF
--- a/brkt_cli/aws/update_ami.py
+++ b/brkt_cli/aws/update_ami.py
@@ -130,6 +130,7 @@ def update_ami(aws_svc, encrypted_ami, updater_ami, encrypted_ami_name,
             aws_svc, encrypted_guest.id, state="running")
         updater = wait_for_instance(aws_svc, updater.id, state="running")
 
+        encrypted_guest_mv_root_device_name = encrypted_guest.root_device_name
         aws_svc.stop_instance(encrypted_guest.id)
         encrypted_guest = wait_for_instance(
             aws_svc, encrypted_guest.id, state="stopped")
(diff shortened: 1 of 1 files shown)


brkt_cli/aws/update_ami.py
```diff
@@ -255,10 +256,10 @@ def update_ami(aws_svc, encrypted_ami, updater_ami, encrypted_ami_name,
         aws_svc.attach_volume(
             new_mv_vol_id,
             encrypted_guest.id,
-            MV_ROOT_DEVICE_NAME
+            encrypted_guest_root_device_name
```
bonzotronix: Can you confirm that you do *not* want to also replace MV_ROOT_DEVICE_NAME on (old) line 241?

brkt_cli/aws/update_ami.py
```diff
@@ -214,14 +215,14 @@ def update_ami(aws_svc, encrypted_ami, updater_ami, encrypted_ami_name,
...
-            if name == MV_ROOT_DEVICE_NAME:
+            if name == encrypted_guest_root_device_name:
                 new_dev['Ebs']['DeleteOnTermination'] = True
                 new_dev['Ebs']['VolumeType'] = 'gp2'
 
             new_bdm.append(new_dev)
 
         # Step 4. Detach old BSD drive(s) and delete from encrypted guest
-        old_mv_dev = boto3_device.get_device(guest_bdm, MV_ROOT_DEVICE_NAME)
+        old_mv_dev = boto3_device.get_device(guest_bdm, encrypted_guest_root_device_name)
```
bonzotronix: Wait, is this disk the Metavisor root or the encrypted guest root?  The comment suggests the former, but the new variable name suggests the latter.

brkt_cli/aws/update_ami.py
krishnanrs: That is correct. In that case you can see that is is referring to the updater instance, which *is* a Metavisor image and will have the MV root at /dev/sda1 as usual.

brkt_cli/aws/update_ami.py
krishnanrs: We are basically getting the MV device from the original (encrypted) guest image/instance. This is the right way instead of assuming that this would always be /dev/sda1.

bonzotronix: I think I see my confusion: I've been parsing the variable name as "the device name of the encrypted guest root" instead of "the root device name" of the "encrypted guest" (instance).  Maybe "encrypted_instance_root_device_name"?

krishnanrs: I am OK with that. I picked the current name because we are getting the root device name of the encrypted guest (encrypted_guest_root_device_name). Although we have an encrypted instance, it is launched purely to do the upgrade of the MV. Let me know if you still insist on the name change and I will go ahead and make the change.

bonzotronix: It would be good to clarify things with the variable name somehow.  "encrypted_guest_mv_root_device_name"?

krishnanrs: Please review again

PR #767, opened 2017-10-12
Validate instance type for wrap operations

krishnanrs:
Metavisor requires instances with atleast 1 vCPU and 1GB RAM. This
change validates that unsupported instances type (t1.micro and
t2.nano in AWS and f1-micro in GCP) are not allowed with the wrap
commands


krishnanrs:
Missed the validation for the GCP `launch` command. Added the same. Also added a unit test for the AWS instance type validation.

PR #770, opened 2017-10-19
YETI-3635: Add support for single-disk encryption on GCP

xcllnt:
Single-disk encryption is a way of encrypting whereby the encrypted image has the same number of disks as the unencrypted/stock image. While currently we assume there's only a root disk, the mechanism extends to images with multiple disks.

The immediate upshot is that bracketized images can be launched from the GCP or Azure console and that we therefore don't need to use the launch command in the CLI. There are less immediate advantages as well on AWS and ESX, which is why single-disk encryption is not specific to any CSP, but rather the new way to encrypt across all platforms.

Globally:
 o  Introduce METAVISOR_DISK_SIZE. It holds the size of the
    metavisor disk.
 o  Add options --single-disk and --no-single-disk
    Provide a reasonable default when neither is specified.
    Add 'single-disk' to the bracket configuration for GCP.
 o  Add encryptor_did_single_disk()

For GCP:
 o  Support encrypting legacy and single-disk on GCP.
 o  Support updating legacy and single-disk on GCP

For AWS and ESX:
 o  Remove single_disk from the bracket configuration. They'll
    support single-disk encryption in a follow-up commit to
    minimize risk.

While here:
 o  Remove redundant handling of crypto_policy_type

Jira: YETI-3635

### DIFF
--- a/brkt_cli/aws/encrypt_ami.py
+++ b/brkt_cli/aws/encrypt_ami.py
@@ -118,7 +118,6 @@ def _run_encryptor_instance(
 
     if instance_config is None:
         instance_config = InstanceConfig()
-    instance_config.brkt_config['crypto_policy_type'] = crypto_policy
 
     # Use 'sd' names even though AWS maps these to 'xvd'
     # The AWS GUI only exposes 'sd' names, and won't allow
--- a/brkt_cli/encryptor_service.py
+++ b/brkt_cli/encryptor_service.py
@@ -65,6 +65,21 @@ class BaseEncryptorService(object):
         self.hostnames = hostnames
         self.port = port
 
+    def fetch(self, path, timeout_secs=2):
+        exceptions_by_host = {}
+        for hostname in self.hostnames:
+            url = 'http://%s:%d%s' % (hostname, self.port, path)
+            try:
+                r = urllib2.urlopen(url, timeout=timeout_secs)
+                data = r.read()
+                self.hostnames = [hostname]
+                return data
+            except IOError as e:
+                log.debug('Unable to connect to %s:%s - %s', hostname,
+                          self.port, e)
+            exceptions_by_host[hostname] = e
+        raise EncryptorConnectionError(self.port, exceptions_by_host)
+
     @abc.abstractmethod
     def is_encryptor_up(self):
         pass
@@ -100,42 +115,17 @@ class EncryptorService(BaseEncryptorService):
             log.debug("Couldn't get encryptor status: %s", e)
             return False
 
-    def get_status(self, timeout_secs=2):
-        exceptions_by_host = {}
-        info = None
-        successful_hostname = None
-
-        for hostname in self.hostnames:
-            url = 'http://%s:%d' % (hostname, self.port)
-            try:
-                r = urllib2.urlopen(url, timeout=timeout_secs)
-                data = r.read()
-            except IOError as e:
-                log.debug(
-                    'Unable to connect to %s:%s - %s',
-                    hostname, self.port, e)
-                exceptions_by_host[hostname] = e
-                continue
-
-            info = json.loads(data)
-            info['percent_complete'] = 0
-            bytes_total = info.get('bytes_total')
-            if info['state'] == ENCRYPT_SUCCESSFUL:
-                info['percent_complete'] = 100
-            elif ((bytes_total is not None) and
-                  (bytes_total > 0)):
-                ratio = float(info['bytes_written']) / info['bytes_total']
-                info['percent_complete'] = int(100 * ratio)
-            successful_hostname = hostname
-            break
-
-        if info:
-            # Don't try the other hostnames again, now that we have one that
-            # is known to work.
-            self.hostnames = [successful_hostname]
-            return info
-        else:
-            raise EncryptorConnectionError(self.port, exceptions_by_host)
+    def get_status(self):
+        data = self.fetch('/')
+        info = json.loads(data)
+        info['percent_complete'] = 0
+        bytes_total = info.get('bytes_total')
+        if info['state'] == ENCRYPT_SUCCESSFUL:
+            info['percent_complete'] = 100
+        elif bytes_total is not None and bytes_total > 0:
+            ratio = float(info['bytes_written']) / info['bytes_total']
+            info['percent_complete'] = int(100 * ratio)
+        return info
 
 
 def wait_for_encryptor_up(enc_svc, deadline):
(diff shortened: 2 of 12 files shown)


brkt_cli/aws/encrypt_ami.py
```diff
@@ -118,7 +118,6 @@ def _run_encryptor_instance(
 
     if instance_config is None:
         instance_config = InstanceConfig()
-    instance_config.brkt_config['crypto_policy_type'] = crypto_policy
```
krishnanrs: I believe we still need to pass this value in the instance_config to the encryptor. I am not sure if the metavisor defaults to XTS if this is not present, however in case the user specifies a different crypto policy using the command line args, this field would be required.

brkt_cli/gcp/encrypt_gcp_image.py
```diff
@@ -4,178 +4,152 @@
...
+        gcp_svc.wait_for_detach(values.zone, instance_name)
+        guest_disk_size = gcp_svc.get_disk_size(values.zone, instance_name)
+
+        # Create a blank disk. The size of which depends on the crypto
+        # policy and whether we do single-disk encryption or not.
+        encrypted_disk_size = guest_disk_size
+        if values.crypto == CRYPTO_GCM:
+            encrypted_disk_size *= 2
+        if values.single_disk:
+            encrypted_disk_size += METAVISOR_DISK_SIZE
```
krishnanrs: Shouldn't this be METAVISOR_DISK_SIZE + 1 to accommodate for the encryption headers?

bburtin:
Can you please update the commit comment with some information about what single disk encryption is, and why we're making this change?  It would really help reviewers understand the motivation of the work, especially if they've been (ahem) out of the loop for a while.

brkt_cli/gcp/encrypt_gcp_image.py
```diff
@@ -4,178 +4,152 @@
...
-        log.info("Creating metavisor image")
-        gcp_svc.create_gcp_image_from_disk(zone, encrypted_image_name, encryptor)
+        if not values.single_disk:
+            # Snapshot the encrypted guest disk
+            log.info("Creating snapshot of encrypted image disk")
+            gcp_svc.create_snapshot(values.zone, encrypted_image_disk,
+                                    encrypted_image_name)
+            root_disk = encryptor
+        else:
+            root_disk = encrypted_image_disk
```
krishnanrs: So is the actual magic of creating a MV and guest partition in the single disk being handled by the MV/encryptor code?

brkt_cli/encryptor_service.py
```diff
@@ -260,7 +250,14 @@ def wait_for_encryption(enc_svc,
     raise EncryptionError('Encryption service unavailable')
 
 
+def encryptor_did_single_disk(enc_svc):
+    sd = enc_svc.fetch('/single_disk')
```
bburtin: Do we really want an API endpoint that returns `text/plain`?  Would it make sense to add some sort of Metavisor properties and point and have single-disk encryption be a field in the response?  That way we don't need to add a new endpoint for every property we add in the future.

brkt_cli/gcp/__init__.py
```diff
@@ -50,6 +53,19 @@ def run_encrypt(values, config):
     if not values.verbose:
         logging.getLogger('googleapiclient').setLevel(logging.ERROR)
 
+    if not values.encryptor_image:
```
bburtin: Should this work be in a separate commit?  It's not immediately obvious how it relates to single-disk encryption.

brkt_cli/gcp/__init__.py
```diff
@@ -60,25 +76,15 @@ def run_encrypt(values, config):
         brkt_env=brkt_env,
         launch_token=lt
     )
+
     encrypted_image_id = encrypt_gcp_image.encrypt(
         gcp_svc=gcp_svc,
         enc_svc_cls=encryptor_service.EncryptorService,
-        image_id=values.image,
-        encryptor_image=values.encryptor_image,
+        values=values,
```
bburtin: I'm not sure we want to go in this direction.  I understand that it's painful to have 20 arguments to a function, but passing in an unstructured dictionary is also not great.  Someone reading the `encrypt()` source won't know which values are being passed in.  If you want to encapsulate the arguments, you could add an `EncryptValues` class or something similar.

brkt_cli/aws/encrypt_ami.py
xcllnt: Common code now sets crypto_policy_type. See instance_config_args.py.

brkt_cli/gcp/encrypt_gcp_image.py
xcllnt: No need. There's enough unallocated space on the MV disk to cover the overhead.

brkt_cli/gcp/encrypt_gcp_image.py
xcllnt: Yes. The prime responsibility of the CLI is to size the empty disk correctly and to construct the final image with the right disks. The metavisor will do the rest.

brkt_cli/encryptor_service.py
xcllnt: Why not? There's nothing wrong with another endpoint. And since this is temporary, it's actually better not to add it to the status. This is especially true given how the metavisor is implemented.

brkt_cli/gcp/__init__.py
xcllnt: It's part of the same same restructuring of eliminating a zillion arguments and only passing the values. By moving this up one level, we have a clear separation between filling in the blanks at the top-most CSP-specific action and read-only use further down the stack.

brkt_cli/gcp/__init__.py
xcllnt: It's not merely painful, it's actually worse. Adding an extra argument was beyond my capabilities.

Note that arguments are actually a lot more unstructured than a name space, especially if you start providing default arguments. Looking at a function you simply won't know where values are coming from, which means that understanding the code is a lot harder than it has to be.

Passing values as-is, makes it a lot clearer that they're effectively coming from the command-line with some post-processing. Moving that post-processing, as high up the stack as possible and treating values as read-only as much as possible, makes the logic a lot more understandable.

Not to mention that a function call with 20 arguments is not readable at all.

I think it's a huge improvement.

bburtin: What about the idea of adding a new class that encapsulates the fields that are passed to the `encrypt()` function?  It's brittle when application code depends on the structure of the command line spec.  Having a separate class is a little more tedious, but I think it's less painful than tightly coupling the application code to the format of the command line arguments.

xcllnt: It's already a class. values is a proper object. There's no dependency on command-line order or its structure. Hence it's no more brittle than creating an encapsulating class. What we have is less obfuscating, which in my eyes means it's preferred. So, once we have a good enough reason to abstract (such as when we want to move more CSP-specific code to common code to reduce the duplication and find that the ad-hoc and different naming of fields in values makes that painful), then I'd say "yeah, maybe" (I would first simply rename the fields to remove the gratuitous differences :-)

PR #772, opened 2017-10-26
Default crypto policy to XTS

krishnanrs:
While single-disk encryption is enabled for GCP, for the other CSPs
the crypto_policy is being defaulted to GCM instead of XTS resulting
in the guest root volume being 2x the size.


xcllnt:
LGTM

PR #774, opened 2017-10-30
Better handling of encryption status for large volumes

krishnanrs:
With very large root volumes, the encryption progress, which is currently
printed as a integer might take a long time to increment and as a result
the encryption itself might timeout interpreting as a lack of progress.
This change:
 - Prints the encryption progress as a float with 2 significant decimal places
 - Checks the bytes_written for the encryption status instead of the progress percentage


krishnanrs:
The sure-fire way to make sure that encryption is progressing is simply to check the bytes_written. Hence the alternative would be to continue printing the progress percentage as an integer, even if it does not increment for a long time, while simply compare the bytes_written to verify if the encryption is progressing. This is especially important for encrypting extremely large (root) volumes or alternatively when the progress is slow (slow IO, CSP issues etc.)

PR #775, opened 2017-10-31
Fix traceback error during image update

krishnanrs:
When doing an image update, there is no 'bytes_written' field returned
in the status. Hence check if the field exists. If not, initialize it
to 0.

### DIFF
--- a/brkt_cli/encryptor_service.py
+++ b/brkt_cli/encryptor_service.py
@@ -211,7 +211,8 @@ def wait_for_encryption(enc_svc,
 
         state = status['state']
         percent_complete = status['percent_complete']
-        bytes_written = status['bytes_written']
+        # For image updates, there is no bytes_written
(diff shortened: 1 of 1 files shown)


brkt_cli/encryptor_service.py
```diff
@@ -211,7 +211,11 @@ def wait_for_encryption(enc_svc,
 
         state = status['state']
         percent_complete = status['percent_complete']
-        bytes_written = status['bytes_written']
+        # For image updates, there is no bytes_written
+        if 'bytes_written' in status:
+            bytes_written = status['bytes_written']
+        else:
+            bytes_written = 0
```
bonzotronix: Or be suave with a nice
> bytes_written = status.get('bytes_written', 0)

krishnanrs: Please review again

